PR #4, opened 2023-03-23
Address build issues with running development servers

dmfalke:
### Description

Removing generated code breaks development servers.

#### Build files retained

This can be tested by starting a dev site:

```
yarn nx start @veupathdb/clinepi-site
```

and then running the following in a new terminal, once the site comes up:

```
yarn nx run-many --target=build-npm-modules --skip-nx-cache
```

Webpack should not throw any errors.

#### Changes are picked up by the development server

When running a development server (such as `yarn nx start @veupathdb/eda` or `yarn nx start @veupathdb/clinepi-site`),
use the following command to rebuild changes made to dependencies, and to have the dev site reload with the changes:

```
cd packages/lib/<package>
yarn build-npm-modules
```

_Using the equivalent `nx` command (`yarn nx build-npm-modules @veupathdb/<package>`) has proven inadequate in this scenario._


packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -1988,6 +1979,7 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
         ]}
       />
 
+      <h1>foo</h1>
```
jernestmyers, line 1982: Don't suspect we want this in our legends 😄

packages/libs/web-common/src/App/Showcase/CardList.jsx
```diff
@@ -3,12 +3,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Select from 'react-select';
 import { Link, RealTimeSearchBox } from '@veupathdb/wdk-client/lib/Components';
-import { projectId } from '@veupathdb/web-common/lib/config';
+import { projectId } from 'ebrc-client/config';
```
jernestmyers, line 6: How does this path get resolved?

dmfalke: There is a typescript alias in this package's tsconfig

dmfalke: The previous path was causing typescript to try to recompile the compiled code, which caused a "Cannot write file ... because it would overwrite input file." error.

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
dmfalke: Removed!

README.adoc
```diff
@@ -37,3 +37,15 @@ For example, you can start the MultiBLAST dev server by running
...
 yarn nx run @veupathdb/multi-blast:start
 ```
+
+### Rebuild dependencies when running a development server
+
+When running a development server (such as `yarn nx start @veupathdb/eda` or `yarn nx start @veupathdb/clinepi-site`),
+use the following command to rebuild changes made to dependencies, and to have the dev site reload with the changes:
+
+```
+cd packages/lib/<package>
```
jernestmyers, line 47: Nit: `lib` should be `libs`

jernestmyers:
I was able to make changes in `eda`, `wdk-client`, and `web-common` while running a local `clinepi-site` dev server without any issues. To be clear, I was following @dmfalke's advice by running `yarn build-npm-modules` in the package's directory instead of trying to leverage an `nx` command.

One curiosity that was consistent in my testing was that I needed to manually refresh the page for the first change to be reflected, whereas subsequent changes forced a page refresh. Until this gets resolved, it is sharing this with other developers.

dmfalke:
> One curiosity that was consistent in my testing was that I needed to manually refresh the page for the first change to be reflected, whereas subsequent changes forced a page refresh. Until this gets resolved, we should share this with other developers.

I'm seeing the same issue. I will add a note to the readme.

PR #5, opened 2023-03-23
Pin volta yarn version

dmfalke:


bobular (approved):
For the record... It also worked for me and volta 1.1.1 with
```
   "volta": {
-    "node": "16.17.1"
+    "node": "16.17.1",
+    "yarn": "3.3.1"
   },
```

But this doesn't work with older versions of volta.  (1.0.x?)

@dmfalke said:

> So, we can either advise devs to upgrade to the latest version of volta, or we can pin a version of yarn that is supported by older versions of volta.
> In the latter case, yarn will see the packageManager field and install that version of yarn locally

bobular (approved):

PR #10, opened 2023-03-23
add banner for continuous overlay variable for scatterplot Viz

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1697.

An example screenshot is attached:

scatter-banner-continuous-variable1


bobular:
Had to merge `main` to get `yarn` working

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -561,6 +553,16 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
       )
         return undefined;
 
+      // prevent data request for the case of  plot option != Raw when using continuous overlayVariable
+      if (
```
bobular, line 557: For maintainability, we should perform this logic test once and store it in a `const`.
See what we do for `showLogScaleBanner`.  However, we don't do it properly for the logscale banner either.
We repeat the logic test in the `return null` part of the `data = usePromise...`.  I think it would better in both cases to use the once-calculated boolean for both rendering and the `return null`.  This may mean an extra dependency or two for the usePromise but we gain some readability and maintainability.

bobular (changes requested):
Works great - just one suggestion for improved maintainability.

There were some indentation-only diffs that I have ignored. I hope they weren't hiding any actual changes.

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: @bobular Thank you for your tests, reviews, and suggestions (and updating this PR too!) :thumb I also so your new ticket concerning formatting and yes I didn't touch those things. Perhaps something related to such a formatting (e.g., ESLint and/or prettier) is updated while creating the monorepo.

I also thought of that simplification. I guess it would not be that difficult so will check it out tonight.

moontrip: @bobular I think I addressed it so it would be great if you can check it out when you are available 👍

bobular (approved):
Looks neat and tidy. Works too! Thanks :-)

moontrip:
@bobular Thank you for your review and tests 👍

PR #11, opened 2023-03-24
Cache yarn operation

dmfalke:
Testing out the cache option for github actinos. See https://github.com/actions/setup-node#caching-global-packages-data

**Edit** See the cache associated with this PR: https://github.com/VEuPathDB/web-monorepo/actions/caches

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -7,13 +7,13 @@ jobs:
   main:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


jernestmyers:
Out of curiosity, did you attempt to use `cache-dependency-path` as shown i[n the link you provided](https://github.com/actions/setup-node#caching-global-packages-data)?

dmfalke:
> Out of curiosity, did you attempt to use `cache-dependency-path` as shown in [the link you provided](https://github.com/actions/setup-node#caching-global-packages-data)?

No. My understanding is that it uses the root `yarn.lock` file to generate the cache key, so I don't think we need to set that parameter.

PR #14, opened 2023-03-24
Eda/1510 add citations

jernestmyers:
Closes https://github.com/VEuPathDB/web-eda/issues/1510


packages/libs/eda/src/lib/workspace/DownloadTab/PastRelease.tsx
```diff
@@ -96,13 +98,17 @@ export default function PastRelease({
       id={`Past Release Dataset - ${release.releaseNumber}`}
       style={{ marginBottom: 35 }}
     >
+      {/**
+       * debt: change string types to ReactNode types as appropriate
```
asizemore, line 102: worth a tech debt ticket?

packages/libs/eda/src/lib/workspace/DownloadTab/StudyCitation.tsx
```diff
@@ -0,0 +1,151 @@
...
+          modifiers: {
+            computeStyle: {
+              fn: computeStyleFn,
+              gpuAcceleration: true,
+            },
+          },
+        }}
+      >
+        <span
+          onMouseOver={() => (hoveredState ? undefined : setHoveredState(true))}
```
asizemore, line 117: It felt a bit odd to have hovering/copying over both the citation string and the button. Can we do just one or the other? I'm partial to the copy button strategy :)

asizemore (approved):
Looking good!
I had a few styling questions, but the code works and the problem is resolved ✅ 

Right now the dataset citation is underneath the "Data Accessibility: {value}". That's not really the appropriate heading for the citation. Maybe "About this study" and then underneath that is data accessibility and the citation? Probably a question for the clinepi folks. I ask bc i would miss the dataset citation because i wouldn't be looking for it under the current heading. Probably even just changing to "Data Accessibility" and then the actual value could be underneath the heading, along wtih the citation, would be enough. What do you think?

jernestmyers:
> Right now the dataset citation is underneath the "Data Accessibility: {value}". That's not really the appropriate heading for the citation. Maybe "About this study" and then underneath that is data accessibility and the citation? Probably a question for the clinepi folks. I ask bc i would miss the dataset citation because i wouldn't be looking for it under the current heading. Probably even just changing to "Data Accessibility" and then the actual value could be underneath the heading, along wtih the citation, would be enough. What do you think?

I agree now that you point it out. @nkittur-uga , what are your thoughts?

packages/libs/eda/src/lib/workspace/DownloadTab/StudyCitation.tsx
jernestmyers: I can revert these changes and just do the copy button. The `followCursor` attribute is available in newer MUI versions, so at least it was fun implementing this logic if we are to remove it! 😂

nkittur-uga:
Yes, @jernestmyers and @asizemore I agree that the heading does not look right. 

Would it be easiest to make the text "Dataset citation" as large (font size) as "Dataset accessibility"?

jernestmyers:
@asizemore Here's how @nkittur-uga 's request would look with a few tweaks:
- I moved the "Please cite" statement beneath the Dataset Citation heading
- I removed "Dataset" from "Dataset Citation" since it now feels redundant to the heading itself

Thoughts?

image

nkittur-uga:
@jernestmyers We like how that looks!! Just one suggestion-

Could the existing "Please cite..." text be edited to below:

"Please cite that you accessed data via ClinEpiDB. The citation below can be used to reference the latest version of the data. Citations for previous data versions can be found under each release."

nkittur-uga:
@jernestmyers One last comment- Danica wonders if we should remove the "Citation:" in front of the citation- is it redundant? I think it does the job of separating the actual citation from the text before it. Maybe removing "Citation:" and instead italicizing or bolding the citation would do the job. Please do as you see fit.

asizemore:
Looks great! Thank you!!
I agree that we can remove "Citation:" but i think we should either leave it styled as-is or italicize. The big heading draws enough attention to it that i don't think it needs bolding.

jernestmyers:
Just pushed the "final" changes:
1. Added the extra verbiage to the "Please cite" text
2. Italicized the citation. This looked better to me than retaining the styling as-is
3. Removed the ability to copy the citation by clicking the citation itself, thereby removing the custom tooltip logic (now we just leverage the tooltip built into the CoreUI buttons, for better or worse). Citation can only be copied via the button.

image

asizemore:
Looks great! Thanks @jernestmyers !

PR #15, opened 2023-03-24
Place filter chips above tabs and remove show all filters button

jernestmyers:
Closes https://github.com/VEuPathDB/web-eda/issues/1621

Note: I've chosen to only comment-out the code related to the global filters button in the event we want to keep it.

### DIFF
--- a/packages/libs/eda/src/lib/workspace/AnalysisPanel.tsx
+++ b/packages/libs/eda/src/lib/workspace/AnalysisPanel.tsx
@@ -155,12 +155,10 @@ export function AnalysisPanel({
    * 2. split lastVarPath into an array structured as ['', entityId, variableId]
    */
   const isSubsetTabSelected = location.pathname.includes('variables');
-  const splitLastVarPath = lastVarPath.split('/');
-  const selectedEntityId = isSubsetTabSelected
-    ? splitLastVarPath[1]
-    : undefined;
+  const lastVarPathList = lastVarPath.split('/');
+  const selectedEntityId = isSubsetTabSelected ? lastVarPathList[1] : undefined;
   const selectedVariableId = isSubsetTabSelected
-    ? splitLastVarPath[2]
+    ? lastVarPathList[2]
     : undefined;
 
   const permissionsValue = usePermissions();
--- a/packages/libs/eda/src/lib/workspace/AnalysisPanel.tsx
+++ b/packages/libs/eda/src/lib/workspace/AnalysisPanel.tsx
@@ -54,6 +54,7 @@ import { FullScreenAppPlugin } from '../core/types/fullScreenApp';
 import FullScreenContainer from '../core/components/fullScreenApps/FullScreenContainer';
 import useUITheme from '@veupathdb/coreui/dist/components/theming/useUITheme';
 import { VariableLinkConfig } from '../core/components/VariableLink';
+import FilterChipList from '../core/components/FilterChipList';
 
 const AnalysisTabErrorBoundary = ({
   children,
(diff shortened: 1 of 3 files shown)


packages/libs/eda/src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -143,10 +144,24 @@ export function AnalysisPanel({
...
+    useState<boolean>(false);
+
+  /**
+   * Used in FilterChipList to highlight filter chip if/when an on-view Browse & Subset
+   * var matches a filter chip. Logic is:
+   * 1. check pathname to determine if Browse & Subset tab is selected
+   * 2. split lastVarPath into an array structured as ['', entityId, variableId]
+   */
+  const isSubsetTabSelected = location.pathname.includes('variables');
+  const splitLastVarPath = lastVarPath.split('/');
```
adnauseum, line 158: **nit**: we could prolly just call this `lastVarPath` without referencing `.split()`

packages/libs/eda/src/lib/workspace/AnalysisPanel.tsx
```diff
@@ -312,40 +327,59 @@ export function AnalysisPanel({
...
-                display: 'Record Notes',
-                route: '/notes',
-              },
-            ]}
-          />
+          <div className="EDAWorkspaceNavigation">
+            <div className="FilterChips">
+              <FilterChipList
+                filters={filters}
+                removeFilter={(filter) =>
```
adnauseum, line 334: This isn't feedback, but I thought it'd be a fun fact. This is an unnamed function. Function names (if provided) appear in stack traces which can help you quickly identify which function caused the error. This can be helpful when a function is getting passed as a prop.

packages/libs/eda/src/lib/workspace/EDAWorkspace.scss
```diff
@@ -169,9 +170,6 @@
     content: '';
   }
 
-  .expanded-diagram {
```
adnauseum, line 172: Are you sure about this? It looks like mission critical CSS!!! :trollface:

adnauseum (approved):
Awesome job!!! 🚀 

This is probably just restating your plans, but when this PR's out of draft mode I think we should name it closer to what it does. Like commit messages, many folks like PR titles to be in the imperative mood and clear: "Place filter chips above tabs and remove show all filters button." If ever you need to figure out why you did something in the past, the titles are really helpful.

packages/libs/eda/src/lib/workspace/AnalysisPanel.tsx
jernestmyers: Can you please elaborate on this? Initially I didn't declare a new variable (`splitLastVarPath`) and instead did something like `lastVarPath.split('/')[1]`. Since I did it twice, I figured I'd just go ahead and declare it as a variable. I'm not quite understanding how we'd "just call this `lastVarPath`" since that variable name already exists in the namespace.

adnauseum: Ah yes, that's totally a thing that I overlooked.

Maybe something like `lastVarPathParts` or `lastVarPathList`? I was poorly trying to say express that "split" feels verb-y, but that the rest of the variable name is composed of nouns. Moreover, if we somehow change `lastVarPath` from a string to a list of string (doubt we'd do that), then `split` would be confusing.

That all said, this isn't at all a big deal 🙃

jernestmyers: Ah, yes. My mind was reading "split" as a noun (like banana split), but that's not at all intuitive! I'll change its name so it doesn't read as a function name (verb).

adnauseum: Wow, I forgot about banana splits! 

> A banana split is an American ice cream-based dessert consisting of a peeled banana cut in half lengthwise, and served with ice-cream and sauce between the two pieces. There are many variations, but the classic banana split is made with three scoops of ice cream (one each of vanilla, chocolate, and strawberry).
> \- https://en.wikipedia.org/wiki/Banana_split

adnauseum (approved):
🚀

PR #19, opened 2023-03-27
Volcano

asizemore:
Resolves https://github.com/VEuPathDB/web-components/issues/463

Required for first delivery of mbio differential abundance app. More details on the plot [here](https://veupathdb.atlassian.net/wiki/spaces/FG/pages/144769127/Volcano+Plot+Specifications)

This PR adds a volcano plot component using visx. It does not attempt to add everything necessary for the "fully functional" volcano plot that lives in the EDA, but gives the backbones for us to build upon in the future.

To Dos left for later work
- [ ] Tooltips. I did give these a shot following the example on [this page](https://airbnb.io/visx/docs/xychart). The tooltip did appear, but not over all the points. It liked some but not others? Also sometimes it'd show the wrong data for the point on which i was overing. So not really sure about that just yet. https://github.com/VEuPathDB/web-monorepo/issues/66
- [ ] Make the three colors nicer. I've picked three that are good looking. Data viz should give the final say in if these are the best colors or not. https://github.com/VEuPathDB/web-monorepo/issues/67
- [ ] Make custom theme. Visx allows us to make themes for plots (similar to plotly). The styles for the axes, etc. should be put in some other file somewhere and saved as a visx theme so that we can reapply it in other components. https://github.com/VEuPathDB/web-monorepo/issues/68
- [ ] Truncation warnings. The plot is just an svg so my plan for the approach is to draw rectangles on it. https://github.com/VEuPathDB/web-monorepo/issues/69
- [ ] Add "up in group {a}" to xaxis. My plan is to use [annotations](https://airbnb.io/visx/docs/annotation). https://github.com/VEuPathDB/web-monorepo/issues/70
- [ ] Allow significance threshold to be set via p value or FDR https://github.com/VEuPathDB/web-monorepo/issues/71
- [ ] Allow the user to override the default axis ranges https://github.com/VEuPathDB/web-monorepo/issues/72

Other notes:
- my approach for changing the opacity is pretty lame right now. I couldn't figure out how to pass opacity to the glyph series. It must be possible...


asizemore:
One question - i'm starting to change my mind about who does the color choices. Maybe plot.data only sends back data and the component breaks up into colors. Seems like this volcano component could be a little smarter and not rely so much on everyone else...

asizemore:
@bobular made the good point that if dividers/thresholds are draggable, then the coloring _has_ to be on the frontend. 
So no one look yet while i do a tiny bit of refactoring :D

asizemore:
All up and running now! Ready for review!

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -0,0 +1,274 @@
+import { PlotProps } from './PlotlyPlot';
+
+import { significanceColors } from '../types/plots';
+import { VolcanoPlotData } from '../types/plots/volcanoplot';
+import { NumberRange } from '../types/general';
+import {
+  XYChart,
+  Tooltip,
```
adnauseum, line 8: **nit**: `'Tooltip' is declared but its value is never read.`

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+  const dataXMin = min(data.foldChange.map(Number));
+  const dataXMax = max(data.foldChange.map(Number));
+  const dataYMin = min(data.pValue.map(Number));
+  const dataYMax = max(data.pValue.map(Number));
+
+  // Determine mins, maxes of axes in the plot.
+  // These are different than the data mins/maxes because
+  // of the log transform and the little bit of padding.
+  //
```
adnauseum, line 84: **nit**: Dangling `//` 👯

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -0,0 +1,274 @@
...
+  if (dataYMin && dataYMax) {
+    yMin = -Math.log10(dataYMax);
+    yMax = -Math.log10(dataYMin);
+    yMin = yMin - (yMax - yMin) * 0.05;
+    yMax = yMax + (yMax - yMin) * 0.05;
+  } else {
+    yMin = 0;
+    yMax = 0;
+  }
+  console.log(yMin);
```
adnauseum, line 110: **question**: did you want this to stay here?

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -0,0 +1,274 @@
...
+  };
+  const gridStyles = {
+    stroke: '#dddddd',
+    strokeWidth: 0.5,
+  };
+  console.log(yMin);
+
+  return (
+    // From docs " For correct tooltip positioning, it is important to wrap your
+    // component in an element (e.g., div) with relative positioning."
```
adnauseum, line 179: **random**: If you or anyone would like to understand absolute/relative positioning, [this Codepen might help](https://codepen.io/skendrick366/pen/WNabrep).

The docs say this because it sounds like the tooltips are positioned absolutely. Absolutely position elements position themselves in relation to their nearest parent that's positioned relatively. 🤯 

You likely know this, but I just wanted to put it in there in case!

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+ * Assign color to point based on significance and magnitude change thresholds
+ */
+function assignSignificanceColor(
+  xValue: number, // has already been log2 transformed
+  yValue: number, // the raw pvalue
+  significanceThreshold: number,
+  log2FoldChangeThreshold: number,
+  significanceColors: string[] // Assuming the order is [high, low, not significant]
```
adnauseum, line 253: **suggestion**: This isn't at all unclear code or anything complicated, so this is really just a suggestion. I don't know if it makes the code better. Given this assumption, you could name the indicies:

```js
const HIGH = 0;
const LOW = 1;
const INSIGNIFICANT = 2;

 if (yValue >= significanceThreshold) {
    return significanceColors[INSIGNIFICANT];
  }

 if (xValue >= log2FoldChangeThreshold) {
    return significanceColors[HIGH];
  }
...
```

packages/libs/components/src/plots/VolcanoPlot.tsx
asizemore: i did not know this! Thank you! That was a really clear example

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -0,0 +1,272 @@
...
+  let xMin: number;
+  let xMax: number;
+  let yMin: number;
+  let yMax: number;
+
+  // Log transform for plotting, and add a little margin for axes
+  if (dataXMin && dataXMax) {
+    xMin = Math.log2(dataXMin);
+    xMax = Math.log2(dataXMax);
+    xMin = xMin - (xMax - xMin) * 0.05;
```
adnauseum, line 95: **nit/suggestion**: 0.05 is the measure of statistical significance, right? Maybe we can save that to a constant called `STATISTICAL_SIGNIFICANCE_THRESHOLD`?

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -0,0 +1,272 @@
...
+      pointId: data.pointId[ind],
+      color: assignSignificanceColor(
+        Math.log2(Number(fc)),
+        Number(data.pValue[ind]),
+        significanceThreshold,
+        log2FoldChangeThreshold,
+        significanceColors
+      ),
+    });
+  });
```
adnauseum, line 133: **fun fact**: this totally works! For some food for thought, some folks would write it like this:

```js
  const dataPoints: DataPoint[] = data.foldChange.reduce((datapoints: DataPoint[], fc, ind) => {
    return [
      ...datapoints,
      {
        foldChange: fc,
        pValue: data.pValue[ind],
        adjustedPValue: data.adjustedPValue[ind],
        pointId: data.pointId[ind],
        color: assignSignificanceColor(
          Math.log2(Number(fc)),
          Number(data.pValue[ind]),
          significanceThreshold,
          log2FoldChangeThreshold,
          significanceColors
        ),
      },
    ];
  }, []);
```

asizemore:
@dmfalke and @adnauseum ready for your review again!

I didn't end up making the find-min-and-max part into a function since we were sort of back and forth about it and it's a little unique to volcano. If you feel strongly about it though i'm happy to do it!

One new change - we learned that we don't expect to ever need the  raw fold change values, so we're just going to have the backend return log2(fold change) so that the frontend doesn't have to do it.

I'll turn to working on the [confluence](https://veupathdb.atlassian.net/wiki/spaces/~61aa65eec15977006afd91ad/pages/247103686/Working+with+Visx) doc for a bit while it's all fresh in my mind.

adnauseum (approved):
🚀 🌋 🙌 

LGTM!!!

packages/libs/components/.storybook/main.js
```diff
@@ -25,7 +25,11 @@ module.exports = {
       test: /\.(js|jsx)$/,
       loader: require.resolve('babel-loader'),
       options: {
-        plugins: ['@babel/plugin-proposal-nullish-coalescing-operator'],
+        plugins: [
+          '@babel/plugin-proposal-nullish-coalescing-operator',
+          '@babel/plugin-syntax-class-properties',
```
dmfalke, line 30: Why was this added? I don't see class properties used anywhere.

dmfalke (approved):
This all looks great! Just one question about the babel plugin.

packages/libs/components/.storybook/main.js
adnauseum: Thank you!! I forgot to ask about this in our PR review! I was curious since it's scoped to storybook.

packages/libs/components/src/types/plots/addOns.ts
```diff
@@ -283,6 +283,9 @@ export const gradientConvergingColorscaleMap = scaleLinear<string>()
   .range(ConvergingGradientColorscale)
   .interpolate(interpolateLab);
 
+// Significance colors
+export const significanceColors = ['#AC3B4E', '#0E8FAB', '#B5B8B4'];
```
d-callan, line 287: This is the only thing I saw that made me a bit uncomfortable.. say some instances of volcano plot want to use two colors rather than three, and not distinguish high vs low?

d-callan (approved):

packages/libs/components/.storybook/main.js
asizemore: ahh yes good catch! This was a byproduct of my adventures trying to install some of the visx packages a while back. I just removed it and all seems to work  👍

packages/libs/components/src/types/plots/addOns.ts
asizemore: I swapped the order around so it goes 'not significant, high, low". That way when we do have a use case for two colors only, the component can get a prop that says use 2 or 3 colors, and it'll just use the first two or first three colors from the list. If for some reason we wanted four colors later, we could just add onto this array.
This is assuming that we're happy with the colors being repeated in this way. Worth a discussion when that comes around i'd imagine.

I also am not considering these choices of colors final. I think they should go through data viz first. I'll make a new issue to be sure it doesn't get lost!

asizemore:
Thanks everyone for the feedback!!

PR #20, opened 2023-03-28
use reset icon for histogram filter

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1624. A screenshot is attached in the following:

histogramfilter-reset1


packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -654,122 +656,148 @@ function HistogramPlotWithControls({
...
+          >
+            <LabelledGroup label="X-axis controls"> </LabelledGroup>
+            <div style={{ marginLeft: '-2.6em', width: '50%' }}>
+              <ResetButtonCoreUI
+                size={'medium'}
+                text={''}
+                themeRole={'primary'}
+                tooltip={'Reset to defaults'}
+                disabled={false}
+                onPress={handleIndependentAxisSettingsReset}
```
asizemore, line 676: does this function cause the loading icon to show up? Every time i press the button i see the little spinner friend atop the plot.

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -654,122 +656,148 @@ function HistogramPlotWithControls({
...
+          <div
+            style={{
+              display: 'flex',
+              flexDirection: 'row',
+              alignItems: 'center',
             }}
-          />
-        </LabelledGroup>
+          >
+            <LabelledGroup label="X-axis controls"> </LabelledGroup>
```
asizemore, line 668: Does this still need to be a LabelledGroup? I think this can just be a heading?

asizemore (commented):
just a question or two! Looks good and works really well!

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
moontrip: Hi @asizemore Thank you for your tests and reviews. I have tested this and found that it even happened on old button style as well (tested using qa site) so it looks like it existed quite a long time without notice. I think that this deserves another ticket to be addressed.

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
moontrip: @asizemore We are using the LabelledGroup for all Visualizations as well. Technically it is not necessary to use it but for consistency it may need to be changed overall. Akin to above, this may be handled at another ticket but it is more like a technical debt level, I suppose.

asizemore (approved):
Great. Thanks for answering my questions!

moontrip:
@asizemore Thank you for your tests and reviews 👍

PR #21, opened 2023-03-28
Improve focus and selected state styles in SingleSelect

jernestmyers:
Closes https://github.com/VEuPathDB/EdaNewIssues/issues/559

Changes made include:
- A checkmark is rendered next to the selected item, similar to native HTML dropdowns. Previously, the selected item had the same style as the hovered item, which was confusing.
- Previously, "focus" state's styling also had the same style as the hovered item. Now, the focus state mimics Firefox's native focus state with a thin, dotted outline. _**Note**_: Before a selection is made, the dropdown will focus on the first item. This is so users can use the keyboard to navigate to their selection.

Selection made:
image

Before a selection is made:
image

Demos hover and selected state:
image


jernestmyers:
@danicahelb How's this implementation in response to the linked issue re: the confusing blue "focus" state for dropdowns when an item hasn't been selected yet?

danicahelb:
hey @jernestmyers what is the rational for this being different than the variable selectors on the viz tab in clinepi? the variables get underlined when i mouse over them. when i click to select one, the dropdown auto-closes and is populated with the variable i chose

in your mockups i'm not seeing how the dropdown goes away after the selection is made

jernestmyers:
> hey @jernestmyers what is the rational for this being different than the variable selectors on the viz tab in clinepi? the variables get underlined when i mouse over them. when i click to select one, the dropdown auto-closes and is populated with the variable i chose
> 
> in your mockups i'm not seeing how the dropdown goes away after the selection is made

The screenshots are all demonstrating how the options look when the button has been clicked but a _**new**_ selection hasn't been made. For example, in the third screenshot `A for Alligator` had previously been selected, and so when opening the dropdown `A for Alligator` remains the button's label and we can see the checkmark that indicates its selected state. The blue "hover" state was from my cursor being on `B for Beluga` but I had not actually clicked a new value yet, so the dropdown remained open. It remains the case that clicking (or keying up/down then pressing enter) will:
1. select an option
2. update the button's text
3. close the dropdown

I based this component off of [the native HTML select element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) and [an old `WDKClient` version](https://github.com/VEuPathDB/WDKClient/blob/master/Client/src/Components/InputControls/SingleSelect.tsx) of a `SingleSelect` component (that basically wraps the native select element). I think it serves a very different function from the variable tree dropdown, so its dropdown reflects that difference in my mind.

adnauseum:
> @adnauseum For PR purposes:
> 
> This is a relatively new component and thus is only used in a handful of instances. If you have the time and energy, I think it would be really helpful if you could assess its accessibility in addition to this PR's scope. I tried to add accessibility features when creating this component, but I fear I may have violated the adage "it's better to do nothing than to poorly implement accessibility" (or whatever the adage is 🥲).

I do have the time and energy!  I'm just going to post these notes here first. We can do the PR review together over Slack later today. 🗓️ 

First of all, I think it's great that you care about offering an accessible solution to our users. I don't at all want to undermine the improvements this PR brings! 👏  

Our team should refactor this component to either boil down to using the `<select>` or outsource it to a third party that has nailed it. Here's where our component accords and deviates from the native web standard `<select>`. Accessibility is like honesty: the honest person isn't usually the most loved, popular, or pretty, but they're doing it right! Whenever we "lie" to users (i.e., "this is a `<select>`, I promise") we have to work really hard to make it true. That's why all the popular kids in high school spend so much time fretting and laboring to try to make true the lie that they're cool. Anyways, all of this is beyond the scope of this PR. Please sit down, Sam.

Since you asked, here's a careless a11y audit of our component. I obviously have all the W3 spec memorized and DID NOT AT ALL consult [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#usage_notes). I'm a 10X dev and I actually know everything.

Keyboard users can select multiple contiguous items by:

✅ Focusing on the `<select>` element (e.g. using <kbd>Tab</kbd> ).
✅  Selecting an item at the top or bottom of the range they want to select using the <kbd>Up</kbd> and <kbd>Down</kbd> cursor keys to go up and down the options.
❌ Pressing <kbd>Space</kbd> to select/deselect "focused" select options. Our's presently uses <kbd>Enter</kbd>.
❌ After making a selection, focus is retained on the select component.
❌ After making a selection, focus is retained on the select component and users and open the select menu by pressing either <kbd>Up</kbd>, <kbd>Down</kbd>, or <kbd>Space</kbd>.

packages/libs/coreui/src/components/inputs/SingleSelect.tsx
```diff
@@ -1,142 +1,156 @@
...
-        if (key === 'Enter') {
-            handleSelection(newValue)
-        } else if (key === 'ArrowUp') {
-            indexOfFocusedElement !== 0 ? setIndexOfFocusedElement(prev => prev - 1) : null;
-        } else if (key === 'ArrowDown') {
-            indexOfFocusedElement !== items.length - 1 ? setIndexOfFocusedElement(prev => prev + 1) : null;
-        }
+  const onKeyDown = (key: string, newValue: T) => {
+    if (!isPopoverOpen) return;
+    if (key === 'Enter') {
```
adnauseum, line 59: **suggestion**: If we wish to emulate the native `<select>` keyboard behavior, then the focus should be put onto the select component after the user has made a selection.

adnauseum (approved):
🚀

PR #22, opened 2023-03-28
Move floating viz "activeViz" to parent outside of `FloatingVizManagement`

bobular:
I think this is probably the way to go. Reverses a lot of what I did the other day...

The alternative is to have all menu tabs rendered all the time (but not necessarily visible). I'm a bit uneasy about that.


adnauseum (approved):
🙇 

I think this is the way to go! 🚀

PR #24, opened 2023-03-28
fix star tooltip issue

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1638.

While it did not happen at Firefox as Bob said, I could replicate the issue at Google Chrome like Danica reported: you can test this at qa/Live site. Honestly, I don't know the exact reason even after fixing, but I guess that button disabled is conflicted with the tooltip at Chrome: disabling button actually do nothing in this case. After removing the disabled prop, the mouseover tooltip works just fine. Below is a video capture to demonstrate the fix.

https://user-images.githubusercontent.com/12802305/228326300-949962e9-de26-495d-98bf-c903e77e3f7e.mp4


dmfalke:
Sorry for the delayed review. I noticed that the `display: flex` is causing this issue. I wonder if we should explore a non-flex based layout instead of removing the `disabled` attribute?

moontrip:
> Sorry for the delayed review. I noticed that the `display: flex` is causing this issue. I wonder if we should explore a non-flex based layout instead of removing the `disabled` attribute?

Thank you for your suggestion, @dmfalke. I quickly tried to disable `display: grid` at the button element and test via dev tool. Although I could see that it mostly worked however the issue still lingered with some specific mouse-over location(s). Thus I doubt that it may not fully resolve the issue using non-flex/grid approach.

dmfalke:
Okay, thanks for checking! I guess your solution will do 🙂

dmfalke (approved):

moontrip:
Thanks @dmfalke for your reviews 👍

PR #25, opened 2023-03-28
Add error messaging for forbidden analysis imports

jernestmyers:
Closes https://github.com/VEuPathDB/web-eda/issues/1570


packages/libs/eda/src/lib/workspace/ImportAnalysis.tsx
```diff
@@ -44,15 +55,63 @@ export function ImportAnalysis({ analysisClient, analysisId }: Props) {
...
+        (error) => {
+          const enforcedErrorObject =
+            error instanceof Error ? error : new Error(String(error));
+          const errorMessageList = enforcedErrorObject.message.split('\n');
+          try {
+            setError(JSON.parse(errorMessageList[1]));
+          } catch {
+            setError(enforcedErrorObject.message);
+          }
+        }
```
adnauseum, line 67: **question**: do you think we should 1) rename the argument `error` to `errorObjectOrString` or 2) add a comment in this block to explain what things `error` could be?

I read this code to myself like this:

> So this block of code is the error callback on `Task.fromPromise().chain().run()`. It accepts an error callback that's invoked with an argument we're calling `error`. What is error? Oh, I see, the type of this argument is either `Error` or something else. What's the something else? 🤔  I see we're casting it to a string which is fed into the `Error` constructor.

Maybe this only gels with me 🤷 . Below is a a summary of the two options I could think of—feel free to take one, both, or neither.

```js
        (errorObjectOrString) => {
          const error =
            error instanceof Error ? error : new Error(String(error));
          const errorMessageList = enforcedErrorObject.message.split('\n');
          try {
            setError(JSON.parse(errorMessageList[1]));
          } catch {
            setError(error.message);
          }
        }
```

```js
        (error) => {
          // `error` is either an instance of `Error` or XYZ
          const enforcedErrorObject =
            error instanceof Error ? error : new Error(String(error));
          const errorMessageList = enforcedErrorObject.message.split('\n');
          try {
            setError(JSON.parse(errorMessageList[1]));
          } catch {
            setError(enforcedErrorObject.message);
          }
        }
```

packages/libs/eda/src/lib/workspace/ImportAnalysis.tsx
```diff
@@ -44,15 +55,63 @@ export function ImportAnalysis({ analysisClient, analysisId }: Props) {
...
 
+  const handleClick = useCallback(
+    (event: React.MouseEvent<HTMLButtonElement>) => {
+      event.preventDefault();
+
+      if (user === undefined || user.isGuest) {
+        dispatch(showLoginForm());
+      }
+    },
+    [history, user, dispatch]
```
adnauseum, line 79: > React Hook useCallback has an unnecessary dependency: 'history'. Either exclude it or remove the dependency array.eslintreact-hooks/exhaustive-deps)

I've got a handy ESLint plugin for VS Code that squiggles my hooks when they see something.

image

packages/libs/eda/src/lib/workspace/ImportAnalysis.tsx
```diff
@@ -44,15 +55,63 @@ export function ImportAnalysis({ analysisClient, analysisId }: Props) {
...
+          </p>
+        </>
+      ) : (
+        <pre>
+          status: {error.status}; message: {error.message}
+        </pre>
+      )
+    ) : (
+      <pre>{error}</pre>
+    );
```
adnauseum, line 109: **nit**: do you think this file would be improved by factoring this logic into a component?

Since this variable is JSX, we could make this logic more readable by creating a component at the bottom of this file. I personally get lost in nested ternaries. If the type wasn't JSX, I'd put this in an iife. But because we're in JSX land, we can easily factor this out into a component which is the same thing: a function that takes input and returns an output.

```js
function ErrorMessage({ error }) {
    if (!error) return whatever;

    const isErrorObject = typeof error === 'object';
    if (isErrorObject) {
        if (error.status === 'forbidden') {
            return something;
        }
    }

    
    return <pre>{error}</pre>
}
```

jernestmyers:
Could you give this one more pass-over? I am unable to find a reliable way to test these changes, so would feel (a bit) better with another set of eyes on the changes.

packages/libs/eda/src/lib/workspace/ImportAnalysis.tsx
```diff
@@ -76,20 +83,38 @@ export function ImportAnalysis({ analysisClient, analysisId }: Props) {
...
+type ErrorMessageProps = {
+  error: string | ImportAnalysisError | null;
+  user?: User;
+  handleLogin: (event: React.MouseEvent<HTMLButtonElement>) => void;
+};
+
+function ErrorMessage({ error, user, handleLogin }: ErrorMessageProps) {
+  if (!error) return null;
+  if (typeof error === 'object') {
+    if (error.status === 'forbidden') {
```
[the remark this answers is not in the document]
jernestmyers: TS isn't so happy with this idea, so I'll just leave it for now. 😅

PR #31, opened 2023-03-29
1690: put visualization component into draggable panel

adnauseum:
**Changes**:
- Rename `FloatingVizManagement` to `MapVizManagement`
- Add styling & behavior tweaks from Map UX meeting.
- Add grouped visualization picker: `<NewVisualizationPickerGrouped />`
- Render visualizations in a draggable panel.
- Reactor styling & behavior of visualization list.

**Before**:

image

**After**: 

https://user-images.githubusercontent.com/24446573/230714117-1daebbc4-78cf-471b-b277-bd5e27f51e73.mp4


bobular:
Looks great!  Just a note about the title.

The floating viz panel is not a "Configure XYZ" panel. It's a "XYZ" panel.  I know we may have ambitions to have a multi-step plot configuration that saves on screen space, but this would be the topic of another Map UX call (maybe tomorrow). Also, implementing that isn't trivial at the moment.

I would just make the floating plot draggable big enough to contain most of the plot for now, and we focus on wiring up other things and come back to optimizing the footprint of the floaters ASAP.

packages/libs/coreui/src/components/containers/DraggablePanel/index.tsx
```diff
@@ -1,28 +1,28 @@
-import { CSSProperties, ReactNode, useEffect, useState } from "react";
```
adnauseum, line 1: 🤔 I wonder if my editor config is fighting with some other config...

adnauseum: Ignore this file for now, I suppose.

bobular: After every commit the code should be going through `prettier` so our editors' quirks shouldn't matter. However, during the monorepo migration, something happened in this arena...

bobular (commented):
The draggable floating viz container is great for now.  Obviously longer term we have to save space and not make the user scroll down (which will require a bigger/wider refactoring), but for this issue/PR it is great.

To wrap up the parent issue - wait which issue is that? - ah [here it is](https://github.com/VEuPathDB/web-eda/issues/1499) - could we just tidy up the list of existing visualizations (with some "active status" styling) and make the new viz icons fit on the screen? (see my screenshot below) I would remove the "Unnamed computation (pass)" heading possibly get rid of the `[+ new plot]` button as previously discussed, but I'll leave that up to you. Thanks!

image

(damn - I did all this yesterday and forgot to submit it)

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -368,7 +368,7 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
         icon: <BarChartSharp />,
         renderWithApp: (app) => {
           return (
-            <FloatingVizManagement
+            <MapVizManagement
```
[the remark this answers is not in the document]
adnauseum: Thanks for catching that! Interesting behavior on VS Code's part. I went to <kbd>rename file</kbd> and it did a copy and kept the originals!

I deleted the originals in: https://github.com/VEuPathDB/web-monorepo/pull/31/commits/a99ff498c7991bca9e9602940689c82ae85ea633

packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -871,3 +971,47 @@ function deleteComputationWithNoVisualizations(
     ]);
   }
 }
+
+type VisualizationGroup = {
+  groupName: string;
+  groupDescription: string;
+  groupVisualizationsList: VisualizationOverview[];
+};
+function groupVisualizations(
```
adnauseum, line 980: Danielle mentioned that she can do this on the backend soon!

packages/libs/coreui/src/components/containers/DraggablePanel/index.tsx
adnauseum: > I think CoreUI had a different config

Ahh, that makes total sense! We could do a massive formatting pr in the monorepo!

packages/libs/eda/src/lib/map/index.ts
```diff
@@ -12,4 +12,4 @@ export type SiteInformationProps = {
 export const mapNavigationBackgroundColor = 'white';
 export const mapNavigationBorder: CSSProperties['border'] = '1px solid #D9D9D9';
 export const mapSideNavigationActiveMenuItemBackground: CSSProperties['background'] =
-  '#E1F5FF';
+  '#ECF7F9';
```
dmfalke, line 15: **Question:** should this value come from the theme palette? Not sure if we have something comparable.

dmfalke (approved):
This all looks great to me! See my question below. Otherwise, I think this is good to go.

packages/libs/eda/src/lib/map/index.ts
adnauseum: Whoops! I forgot to delete that! It was a dangler placeholder!! The component uses the theme hook!

https://github.com/VEuPathDB/web-monorepo/pull/31/commits/3a86b1c56dca486bc86d3fc28de4e41b0ff06834

packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
adnauseum: https://github.com/VEuPathDB/EdaDataService/pull/258

PR #35, opened 2023-03-30
introducing scientific notation for histogram stats and plot

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1677 (stats) and https://github.com/VEuPathDB/EdaNewIssues/issues/603 (plot). Currently significant figures are set to be 5.

A screenshot is attached here (same example used at https://github.com/VEuPathDB/web-eda/issues/1677#issuecomment-1454240302 is utilized here for a better comparison):

histogram-filter-segnis1

### DIFF
--- a/packages/libs/components/src/plots/Histogram.tsx
+++ b/packages/libs/components/src/plots/Histogram.tsx
@@ -337,44 +337,43 @@ const Histogram = makePlotlyPlotComponent(
       }
     }, [selectingRange, selectedRange, orientation, data.series]);
 
-    const standardIndependentAxisRange:
-      | NumberOrDateRange
-      | undefined = useMemo(() => {
-      if (binSummaries.length === 0) return undefined;
+    const standardIndependentAxisRange: NumberOrDateRange | undefined =
+      useMemo(() => {
+        if (binSummaries.length === 0) return undefined;
 
-      // If independentAxisRange (x-axis) is provided
-      // adjust the min of the range to the binStart of the bin that contains that value.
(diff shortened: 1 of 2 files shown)


moontrip:
Hi @asizemore, I am asking your review because you initiated the tickets 😃 
I think that when you run `@veupathdb/eda`, then you may need to do `yarn build-npm-modules` at `@veupathdb/components` because of the change at @veupathdb/components for plot

asizemore (approved):
looks good! Since there was some logic around years removed, it'd be helpful for whoever qa's this to check how it works for any variable with years that are not dates (not sure if we have any of those, but just in case :) )

moontrip:
@asizemore Thank you for your reviews and tests.  Yes there may be a year variable but I cannot remeber those variables. I will check it out if possible, and see how it works. 👍

moontrip:
@asizemore I found one variable that has year format: I may alleviate the case if not perfect. But one question to you is how to handle the mean value of the year. Is it okay to round up? e.g., if 2003.11881231231, then 2003? Or do you want to have a decimal points for that like 2003.12?

moontrip:
@asizemore Thank you for sharing your opinion. I made an adjustment to minimize the effect of year variable by checking type: integer, value >= 1900 & <=2100 which is not perfect but may work in most cases. At Live site, I noticed it uses 4 decimal points for the mean stats of the year so I followed it. Here is a screenshot after this change. Plot is fine without further change as the year number has 4 digit so scientific notation is not used in this case.

Amazonia ICEMR Brazil Cohort year variable1

PR #38, opened 2023-04-04
Fix entity constraints for computed variables

asizemore:
With [data service PR](https://github.com/VEuPathDB/EdaDataService/pull/254), resolves data service issue [246](https://github.com/VEuPathDB/EdaDataService/issues/246).

Original problem: When we decided which vars should be allowed in the var picker, we did not take into consideration which computed variables were already in the plot. Computed vars get an entityid (currently derived from the collection var from which they were computed), so we were able to make improper plot requests. For example, we could have alpha diversity computed from 16S data on the y axis, but then when selecting the x axis the var picker showed metagenomic data was available to be selected. If i _did_ select that metagenomic data, the backend would 500 because the request involves going around a branch point.

To solve this problem, box and scatter now have a new const `variablesForConstraints` that goes into `InputVariables`. Its function is help determine which vars should be used when computed or collection vars are involved. So generally it should be something like the union of computed, collection, and selected variables*.

*This is not always the case! For ranked abundance scatter, the collection entity lives on the y axis and in the overlay. If we added the collection var overlay to `variablesForConstraints`, sine the dataElementDependencyOrder is y, x, overlay, the x var would get squeezed into being only part of that same entity (the collection var entity). We want to be able to make plots with "Age" on the x axis, for example, so squeezing the x axis to fit into the collection var entity is undesirable. We should probably revisit constraints with the larger group soon anyways, but that's for a later time :)

To test, either use the data service [PR](https://github.com/VEuPathDB/EdaDataService/pull/254) or just revert the last commit to pretend we're getting the new `dataElementDependencyOrder` from the ds. Also, only studies with a branch point should be affected by this change. Mbio has 4, including FARMM and Preterm Infant Resistome 1.


asizemore:
ds pr has been merged. Should be deployed on my dev site if you'd like to use that for a backend.

jernestmyers:
> ds pr has been merged. Should be deployed on my dev site if you'd like to use that for a backend.

Thanks, I'll look at this now. I briefly tried to get things up and running yesterday but didn't know how to and then.... forgot about it!

bobular (approved):
I haven't tested but the code looks good and I _think_ I understand the thing about scatterplot. Basically the primary variable returned by the compute is for the y-axis, but it's stratified/overlaid by the taxon, but _the user can't select the overlay variable_ so there's no need to pass the overlay variable information through to InputVariables.  Or to paraphrase, the y+overlay always come together and the user can't change that.

asizemore:
> Basically the primary variable returned by the compute is for the y-axis, but it's stratified/overlaid by the taxon, but the user can't select the overlay variable so there's no need to pass the overlay variable information through to InputVariables. Or to paraphrase, the y+overlay always come together and the user can't change that.

Yes exactly! To add on, we also don't want to constrain on the overlay in this scenario because it will block out all interesting vars from the x axis. The y and overlay are of the same entity, so then the x would have to be from that entity, too. This is what i meant by the x getting squeezed into an entity. It's around on clinepi as well - check out any study and make a viz by picking some var from Sample for y, some var from Participant for x, then a var from Sample for overlay (in that order!). Then go back to change the x var. You'll see that the only allowed entity is Sample now, even though you have a plot where you're using a Participant var! Example below from the first Amazonia

Screen Shot 2023-04-07 at 6 30 13 AM

asizemore:
Thanks both! @jernestmyers found that it seems i broke the alpha div scatterplot with my backend merge, so i'll do some more fiddling around with that and then merge this PR if the frontend part still looks good

asizemore:
Turns out it was just a missing [comma](https://github.com/VEuPathDB/EdaDataService/pull/257). Anyways, confirmed it was a backend thing.

So thank you for the reviews! I'll do a final test and then merge!

PR #40, opened 2023-04-05
Account for site headers' vertical height in EDAWorkspaceNavigation

jernestmyers:
Resolves https://github.com/VEuPathDB/web-eda/issues/1621

I tested in both Mbio and ClinEpi sites. Screenshots below.

Note to reviewer: since `web-eda` dev server doesn't have the sticky site header, I didn't account for it. This is another instance in which the dev environment deviates from the final styling, which can lead to errors like this. I will open an issue that will articulate the need to "sync up" the styles applied in our dev environments and/or create an SOP for how we should be testing changes more robustly.

image

image

### DIFF
--- a/packages/libs/eda/src/lib/workspace/EDAWorkspace.scss
+++ b/packages/libs/eda/src/lib/workspace/EDAWorkspace.scss
@@ -142,7 +142,8 @@
 
   .EDAWorkspaceNavigation {
     position: sticky;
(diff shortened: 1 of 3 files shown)


adnauseum (approved):
This is a really great thing to bring up when we gather to talk about FE workflow / testing! 

Thank you!! 🚀

PR #42, opened 2023-04-05
Gradient legend refactor for scatterplot continuous overlay

bobular:
Here is a suggestion to make sure the colors in the legend gradient bar match the colors in the plot.

Related to issue https://github.com/VEuPathDB/web-components/issues/454 and the PR https://github.com/VEuPathDB/web-monorepo/pull/16 which I thought was too complex to maintain long-term, and we don't want to have parallel code that maps values to colors.

I've left as a draft PR because I'm not suggesting we merge this to main. There is other work in #16 that I haven't incorporated here. If there are any merge conflicts, I am happy to help.

It ended up being a bit of a rabbit hole because I had to fix the stories too, but they illustrate the new props to the gradient legend (overlayMin, overlayMax and a function to convert overlayValue to a color).


moontrip:
@bobular Thank you for your refactoring. Didn't look into them carefully yet, but at a glance, a big difference between your approach and mine is that you changed <stop> while I modified gradientColorscale directly. Basically our idea is similar each other to use a concept of bin to set a new colormap where you employed d3.range(). 😃 

I will examine your work and then make another comment 👍

moontrip:
@bobular I have carefully seen your refactoring. Although my approach works as well, to be honest, this is easier/simpler to understand than mine 👍 Also tested several cases including a use of arbitrary overlayMin/Max values and it seemed to work fine: didn't review/test stories though.

I realized that the computation of overlayMin/Max part, which exists at my PR, was not here yet, so I will add them in this PR tomorrow morning: it would be much better than forcibly merging from/to my PR due to many potential conflicts. Then, we can use this PR as the solution of the ticket instead of mine. Will get back to you after done 🤞

packages/libs/components/src/components/plotControls/PlotGradientLegend.tsx
```diff
@@ -26,36 +21,34 @@ const legendEllipsis = (label: string, ellipsisLength: number) => {
...
-    return (
-      <stop
-        offset={stopPercentage}
-        stopColor={color}
-        key={'gradientStop' + index}
-      />
-    );
-  });
+  // Create gradient stop points from the colorscale from values [legendMin TO legendMax] at nTicks steps
+  // Do we need to use more than 5 stop points to accurately reproduce the original color gradient?
```
asizemore, line 36: My first thought was yes, five seems fine. But then i remembered the diverging colorscale - it moves through more colors so would be harder to capture with just five stop points. 
I compared with the old legend and confirmed it looks like the 5 stop points is having trouble capturing the gradient. I'd probably bump this up to 7 or 9 for diverging/converging gradients. I'll add a pic in the comments to show what i mean

asizemore (commented):
Very slick! I like the new `valueToColorMapper` organization. 

I added a comment about the number of stop points and wanted to add a pic to go along with it:
Screen Shot 2023-04-06 at 6 12 18 AM
The left is a screenshot of a previous version of the legend and the right is the diverging colormap story using this new code. We seem to be missing the midpoint and there are bright bands between ticks 2,3 and 3,4.

bobular:
Hi @moontrip 
I hope it's useful. Mainly refactoring Ann's code to be able to display partial color gradients in the legend.

> realized that the computation of overlayMin/Max part, which exists at my PR, was not here yet

Ah yes. Oops.  I also duplicated the `max(abs(...))` calculations and returned `overlayMin` and `overlayMax` as part of the `data` payload, even though these are not derived from the data response at all.  Oops again!  [This whole section](https://github.com/VEuPathDB/web-monorepo/pull/42/files#diff-e6086b984097a27872aca20233837b9044c856828e2a57a16e6946395605deb9R620-R697) can probably be dramatically reduced.

packages/libs/components/src/components/plotControls/PlotGradientLegend.tsx
bobular: Right - thanks, yes, I thought I saw similar but didn't do the side-to-side comparison.  That is clearly not right.

We could just use 50 (100?) stop points and be done with it? This one doesn't need to be an odd number, like the text-tic marks.  There are 21 "stop points" in the original `DivergingGradientColorscale` but this component doesn't know that any more!

bobular: I just [pushed](https://github.com/VEuPathDB/web-monorepo/pull/42/commits/cb8db5e151a37847d08fcbe95539cad98f09ec9f)  an arbitrary 50 stop points and it has improved dramatically.

image

moontrip:
> Hi @moontrip I hope it's useful. Mainly refactoring Ann's code to be able to display partial color gradients in the legend.
> 
> > realized that the computation of overlayMin/Max part, which exists at my PR, was not here yet
> 
> Ah yes. Oops. I also duplicated the `max(abs(...))` calculations and returned `overlayMin` and `overlayMax` as part of the `data` payload, even though these are not derived from the data response at all. Oops again! [This whole section](https://github.com/VEuPathDB/web-monorepo/pull/42/files#diff-e6086b984097a27872aca20233837b9044c856828e2a57a16e6946395605deb9R620-R697) can probably be dramatically reduced.

Hi @bobular No problem. I can take care of that range part. During that work, I will see if I can simplify those stuffs you mentioned 👍

moontrip (approved):

PR #43, opened 2023-04-06
Fix broken SearchBox icon placement in Safari

jernestmyers:
Recent updates to UPenn Macs surfaced an issue in Safari with the icon placement breaking in the `SearchBox` component. A bit of searching revealed that Safari has historically had issues with `position: absolute`. Also worth noting: the misaligned icon actually interferes with the functionality by blocking the "expand" toggle for the top-most item in the tree (`Haemoproteidae` in this case).

I initially used `position: absolute` as a wrapper with no height or width so that I could remove its contents from the document flow and then use `position: relative` on the icon itself. This was working fine cross-browser until this latest update.

I've removed the `position: absolute` wrapper. Instead, I added a "do nothing" `transform: rotate(0)` to the icon's parent container (`<label>`) and applied `positon: fixed` to the icons. The `transform` property changes the container that `position: fixed` is referencing, so the icons will positionally reference the `<label>` instead of the viewport.

Screenshot of the issue:
image

Screenshot with this branch applied:
image

### DIFF
--- a/packages/libs/coreui/src/components/inputs/SearchBox/SearchBox.tsx
+++ b/packages/libs/coreui/src/components/inputs/SearchBox/SearchBox.tsx
@@ -17,8 +17,7 @@ const StyledTooltip = withStyles((theme: Theme) => ({
   },
 }))(Tooltip);
 
-type Props = {
-
+type SearchBoxProps = {
   /** Set the autofocus property of the underlying HTMLTextInputElement */
   autoFocus?: boolean;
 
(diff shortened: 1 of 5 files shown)


adnauseum (approved):
TIL. 🚀 

Quick question: should we apply formatting to the entire codebase?

jernestmyers:
> Quick question: should we apply formatting to the entire codebase?

We hoped to do that during the migration by running `yarn nx format:write --all` but it broke some things. I do think we should revisit the idea. It'll be a massive git diff, but only one commit.

adnauseum:
> > Quick question: should we apply formatting to the entire codebase?
> 
> We hoped to do that during the migration by running `yarn nx format:write --all` but it broke some things. I do think we should revisit the idea. It'll be a massive git diff, but only one commit.

Roger that.

Yeah the only downside I can think of to doing a huge formatting PR is that the blame history will be something like `Jeremy, 3 weeks ago • Apply formatting to code base`.

PR #44, opened 2023-04-06
Computing range

moontrip:
This PR branched out from the https://github.com/VEuPathDB/web-monorepo/tree/gradient-legend-suggestion (https://github.com/VEuPathDB/web-monorepo/pull/42) to update computation of the range. In addition, some reorganizations are made.

@bobular, can you please check this out? That normalize part may be out of data part as well but I just leave it as is because of the condition to run it, which requires backend response to check. This PR can be merged into your branch or main, but for now its destination is set to be your branch.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -515,21 +515,17 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
     (vizConfig.valueSpecConfig === 'Smoothed mean with raw' ||
       vizConfig.valueSpecConfig === 'Best fit line with raw');
 
-  // non-zero baseline for continuous overlay variable
-  const defaultOverlayMinMax = useDefaultAxisRange(
-    overlayVariable,
-    undefined,
-    undefined,
-    undefined,
-    vizConfig.dependentAxisLogScale,
-    'Full',
-    // non-zero baseline
-    true
-  );
-
   // If numeric overlay, record the min and max and make a value to color map function
-  const overlayMin: number | undefined = defaultOverlayMinMax?.min as number;
-  const overlayMax: number | undefined = defaultOverlayMinMax?.max as number;
+  // assign 0 to avoid undefined
+  const overlayMin: number | undefined =
+    overlayVariable?.type === 'number' || overlayVariable?.type === 'integer'
+      ? overlayVariable?.distributionDefaults?.rangeMin
+      : 0;
+  const overlayMax: number | undefined =
+    overlayVariable?.type === 'number' || overlayVariable?.type === 'integer'
+      ? overlayVariable?.distributionDefaults?.rangeMax
+      : 0;
+
   // Diverging colorscale, assume 0 is midpoint. Colorscale must be symmetric around the midpoint
   const maxAbsOverlay =
     Math.abs(overlayMin) > overlayMax ? Math.abs(overlayMin) : overlayMax;
@@ -545,7 +541,6 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
         ? 'sequential reversed'
         : 'divergent'
       : undefined;
-  let overlayValueToColorMapper: ((a: number) => string) | undefined;
 
   const data = usePromise(
     useCallback(async (): Promise<ScatterPlotDataWithCoverage | undefined> => {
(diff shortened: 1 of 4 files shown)


packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -517,6 +515,38 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
...
+    | 'divergent'
+    | undefined =
+    overlayMin != null && overlayMax != null
+      ? overlayMin >= 0 && overlayMax >= 0
+        ? 'sequential'
+        : overlayMin <= 0 && overlayMax <= 0
+        ? 'sequential reversed'
+        : 'divergent'
+      : undefined;
+  let overlayValueToColorMapper: ((a: number) => string) | undefined;
```
bobular, line 548: This should go inside the `data = usePromise(...)` for two reasons
1. this particular variable is set inside the usePromise (a side-effect that we don't want) but is not accessed later...
2. ...because the overlayValueToColorMapper is returned as part of the `data` payload (from whence it is used to make the legend)

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -517,6 +515,38 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
     (vizConfig.valueSpecConfig === 'Smoothed mean with raw' ||
       vizConfig.valueSpecConfig === 'Best fit line with raw');
 
+  // non-zero baseline for continuous overlay variable
+  const defaultOverlayMinMax = useDefaultAxisRange(
+    overlayVariable,
+    undefined,
+    undefined,
+    undefined,
+    vizConfig.dependentAxisLogScale,
```
bobular, line 524: I'm not sure we should be passing `dependentAxisLogScale` for the "overlay scale" range?  `false` should be OK?

bobular (commented):
This is looking easier (for me of little brain) to review ;-) 

I had another look at the original request from Danica [here](https://github.com/VEuPathDB/web-components/issues/454#issuecomment-1433570548) and think we may need tweak the logic inside useDefaultAxisRange when it's in "overlay gradient mode". Or perhaps we don't use it at all. Sorry... this is turning into a very long process and I have probably given conflicting advice throughout.

In Danica's comment, she notes that the range for "mean height" shouldn't start at zero.  Unfortunately our code currently does start at zero (screenshot below is from this PR)

image

The reason for this is because the study variable metadata tells it to:

distributionDefaults |  
-- | --
displayRangeMin | 0
displayRangeMax | 121
rangeMin | 12.07
rangeMax | 120.07
binWidth | 0.8

And we take the min() of the mins and the max() of the maxes.

Maybe we need to take the max() of the mins and min() of the maxes for continuous overlays?  I will have to think about it again in the morning!

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
moontrip: I think you are probably right. Passing false would be good enough.

moontrip:
> 
 
I think that not starting from zero is the case of sequential type. Perhaps we can judge the type and then use range.min if sequential type? Either way, the range function will be more complex unfortunately.

bobular:
> the case of sequential type

Not sure what you mean by this?

bobular:
I think we should probably just use `rangeMin` and `rangeMax` for continuous overlay limits.

moontrip:
@bobular what I meant is that divergent type always starts from a negative value, thus the case to consider starting from (or ending to) zero is a sequential type.

I agree with you that using the rangeMin/Max only would work. I think that in this case we don’t need to use the range hook: no need to make the hook/function have more conditions for this specific case for now, I suppose.

bobular:
Sounds good. Sorry my "help" has prolonged the issue!

moontrip:
@bobular Your helps and suggestions were great 👍 Have a great weekend!

moontrip:
@bobular I made changes following your suggestions and our discussions. Please note that I assigned overlayMin/Max as zero instead of undefined to avoid type issue at the normalize function. 

If you don't mind, we can merge this to main directly instead of your branch.

bobular:
Hi @moontrip 

It looks like the diverging colorscale isn't being used here. Which is very odd because the code looks OK to me.

image

PR #48, opened 2023-04-10
Reorder zIndexes in MiniMap, RequiredInputPrompts, EDAWorkspaceNavigation

### LINKED ISSUE #46: genomics header menu and EDA tabs z-index issue
Screen Shot 2023-04-08 at 5 04 20 PM

jernestmyers:
Resolves #46 

Checked these changes in local development site servers for ClinEpi and VectorBase. 

Screenshot shows the `MiniMap` rendering beneath the header and the dropdown menu rendering above the workspace nav:
image

Screenshot shows the map and "required inputs" prompt rendering beneath the workspace nav:
image


dmfalke:
I think there is a simpler solution. If we add `z-index: 1` to https://github.com/VEuPathDB/web-monorepo/blob/d94203b46082174b182dcb5c31fb2d6bf80ea68f/packages/libs/web-common/src/styles/main.scss#L6-L7, it should address these issues.

The reason this works is that it will turn the `.vpdb-Main` element into a new stacking context, at the same "level" as the header element. This will prevent child elements of `.vpdb-Main` from stacking on top of the header.

jernestmyers:
> I think there is a simpler solution. If we add `z-index: 1` to
> 
> https://github.com/VEuPathDB/web-monorepo/blob/d94203b46082174b182dcb5c31fb2d6bf80ea68f/packages/libs/web-common/src/styles/main.scss#L6-L7
> , it should address these issues.
> 
> The reason this works is that it will turn the `.vpdb-Main` element into a new stacking context, at the same "level" as the header element. This will prevent child elements of `.vpdb-Main` from stacking on top of the header.

Just pushed this change. Seems to work as expected. Thanks for the simple solution and the explanation!

dmfalke (approved):

PR #50, opened 2023-04-10
Lineplot 0/0 points updates

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1082
Original PR from `web-eda`: https://github.com/VEuPathDB/web-eda/pull/1686


moontrip:
Hi @chowington I am currently testing some cases but some questions were raised. More specifically, When I tested with a continuous xAxisVariable, it looks good and legend shows correct values (arguably one question but will discuss it later), but the case of an ordinal xAxisVariable makes me wonder something. Please note that I didn't do code reviews yet so my questions are based on what I saw in the web with this PR. Will test more.

First of all, here are test conditions:
Study data: SCORE S. mansoni Cluster Randomized Trial
Filter: a) Country=Tanzania; b) Village study arm= Arm 3
X & Y variables and the configuration of the Proportion can be found in the following screenshot(s)

**Q1)** The xAxisVariable (datashape: ordinal) has five categories from Year 1 to Year 5, and I noticed that Year 3 & Year 4 have 0/0 data with the filters and the proportion setting. Somehow it is displayed as two groups separately such as Year 1, Year 2, & Year 5, and Year 3 & Year 4. Honestly I am fine with the current format as it is more clear to me, but some may question whether it should be displayed as its order like from Year 1 to Year 5.

1411 lineplot question01


**Q2)** Data were grouped for 0/0 case as an independent series (seriesType: zeroOverZero): I like this 👍 But when I checked data, I found strange that those independent series have duplicate values at both X and Y data. In the following capture, I expected X: [Year 3, Year 4]; Y: [0, 0] (2 value each) but it has 6 values each. Is it intentional? Actually, I noticed the same thing at a continuous xAxisVariable. 

1411 lineplot data01

 
**Q3)** When I added an overlay variable, I got forEach error. Looks like an array or a value is undefined. 

1411 lineplot overlay error

chowington:
@moontrip, ready for official review! :)

chowington:
Note that `prettier` was insistent on making some benign formatting changes 🤷‍♂️

packages/libs/components/src/components/plotControls/PlotListLegend.tsx
```diff
@@ -49,173 +62,207 @@ export default function PlotListLegend({
...
-                  {item.marker === 'square' && (
-                    <div
-                      style={{
-                        height: defaultMarkerSize,
-                        width: defaultMarkerSize,
-                        borderWidth: '0',
-                        backgroundColor: item.markerColor,
+          {legendItems.map(
+            (item: LegendItemsProps) =>
+              !item.hideFromLegend && (
```
chowington, line 87: Note new `hideFromLegend` logic

packages/libs/components/src/components/plotControls/PlotListLegend.tsx
```diff
@@ -49,173 +62,207 @@ export default function PlotListLegend({
...
+                              backgroundColor:
+                                item.marker === 'circleOutline' ||
+                                markerBodyOpacity === 0
+                                  ? 'transparent'
+                                  : ColorMath.evaluate(
+                                      item.markerColor +
+                                        ' @a ' +
+                                        (markerBodyOpacity * 100).toString() +
+                                        '%'
+                                    ).result.css(),
```
chowington, line 179: Updated color logic

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -1852,38 +1871,32 @@ export function lineplotResponseToData(
...
-      overlayVariable,
-      colorPaletteOverride
-    );
+        vizType,
+        modeValue,
+        independentValueType,
+        dependentValueType,
+        showMissingOverlay,
+        hasMissingData,
+        dependentIsProportion,
```
chowington, line 1891: Added `dependentIsProportion` argument

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -1949,67 +1962,71 @@ type ArrayTypes = PickByType<
...
-    // which are the arrays in the series object?
-    // (assumption: the lengths of all arrays are all the same)
-    const arrayKeys = Object.keys(series)
-      .filter((_): _ is keyof LinePlotDataSeries => true)
-      .filter((key): key is keyof ArrayTypes => Array.isArray(series[key]));
-
-    const otherArrayKeys = arrayKeys.filter((key) => key !== 'y');
+    if (!(series.seriesType === 'zeroOverZero')) {
+      return series;
+    } else {
```
chowington, line 1974: Just added this `seriesType` check

moontrip (commented):
@chowington Overall, it is working great including ordinal variable (xAxis is string case)! 👍 I have just one comment. 

Previously, legend with 0/0 cases shows two legend items such as Data and Undefined Y (denominator of 0), but the latest PR only shows Undefined Y item. I am not quite sure if it is intentional, but I guess one may want to show Data as well? For example, Scatterplot has plot options such as Smoothed mean and Best fit, and in those cases, Data is shown as a legend item even without Overlay variable. 

This may be needed so perhaps you can ask Danica about this?

moontrip:
@chowington Oh I see. Then we can go ahead and see what the outreach team says 😄 I will approve this PR 👍

moontrip (approved):

PR #54, opened 2023-04-11
Use CollapsibleSection for large text displays in BLAST

jernestmyers:
Resolves https://redmine.apidb.org/issues/50314

Implemented a `CollapsibleSection` so users can control the wall of text that can be shown. The "cutoff" has been arbitrarily set at 500 characters.

Default view when page loads:
image

View when user clicks to "show more":
image

Similar logic/behavior was added to the `StepDetails`.


adnauseum (approved):
🚀

dmfalke (approved):

PR #60, opened 2023-04-13
add scatterplot marker opacity to legend

moontrip:
Hi @chowington While I am reviewing your codes, I realized that you did (temporarily) block scatterplot's marker opacity part, perhaps because of the conflict with Lineplot's 0/0 case. Thus, I tried to activate both cases here. I don't know you already did similar stuff but wish this would be helpful for you 😃 

Just in case, this is branched out from your lineplot-zero-over-zero branch


chowington (approved):
Thanks for the extra help, DK!

PR #65, opened 2023-04-14
Fix AnswerFilter bug by stopping propagation of click event

jernestmyers:
Resolves https://github.com/VEuPathDB/ApiCommonWebsite/issues/149

### DIFF
--- a/packages/libs/wdk-client/src/Views/Answer/AnswerFilter.jsx
+++ b/packages/libs/wdk-client/src/Views/Answer/AnswerFilter.jsx
@@ -46,8 +46,7 @@ class AnswerFilter extends React.Component {
     }
   }
 
-  toggleFilterFieldSelector(e) {
-    e.stopPropagation();
+  toggleFilterFieldSelector() {
     this.setState({
       showFilterFieldSelector: !this.state.showFilterFieldSelector,
     });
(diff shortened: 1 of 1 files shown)


packages/libs/wdk-client/src/Views/Answer/AnswerFilter.jsx
```diff
@@ -46,7 +46,8 @@ class AnswerFilter extends React.Component {
     }
   }
 
-  toggleFilterFieldSelector() {
+  toggleFilterFieldSelector(e) {
+    e.stopPropagation();
```
dmfalke, line 50: This will cause an error because this method is passed to `AnswerFilterSelector` and called it with no arguments.

Something different to try is adding also checking if `this.props.open === true` to the condition on this line: https://github.com/VEuPathDB/web-monorepo/blob/84d2f4204b094dec181456ea7f721bebab011f2f/packages/libs/wdk-client/src/Views/Answer/AnswerFilterSelector.jsx#L37

dmfalke (changes requested):
This is a close solution, but ultimately won't work (see comment below, with a suggested alternative approach). Let me know what you think.

packages/libs/wdk-client/src/Views/Answer/AnswerFilter.jsx
jernestmyers: I ended up moving the `stopPropagation` call to a click handler in the button itself. Adding `this.props.open` to that condition didn't prevent the undesired behavior. By the time the code reaches that condition, `this.props.open` still registers as true.

dmfalke: Ah, well, this looks like a great solution!

dmfalke (approved):

PR #74, opened 2023-04-17
formatting stats as power of 10 and add conditions

moontrip:
This will address the feedback from Danica, https://github.com/VEuPathDB/web-eda/issues/1677#issuecomment-1507694481. There is no specific way to display the power of 10 format like plotly, thus made several changes. Also, similar to plotly, some conditions are set to show the power of 10: >10^5 or <0.0001. Some screenshots are demoed in the following:

a) Age (months)
1490 histogramFilter Age01

b) Acidobacteriota: since plotly does not show the power of 10, the stats is also shown as floating numbers
1490 histogramFilter Acidobacteriota01

c) Aggregatibacter segnis: power of 10 format
1490 histogramFilter segnis

d) date variable: just displaying date as is
1490 histogramFilter date01

### DIFF
--- a/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
@@ -43,7 +43,7 @@ import Notification from '@veupathdb/components/lib/components/widgets//Notifica
 // import axis label unit util
 import { variableDisplayWithUnit } from '../../utils/variable-display';
 import { useDefaultAxisRange } from '../../hooks/computeDefaultAxisRange';
-import { min, max } from 'lodash';
+import { min, max, gt, lt } from 'lodash';
 import { useDebounce } from '../../hooks/debouncing';
 import { useDeepValue } from '../../hooks/immutability';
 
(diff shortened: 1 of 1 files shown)


packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -318,6 +318,13 @@ export function HistogramFilter(props: Props) {
     uiState.dependentAxisLogScale
   ) as NumberRange;
 
+  const fgSummaryStatsMin = formatStatValue(fgSummaryStats?.min, variable.type);
```
asizemore, line 321: Let's move these right below the definition of `fgSummaryStats` if possible? That way all of the foreground summary stats things get done right next to each other.

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -842,16 +835,33 @@ function tidyBinLabel(
...
+
+  let formattedValue: string | number | string[] =
+    type === 'date'
+      ? String(value).replace(/T.*$/, '')
+      : // check a possible year variable
+      type === 'integer' && Number(value) >= 1900 && Number(value) <= 2100
+      ? Number.isInteger(value)
+        ? Number(value)
+        : Number(value).toFixed(4)
+      : // set conditions similar to plotly
```
asizemore, line 849: What do you think about breaking logic up into two steps? There is a lot of logic to keep track of here :D 

Possibly we could move the year check to a new step? It looks like some of that year logic is repeated in the new code as well?

asizemore (commented):
Thanks @moontrip ! I had a few comments. This was the first time id seen the `formatStatsValue` function. It's very handy!

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
moontrip: Hi @asizemore thank you for your suggestions. 👍 yes it makes sense, so I will move them

moontrip:
@asizemore I have addressed your feedbacks. Can you please check them out?

asizemore (approved):
Looks great!!

moontrip:
Thanks @asizemore for your test and reviews 👍

PR #78, opened 2023-04-17
Link to dataset in internal gene tables

jernestmyers:
Resolves https://redmine.apidb.org/issues/19659?project=true

image

### DIFF
--- a/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/InternalGeneDataset.tsx
+++ b/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/InternalGeneDataset.tsx
@@ -1,20 +1,43 @@
-import React, { Suspense, useMemo, useState, useEffect, useCallback } from 'react';
+import React, {
+  Suspense,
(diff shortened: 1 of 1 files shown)


aurreco-uga (approved):
it is difficult to tell what has changed in the file.. with all the formatting changes..

jernestmyers:
> it is difficult to tell what has changed in the file.. with all the formatting changes..

Ah, didn't notice. Add this to the migration pains: we need to do a batch commit that applies the monorepo's `pre-commit` formatting so that we don't run into this anymore!

The only changes were [L328](https://github.com/VEuPathDB/web-monorepo/pull/78/files#diff-4185cf16425ca61d2f609aabfd99547abe101d4c7aadf91a3bf122dcfab40a20R328) which pulled out `dataset_id` and [L355-L357](https://github.com/VEuPathDB/web-monorepo/pull/78/files#diff-4185cf16425ca61d2f609aabfd99547abe101d4c7aadf91a3bf122dcfab40a20R355-R357) which wrapped the display name in an anchor tag!

PR #83, opened 2023-04-18
Add test-report script

dmfalke:
This PR adds a utility script to generate a test report. Eventually, I would like the generated files to be included as a release asset (maybe I should just do that now?).

I'm happy to learn if there are other tools out there for generating test reports, but this should suffice for the upcoming website release.


adnauseum (approved):
🚀

PR #87, opened 2023-04-19
Ensure side navigation doesn't render multiple scrollbars constantly 

adnauseum:
**Before**:

image

Certain browsers + OSs will render scrollbars all the time. Scrollbars are a major accessibility concern and browser rendering engines and OSs disagree as to how and when they should be rendered.

**After**:

This PR removes needless `overflow: scroll`s thereby resolving this UI bug.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapSideNavigation.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapSideNavigation.tsx
@@ -63,6 +63,7 @@ export function MapSideNavigation({
         // ensure that the side menu sits atop the map.
         zIndex: 10,
         display: 'flex',
+        overflow: 'scroll',
       }}
     >
       <button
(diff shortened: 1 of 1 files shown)


bobular:
Is my problem that my Windows FF is only on version 102? The About Firefox window says it's up to date but it's 112 on Linux.
On Windows I'm on the "esr" update channel. Whatever that is.

bobular:
This post reproduces the issue for me
https://support.mozilla.org/en-US/questions/1271761
Scrollbars on Windows, non scrollbars on Linux. Versions as in previous comment.

But our problem is not related to a `<canvas>` element.

bobular:
I'm going to leave this for now - but I don't think it's unreasonable to expect that we have users on an extended support version of Firefox.

bobular:
However...

It must be something to do with my laptop (accessibility settings??), because I see the same behaviour (big ugly scrollbars in the bottom right panel from the link below) in Chrome!!

https://jsfiddle.net/hefczx3a/12/

dmfalke:
FYI, I can see this on chromium-based browsers on linux:

image

dmfalke:
Why was this merged? I don't think the scrollbar issue has been fixed.
image

bobular:
That's odd. I don't remember doing this. Looks like I did the equivalent of pocket-dialling the merge. I'll revert it.

bobular:
I can't use github UI to revert it. Looking at the time of merge, it was exactly the same time (to the minute at least) as my merge of https://github.com/VEuPathDB/web-monorepo/pull/45  Very strange.

Sorry for the confusion. I'll make a new ticket and reference this one. There are very few changes, it should be easy to undo or tweak what was done here.

PR #91, opened 2023-04-20
control legend checkbox

moontrip:
This will resolve https://github.com/VEuPathDB/web-components/issues/414. 
A screenshot without checkbox is attached here.

control legend checkbox

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
@@ -5,10 +5,11 @@ import { LegendItemsProps } from '@veupathdb/components/lib/components/plotContr
 interface Props {
   legendItems: LegendItemsProps[];
   title?: string;
+  showCheckbox?: boolean;
 }
 
 export function MapLegend(props: Props) {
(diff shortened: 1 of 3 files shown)


packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
bobular, line 1: Given that `checkedLegendItems` can be `undefined`, I think we can leave it out of the props passed to PlotLegend at the following line. The current value passed is just "everything checked".

https://github.com/VEuPathDB/web-monorepo/pull/91/files#diff-ea5aca173d759a687fc3d1a258dba03ca5fdfbf79ebab328d5ba622114c00413R22

bobular (approved):
Apart from the tiny suggestion - looks great, thanks!

packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
moontrip: @bobular Thank you for your reviews and tests! 👍 The link somehow did not work, but I removed checkedLegendItems prop when passing to PlotLegend from MapLegend component.

PR #101, opened 2023-04-21
Removed the deprecated config

bobular:
Resolves https://github.com/VEuPathDB/web-components/issues/343

Wasn't expecting any problems but the components storybook and the EDA/SAM build and run fine.


bobular:
Came up during SAM ticket wrangling. An easy fix now in the monorepo. Picked you at random Connor!

PR #104, opened 2023-04-24
Update alpha map menu item, and add missing provider components.

dmfalke:
I noticed that the header menu item in genomics sites should point to the stand alone map. I aso noticed some missing provider components when a route is rendered fullscreen. This PR addresses those issues.


adnauseum (approved):
🙌

PR #100, opened 2023-04-21
Use new dataNoun prop in user dataset UI

jernestmyers:
Resolves #98 

Related to https://github.com/VEuPathDB/EdaNewIssues/issues/621

**THE PROBLEM:** 
The `My User Studies` table in EDA (or `My Data Sets` in genomics) **currently** has disparate verbiage ("Datasets", "data sets", "Studies") as shown:
image

**THE SOLUTION:**
In introducing a `dataNoun` prop, we can specify singular and plural forms of the appropriate noun used in genomics or EDA sites.

This screenshot shows this new prop implemented in EDA:
image

This screenshot shows this new prop implemented in genomics:
image

PR #110, opened 2023-04-24
Theme user datasets buttons

jernestmyers:
Resolves https://github.com/VEuPathDB/EdaNewIssues/issues/621 when combined with https://github.com/VEuPathDB/web-monorepo/pull/100

Replaced `Grant Access` and `Delete` buttons with `CoreUI` `MesaButton`s. The `Grant Access` button listens to the site's theme colors. The `Delete` button does not use theme colors but instead relies on style overrides to mimic the old look. Screenshots below.

**Mbio**:
image
image

**ClinEpi**:
image
image

**Plasmo**:
image
image

### DIFF
--- a/packages/libs/coreui/src/components/buttons/index.ts
+++ b/packages/libs/coreui/src/components/buttons/index.ts
@@ -41,7 +41,7 @@ type ButtonStateStyleSpec = {
     offsetX: string;
     offsetY: string;
     blurRadius: string;
-    color: string;
+    color: CSSProperties['color'];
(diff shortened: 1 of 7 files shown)


dmfalke:
Screenshots look great!

What do you think about moving `ThemedGrantAccessButton` into its own file, and also adding a `ThemedDeleteButton`, to avoid repeating all of the overrides?

dmfalke (approved):
:rocket:

PR #115, opened 2023-04-25
113/standalone-map-menu-remaining

### LINKED ISSUE #113: Standalone Map: Add remaining menu items
- [ ] Downloads
- [ ] Study details

dmfalke:
Fixes #113

This PR adds the download and study details content to the standalone map menu.

Note that we might want to consider preventing menu items from being unmounted, since some of them may require service calls. This would effectively cache those calls, which will make switching between menu items faster.


dmfalke:
That's what I was thinking. I'll make another issue for that.

PR #117, opened 2023-04-25
Top-align checkboxes in tree and tweak styles accordingly

jernestmyers:
Resolves https://redmine.apidb.org/issues/51313

Implemented the most straightforward suggestion from the redmine issue, which is to top-align the checkboxes with the label.

Screenshot from issue showing problem and possible solutions:
image

Screenshot of implementation in this PR:
image


adnauseum (approved):
🚀  Much better with just the alignment! I think the separators and striping are helpful, but probably overkill

dmfalke:
What do you think about adding a 2px top and bottom padding, as well? It's true that there is less visible on screen, but I think it really helps.

**Before**
image


**After**
image

adnauseum:
> > 🚀 Much better with just the alignment! I think the separators and striping are helpful, but probably overkill
> 
> I agree, but it's a bit unclear how we should handle striping at the different levels. Should only leaves be striped? Or should leaves, parents, grandfolks, etc be colored differently? I experimented with striping but it got weird pretty quickly!

Yep, I'm all for letting alignment lead!

dmfalke (approved):

PR #118, opened 2023-04-25
connect points in lineplot Viz

moontrip:
This will resolve https://github.com/VEuPathDB/web-eda/issues/1519. It was previously addressed but broken due to https://github.com/VEuPathDB/web-eda/issues/1082, thus reimplemented. Screenshot is attached here:
 
- the same example listed in the ticket is used
1411 lineplot connect line01


chowington (approved):
Easy and effective!

moontrip:
Thanks @chowington for your reviews and tests 👍

Issue #108, opened 2023-04-24
Address scrollbar rendering bugs


adnauseum:
Reopened after https://github.com/VEuPathDB/web-monorepo/pull/87 was accidentally merged - still need to address extra unwanted scrollbars - see screenshot in first comment.

This has been done already in the first incarnation:
<s>Please also add a horizontal scrollbar to the entity diagram in SAM (now in the filter menu item)

There is already horizontal scroll in the EDA so this should be simple?</s>

adnauseum:
Example from Windows: 

![Image](https://user-images.githubusercontent.com/24446573/234092366-0168bcd5-09af-4248-93ee-eeb293e387a9.png)

dmfalke:
I'm fairly certain that there are several `overflow: scroll` styling rules that should change to `overflow: auto`, and that will likely address the issue.

dmfalke:
I did not mean to close this!

bobular:
Reopening and editing to re-focus on just the unwanted scrollbars

dmfalke:
@bobular This appears to be fixed. Can you double check and close if it is?

dmfalke:
I'm pretty sure this has all been addressed

Issue #103, opened 2023-04-23
Adjust the default initial position of draggable plots to be further from the lefthand menu


steve-fischer-200:
We want to clearly indicate to user that the floaters are floating.

PR #119, opened 2023-04-25
Fix scrollbar UI bugs in map components

jernestmyers:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/108 and #103 

Screenshots:
image
image


packages/libs/components/src/EntityDiagram/EntityDiagram.tsx
```diff
@@ -8,6 +8,7 @@ import {
 } from '@visx/hierarchy/lib/types';
 import { Line } from '@visx/shape';
 import { LinearGradient } from '@visx/gradient';
+import './diagram.css';
```
jernestmyers, line 11: This is the only non-formatting change in this file. After [fixing the CSS issues in `components`](https://github.com/VEuPathDB/web-monorepo/pull/122), we can now just let `EntityDiagram` set its own overflow property. I therefore [removed the overflow property being applied to the `.Entities` class](https://github.com/VEuPathDB/web-monorepo/pull/119/files?diff=split&w=1#diff-51e890452fe641482125521d5bae2e9d3a573a7c44c59f492c2a605ba347b992L162).

packages/libs/eda/src/lib/map/analysis/DraggableVisualization.tsx
```diff
@@ -63,7 +63,7 @@ export default function DraggableVisualization({
           confineToParentContainer
           showPanelTitle
           isOpen
-          styleOverrides={{ zIndex: 10 }}
+          styleOverrides={{ zIndex: 10, resize: 'both' }}
```
jernestmyers, line 66: It made more sense to me to take advantage of the `DraggablePanel`'s API that can accept a resize style override and apply it to the panel itself.

adnauseum (approved):
🚀 

Thank you!! 🏆 🙇

PR #122, opened 2023-04-26
Fix build-npm-modules to include CSS files and clean up imports/files

jernestmyers:
While working on #108 I noticed that style classes were not being applied to the entity diagram. Upon further inspection, other CSS files were being ignored.

The issue was that the CSS files were not being copied into `lib` during the `build-npm-modules` script. This PR resolves that and then does some additional clean up.


dmfalke (approved):
The build succeeded and the styles are applied on my local clinepi site, so :rocket: !

PR #126, opened 2023-04-27
Base `key` value on other filters to prevent unmount on selection

dmfalke:
This PR addresses the `gd.emit is not a function` error when selecting a subset in an eda histogram filter.


bobular:
I haven't tested yet, but quick question... I'm assuming the `key` was used here as a render refresh "hack"?  It's not necessary as part of a grid or list, as far as I can tell.  Ah, I see there's a comment to this effect!
https://github.com/VEuPathDB/web-monorepo/blob/8062935dfe3e504edf9bf2a152cb304caa95ef69/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx#L329-L330

I'll test carefully!

dmfalke:
Thanks! To be honest, that comment isn't super clear. If it's referring to the "viewport" range, I don't know why we would want it to reset. Also, changing the `key` prop wouldn't impact that.

I wonder if it was something related to an early implementation that is no longer relevant. It's possible we don't even need this `key` prop, but the current change minimizes the testing surface area.

bobular (approved):
Nice catch!

Tested as many scenarios as I could. Draw filter, remove it with cancel button or chip [x] - change filter - etc.  All worked fine.

PR #128, opened 2023-04-27
Prevent MUI elements from having all caps

chowington:
Fixes https://github.com/VEuPathDB/web-eda/issues/1683

### DIFF
--- a/packages/libs/eda/src/index.tsx
+++ b/packages/libs/eda/src/index.tsx
@@ -19,6 +19,10 @@ import React, {
 } from 'react';
 
 import { partial } from 'lodash';
+import {
+  createTheme as createMUITheme,
+  ThemeProvider as MUIThemeProvider,
+} from '@material-ui/core';
 
 import {
   initialize,
@@ -43,7 +47,8 @@ import reportWebVitals from './reportWebVitals';
 import Header from './Header';
 import MapApp from './lib/map';
 import WorkspaceApp from './lib/workspace';
-import UIThemeProvider from '@veupathdb/coreui/dist/components/theming/UIThemeProvider';
+import CoreUIThemeProvider from '@veupathdb/coreui/dist/components/theming/UIThemeProvider';
+import { workspaceThemeOptions as MUIThemeOptions } from './workspaceTheme';
 
 // Hooks
 import { useAttemptActionClickHandler } from '@veupathdb/study-data-access/lib/data-restriction/dataRestrictionHooks';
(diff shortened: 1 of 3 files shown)


packages/libs/eda/src/index.tsx
```diff
@@ -43,7 +47,8 @@ import reportWebVitals from './reportWebVitals';
 import Header from './Header';
 import MapApp from './lib/map';
 import WorkspaceApp from './lib/workspace';
-import UIThemeProvider from '@veupathdb/coreui/dist/components/theming/UIThemeProvider';
+import CoreUIThemeProvider from '@veupathdb/coreui/dist/components/theming/UIThemeProvider';
```
bobular, line 50: Nice. 

Note to self: This is just a rename for clarity, and to disambiguate from MUI theme stuff.

bobular (approved):
This looks great (I used the `?w=1` github trick)

Does the new wrapping at the root 'index.tsx' level mean that we have to do similar things (rename (Core)UIThemeProvider and add MUIThemeProvider) for the website deployments e.g.
`packages/sites/genomics-site/webapp/wdkCustomization/js/client/componentWrappers.jsx` ?

I think so? New ticket or should we do it in this PR? I have had problems setting up a dev website so can't test easily, but will bring that up on Slack.

dmfalke:
> This looks great (I used the `?w=1` github trick)
> 
> Does the new wrapping at the root 'index.tsx' level mean that we have to do similar things (rename (Core)UIThemeProvider and add MUIThemeProvider) for the website deployments e.g. `packages/sites/genomics-site/webapp/wdkCustomization/js/client/componentWrappers.jsx` ?
> 
> I think so? New ticket or should we do it in this PR? I have had problems setting up a dev website so can't test easily, but will bring that up on Slack.

You should be able to test websites by running

```
yarn nx start @veupathdb/clinepi-site
```

Just make sure you have copied `packages/sites/clinepi-site/.env.sample` to `packages/sites/clinepi-site/.env`, and set appropriate values.

dmfalke:
I'm still seeing plot control headings with all caps:
image

This is when running
```
yarn nx start @veupathdb/clinepi-site
```

dmfalke (changes requested):
The theme provider stuff needs to also be applied to `packages/libs/eda/src/lib/map/MapVeuContainer.tsx`

chowington:
> The theme provider stuff needs to also be applied to `packages/libs/eda/src/lib/map/MapVeuContainer.tsx`

Why does it need to be? I thought `MapVeuContainer` is rendered within the theme providers in `index.tsx`

dmfalke:
> Why does it need to be? I thought `MapVeuContainer` is rendered within the theme providers in `index.tsx`

`index.tsx` is only used by the local dev server in the eda package. It isn’t used by consumers of the eda package.

dmfalke (approved):

Issue #28, opened 2023-03-29
Investigate styling discrepancy in variable picker Featured Variables between dev and "site" sites.


bobular:
Slack discussion [here](https://epvb.slack.com/archives/C012ZK4Q5CZ/p1679685362244619)

The mission, if you choose to accept it, is to diagnose what's going on here:

What it looks like on dev servers (`packages/lib/eda`) recently:
![image](https://user-images.githubusercontent.com/308639/228613826-87bce74b-c020-41e8-997c-fa60d5741191.png)

What it should look like:
![image](https://user-images.githubusercontent.com/308639/228613903-32e1b622-bb76-4051-a566-ab8de0d205ec.png)

jernestmyers:
Looks to be related to emotion's `css` prop, seeing as the trouble areas both coincide with its use.

Screenshot:
![Image](https://user-images.githubusercontent.com/69446567/234994344-255daafb-31ad-46f3-b5d5-34b873ae5d58.png)

Relevant code:
https://github.com/VEuPathDB/web-monorepo/blob/ce00dd460cb575ffcb98c33adeb9686e11c89bb6/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx#L653-L660
https://github.com/VEuPathDB/web-monorepo/blob/ce00dd460cb575ffcb98c33adeb9686e11c89bb6/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx#L975

The site servers use different webpack configs and babel configs, which is a good indication that a solution lies in configuring babel the same or similarly.

PR #129, opened 2023-04-27
Configure babel for emotion css prop

jernestmyers:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/28


dmfalke:
`base-webpack-config` isn't used when running `yarn nx start @veupathdb/eda`. Long term, I think we should ditch `react-scripts` and find a solution that works for all dev server environments.

dmfalke (approved):

Issue #94, opened 2023-04-21
magnify mode for markers


steve-fischer-200:
Reinstate buttons to allow magnify mode

bobular:
Also reinstate + and - buttons and see related https://github.com/VEuPathDB/web-components/issues/396

Issue #125, opened 2023-04-26
Header and menu styling and layout (map viewport change)


bobular:
From https://veupathdb.atlassian.net/wiki/spaces/WG/pages/255033345/2023-04-24+Map+UX 

### Menu

- [ ] Make the line separating menu items from configuration panels darker / thicker to indicate that the 2 configuration panels are connected to each other

- [x] Close drawer buttons
  - Button to close menu drawer is too small → make bigger 
  - Now that header will no longer be close-able, the button to close the menu drawer can be repositioned → move to the top right of the open menu panel

### Header

- [x] Header should remain open; remove the button to collapse the header - already in this issue --> https://github.com/VEuPathDB/web-monorepo/issues/107 

  - The header contains essential information
  - **The header should NOT hide points in the viewport (going forward, exclude points under the header from the viewport)** 

- [x]  Header can blend into the url bar depending on browser configs 

  - Try using a very light “theme” color (ie, light sage for VectorBase)
  -  Alternatively, try using color borders around the header and menu and keeping the white
  - Do not use transparency as markers under the header will be excluded from the viewport

@asizemore has made [mockups](https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?node-id=716%3A17097&t=L0y5pbzD8Hd3qTQ1-1)  - not sure what the consensus is!

asizemore:
I think people generally liked the light background best. We don't have mockups for a larger tab explicitly, though there were some discussions and ideas in the mockups (some are [here](https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?type=design&node-id=283-1128&t=KiRVokzHji2nFdD5-0), and others scattered around in there). If you'd like something in particular let me know!

jernestmyers:
**Note**: there is [a separate issue to maintain visibility of the map header](https://github.com/VEuPathDB/web-monorepo/issues/107).

Additionally, as stated [in the PR](https://github.com/VEuPathDB/web-monorepo/pull/127):
> Also related to https://github.com/VEuPathDB/web-monorepo/issues/125. Note that this does not address the issue of hiding data beneath the header.

PR #137, opened 2023-05-02
positioning mouseTool and creating custom zoom control

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/94 and https://github.com/VEuPathDB/web-components/issues/396 as they are closely related. A screenshot is attached in the end. 

customZoom

Some notes are made for helping the reviews in the following. 

- react-leaflet zoomControl component (even leaflet's zoom control) has an inherent limitation to arbitrarily position the zoom control: only four corners (e.g., top left, top right) can be used but due to the top header of the SAM and other floating Vizs, it was necessary to make a custom zoom control.
- four zoom-related controls are made: zoom-in, zoom-out, zoom to whole world, and zoom to data. 
- FontAwesome is used for zoom to whole world and zoom to data icons because MUI icons suggested by Dave were hard to position correctly
- zoom to whole world: Bob's suggestion using center: [1e8, 1e8] worked for MapViz, however, it seems that the value is too small for SAM (full screen mode) and map does not recognize it correctly. Thus, I increased it to [0.25, 0.25] which is visibly not that much difference from center: [0, 0]. 
- zoom to data: I tried various approaches but settled down using center: [0,0], which will initiate flyTo function. However, some considerations were made: 
a) SAM does not support flyTo yet, thus I could not test zoom to data in there: if necessary, I can work for enabling it (see below), Bob?
b) Thus, MapViz is utilized to test the zoom to data. However, I noticed that flyTo was executed much faster than before full data loading in this case, To avoid this issue, I made additional condition to perform the flyTo only if whole data is visible/loaded. This effectively prevents from running flyTo too quickly without interrupting the original flyTo functionality.
c) As mentioned above, I had to make possibly unnecessary codes for testing zoom to data feature at Mapviz, which are marked as typical(?) DKDK for now :). Not quite sure if we want to use this custom zoom control to the MapViz as well (Bob?). Either way, those DKDK will be treated later.


bobular (changes requested):
Looking good!

A general point is that when https://github.com/VEuPathDB/web-monorepo/issues/125 is done (cc: @jernestmyers ) there will be no need to adjust positioning of controls to take account of the header.

Therefore I strongly recommend that the MapVEuMap should behave identically for SAM or EDA viz deployments. In other words remove the `isStandAloneMap` props.

I don't think we need the `isEqual(completeCasesAllVars, markerSum)` condition in the flyTo logic.  But if we do, please explain why.  We can surely remove this and the `markerSum` code (which will fail with string values `'2k'` etc. I guess).

moontrip:
Thank you for your comments, @bobular. I added my opinion next to your comment in the following (starting with DK):

> Looking good!
> 
> A general point is that when #125 is done (cc: @jernestmyers ) there will be no need to adjust positioning of controls to take account of the header.

> Therefore I strongly recommend that the MapVEuMap should behave identically for SAM or EDA viz deployments. In other words remove the `isStandAloneMap` props.

**DK:** Yes I know that work by Jeremy. That being said, we still need to adjust some because both mouse mod/tool and zoom control are now customized. For the isSAM prop, if I remember correctly, the small map of the the full screen map (FSM) (left to the entity diagram) will be affected without having a specified prop. It could be a pre-existing prop instead of isSam but we will see. 

> 
> I don't think we need the `isEqual(completeCasesAllVars, markerSum)` condition in the flyTo logic. But if we do, please explain why. We can surely remove this and the `markerSum` code (which will fail with string values `'2k'` etc. I guess).

**DK:** As I mentioned at my initial description of this PR (perhaps TL;DR ? 😄 ) it is for functioning zoom to data. Without the additional condition, flyTo is performed faster than full data loading. It works fine if all markers are visible, but if no or partial markers are visible, then flyTo is performed with none or limited markers which lead to incorrect behavior. 
That markerSum can be computed inside useMapMarkers hook before converting large numer to k (e.g., '2k'), I suppose.  It is a temporary one as I mentioned so I think that we can adjust it if necessary.

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -462,3 +430,139 @@ function MapScrollWheelZoom(props: MapScrollWheelZoomProps) {
...
+  const zoomToWholeWorld = (e: React.SyntheticEvent) => {
+    e.preventDefault();
+    // for SAM (or full screen mode), very small number does not work: 0.25, 0.25 is ok
+    map.setView([0.25, 0.25], 2);
+  };
+
+  // zoom to data: using flyTo function implicitly
+  const zoomToData = (e: React.SyntheticEvent) => {
+    e.preventDefault();
+    map.setView([0, 0], 2);
```
bobular, line 461: This will need to be not-hardcoded...

Somehow!

moontrip: @bobular yes you are right. That's what I am thinking too. perhaps passing a prop like default viewport to the MapVEuMap as the default is pre-defined at MapViz/SAM (not sure of FSM but I guess it is?).

dmfalke: > Sounds good. We really need to retire the FSM ASAP. It's causing maintenance overhead already...

I'm happy to do this in a separate PR

moontrip:
@bobular I think that I addressed your feedbacks all for both MapViz and SAM: a) defaultViewport is used for setting zoom-to-data. I did change the default zoom of the SAM from 4 to 2 (same to MapViz). I guess this is alright owing to flyTo functionality? b) SAM now has flyTo as well as zoom-to-data as well; c) adjusted positions considering the not-overlapping-banner. 

Actually there is an redundant part at MouseTool because of its use at FSM: positioning MouseTool. If FSM will be retired, then some simplification can be further made in the future.

packages/libs/eda/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
```diff
@@ -274,7 +265,15 @@ function MapViz(props: VisualizationProps<Options>) {
...
         mouseMode={vizConfig.mouseMode ?? createDefaultConfig().mouseMode}
         onMouseModeChange={onMouseModeChange}
-        //DKDK: temporarily set this for testing custom zoom control at MapViz
+        // pass defaultViewport
+        defaultViewport={{
+          center: [
+            createDefaultConfig().mapCenterAndZoom.latitude,
+            createDefaultConfig().mapCenterAndZoom.longitude,
+          ],
+          zoom: createDefaultConfig().mapCenterAndZoom.zoomLevel,
```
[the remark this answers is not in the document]
dmfalke: Hmm... I think we can do one better. I don't see a reason have a `createDefaultConfig()` function at all. It produces the same output every time, so we could just have the following, instead of https://github.com/VEuPathDB/web-monorepo/blob/b62b7049b5f71db6b61ec71a83ad322cd7b23c0e/packages/libs/eda/src/lib/core/components/visualizations/implementations/MapVisualization.tsx#L53-L64:

```typescript
const defaultConfig = {
  mapCenterAndZoom: {
    latitude: 0,
    longitude: 0,
    zoomLevel: 2,
  },
  baseLayer: 'Street',
  dependentAxisLogScale: false,
  mouseMode: 'default',
} as const;
```

bobular (changes requested):
So to summarise - please remove `isStandaloneMap` - thanks!

packages/libs/eda/src/lib/core/components/visualizations/implementations/MapVisualization.tsx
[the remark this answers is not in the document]
dmfalke: Ah, thank you for reminding me of that. I just saw that it wasn't exported where it was defined, but forgot to check the rest of the file. The `useMemo` approach would be great!

moontrip:
@bobular As I mentioned earlier, if we do not consider FSM, then I can remove that prop isStandAloneMap(): FSM's mouseTool is overlapped with close icon (x). 

By the way, at least for my case, I found the reason why zoom: 0 does not work for flyTo. When I checked viewport after performing zoom-to-data, the appState.viewport.zoom does not change to zero, but becomes 1 thus flyTo does not happen. I am not quite sure if it is due to my 4k resolution but it may be possible as I experienced an issue of viewport center for map's full-screen mode like SAM (in the case of zoom-to-whole-world)

moontrip:
@bobular Thanks 👍 Yes I will also look at it. What I also noticed is that I can't zoom out to zero when using zoom-out icon: the minimum zoom level I can reach is 1. So, perhaps it is related to my resolution.

moontrip:
@bobular addressed your feedbacks. Temporarily default zoom level of the SAM was set to be 1. Will test it more.

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -347,6 +349,8 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
       />
       {/* set ScrollWheelZoom */}
       <MapScrollWheelZoom scrollingEnabled={scrollingEnabled} />
+      {/* use custom zoom control if SAM */}
```
bobular, line 352: no need for "if SAM" in comment ;-)

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -476,3 +442,142 @@ function MapScrollWheelZoom(props: MapScrollWheelZoomProps) {
...
+  const zoomOut = (e: React.SyntheticEvent) => {
+    e.preventDefault();
+    map.setZoom(map.getZoom() - 1);
+  };
+
+  // zoom to whole world
+  const zoomToWholeWorld = (e: React.SyntheticEvent) => {
+    e.preventDefault();
+    // with zoom level 1, then center: [1,1] works
+    map.setView([1, 1], 1);
```
bobular, line 471: (1,1) seems a bit arbitrary, and I think the global view should be the same as the default viewport.

Could we have something like this please?

```
const epsilon = 0.25; // or whatever smallest value works reliably (can we understand why very small values don't work?)
map.setView(props.defaultViewport.center.map((coord) => coord + epsilon), props.defaultViewport.zoom);
```

packages/libs/eda/src/lib/map/analysis/appState.ts
```diff
@@ -62,10 +62,11 @@ export const AppState = t.intersection([
 // eslint-disable-next-line @typescript-eslint/no-redeclare
 export type AppState = t.TypeOf<typeof AppState>;
 
-const defaultAppState: AppState = {
+// export default viewport for custom zoom control
+export const defaultAppState: AppState = {
   viewport: {
     center: [0, 0],
-    zoom: 4,
+    zoom: 1,
```
bobular, line 69: Cool - let's keep this and then investigate zoom zero separately on the other ticket (much less urgent)

bobular (approved):
Hi DK - this looks great, thanks - just one minor request on global view. :-)

bobular:
PS. I think we'll discuss styling these buttons in styling group tomorrow - if you can join, please do!

packages/libs/components/src/map/MapVEuMap.tsx
moontrip: Oops! Thank you for catching it 👍

packages/libs/components/src/map/MapVEuMap.tsx
moontrip: I used the approach, epsilon. Although it looks okay by using map function to sum coord and epsilon, however, it somehow causes type error. Thus, I simply utilized [center[0]+epsilon, center[1]+epsilon]

As I mentioned before, it seems to me that there exists a coordinate resolution per zoom level. For zoom level 1, it looks like 0.703... Thus, to distinguish zoom-to-whole-world from zoom-to-data, epsilon should be larger than 0.703... For this reason, I just used epsilon = 1 considering some margin of error.

bobular: Ah, thank you - I didn't realise it was zoom-level related. Apologies if you explained already...

That 0.703 looks like it could be `1/sqrt(2)`

PR #139, opened 2023-05-02
Use filter selection as vocab for variables

### LINKED ISSUE #138: Filter-aware variable constraints in visualizations
I believe this only affects categorical variables, which can change from high-cardinality to low-cardinality upon filtering down to <=8 values.

When a variable has been filtered down to <=8 values, it should appear as clickable in the overlay variable selector, for example.  Similarly a variable filtered down to <=10 values should be allowed as an x-axis variable for bar or boxplot.

Implementation hint: we probably need to pass `filters` to `InputVariables` and either
1. make a modified version of `entities` (the study metadata) where each variable's `distinctValuesCount` is set from the number of filter values for that variable (if filtering on it, and if it's categorical), and pass this to `filterConstraints` 
2. modify `filterConstraints` to alter `variable.distinctValuesCount` based on filter values when appropriate.

Hope that makes sense. Both are probably equivalent efficiency-wise.

dmfalke:
### Description

Closes #138 

Variable constraints currently operation on dataset-wide characteristics of a variable. For example, if a variable has > 8 values, it will not be available in many contexts of a visualization.

This PR makes it so that if the subset is filtered by a categorical variable, then the filter selection is used in place of the dataset-wide variable's vocabulary. This makes it possible to select <= 8 values, and then use the variable in visualizations.

### Notes
- All inputs are are filter sensitive wrt `{min,max}NumValues`
- Legend labels are constrained to filter selection
- Plot ticks are constrained to filter selection
- Introduced pattern using `useDeepValue` to eliminate eslint warnings re vizConfig properties in visualizations.

### Testing notes
- Check that variable inputs and legends respond to filter changes
- Check UX of when a filter is removed and a selected variable is no longer allowed
- Check standalone map


dmfalke:
@bobular ready for your review! Feel free to add others, as you see fit.

packages/libs/eda/src/lib/core/utils/visualization.ts
```diff
@@ -188,6 +195,43 @@ export function variablesAreUnique(vars: (Variable | undefined)[]): boolean {
...
+    return (
+      inputSelection &&
+      disabledVariables.some((disabledVariable) =>
+        isEqual(disabledVariable, inputSelection)
+      )
+    );
+  });
+  if (invalidInputs.length) {
+    throw new Error(
+      `The following variables are invalid and must be changed: ${invalidInputs
```
bobular, line 228: If possible can we say "have become invalid"?

bobular (approved):
This is great. I approve and would appreciate it merged ASAP (big overlap with my branch when I start wiring each floating visualisation)

There's one thing I would add a new ticket for:

If I have a visualization on screen, overlaid with a filtered categorical variable, for example, and I `(x)` the filter chip, I get the pink message - great. Then a second later a data request is made, pink message disappears, spinner spins, and the pink message comes back when the request is complete.  It's not a viz plugin request (e.g. `histogram`) it must be something to do with the filtered counts still pending.  I think definitely a new ticket if not an easy fix.

dmfalke:
I'll go ahead and merge. A new ticket would be great. We can see who makes it first 🙂. 

Thanks for the review! Merging now.

PR #141, opened 2023-05-02
set max-width for SAM legend

moontrip:
This will resolve https://github.com/VEuPathDB/EdaNewIssues/issues/632. I couldn't test it as WWARN is not available yet, but tested with smaller legend width (200px) temporarily to check if ellipsis works fine (see below screenshot). Currently the max-width is set to be 400px: same with Viz;s legend width.

test legend width 200


dmfalke (approved):

moontrip:
Thanks @dmfalke for your review. 👍

PR #144, opened 2023-05-03
add flyTo for SAM

moontrip:
This is for enabling flyTo functionality for stand alone map (SAM). As mentioned in https://github.com/VEuPathDB/web-monorepo/pull/137, I intend to use flyTo for zoom to data function.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -21,7 +21,7 @@ import { Download, FilledButton, Filter } from '@veupathdb/coreui';
 import { useEntityCounts } from '../../core/hooks/entityCounts';
 import ShowHideVariableContextProvider from '../../core/utils/show-hide-variable-context';
 import { MapLegend } from './MapLegend';
-import { AppState, useAppState } from './appState';
+import { AppState, useAppState, defaultAppState } from './appState';
 import { FloatingDiv } from './FloatingDiv';
 import Subsetting from '../../workspace/Subsetting';
 import { findFirstVariable } from '../../workspace/Utils';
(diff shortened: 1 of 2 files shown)


moontrip:
@d-callan (and @bobular just in case), may I ask one thing about the new backend response for the stand alone map (SAM)? (map-markers) In the MapViz, the backend response for the map-markers consists of mapElements and config.completeCasesGeoVar (total count), but SAM only returns mapElements. Is it simple to include config.completeCasesGeoVar for SAM's response as well? If so, it would be great if you can add it because it is necessary for flyTo (and zoom to data feature that I am working on).

d-callan:
not sure i understand this request tbh.. @bobular are you saying i shouldnt worry about it?

bobular:
Not sure yet @d-callan - the reason DK wanted those numbers was to control the flyTo behaviour. I'm hoping we can find another way.

d-callan:
I don't understand yet how they help to control the flyto behavior..

moontrip:
> I don't understand yet how they help to control the flyto behavior..

@danicahelb A new functionality, "zoom to data" relies on flyTo behavior. A problem is that flyTo is executed before full data loading if partial markers are existent/visible. To prevent it, comparing between total count to total visible count of markers is one way, which make it work the flyTo properly from my test. I hope this can help you understand the reasoning 😃

bobular:
Follow-up: in my testing, variables with missing data didn't produce the appropriate "total visible count of markers" so flyTo was infinitely delayed in some cases.

We have some other options we are looking into.

d-callan:
Thanks 👍 very helpful.

moontrip:
This is now handled at https://github.com/VEuPathDB/web-monorepo/pull/137 so closing

PR #149, opened 2023-05-03
Change overflow from scroll to auto

jernestmyers:
Trivial change to genomics site headers.

This is the only non-formatting change: https://github.com/VEuPathDB/web-monorepo/pull/149/files#diff-56acfe5c679c34adeeee54984d362e4a3b304afb5c23159c474de6da38553f05R183

Before:
image

After:
image


dmfalke (approved):
:fire:  :rocket:

Issue #147, opened 2023-05-03
Map: (Urgent for demo) Dont force axis through 0 if the term has an annotated display range


d-callan:
We are finding terms representing years being forced through 0. This is undesirable, and needs to be rectified prior to some imminent demos. 

Previously we introduced a rule that continuous axes must go through 0, and so somewhere there is some logic to find the min of things like 0, display range, range. This logic should be updated so that annotated display ranges have priority to 0, even if above 0.

dmfalke:
The study metadata changed at some point.

What used to be:
```json
{
  "distributionDefaults": {
    "displayRangeMin": 2012,
    "displayRangeMax": 2015
  }
}
```

is now:
```json
{
  "distributionDefaults": {
    "rangeMin": 2012,
    "rangeMax": 2015
  }
}
```

dmfalke:
Oh, no. I take that back! `rangeMin` and `rangeMax` have always been there.

dmfalke:
I think the issue is that `displayRangeMin` and `displayRangeMax` are not being populated

moontrip:
From my quick observation, the issue is that we set 0 as one of min: there is a condition but will need to look into it further.

dmfalke:
@moontrip I found a case that includes `displayRangeMin` and `displayRangeMax`:

Study: GEMS1A Case Control
Variable: Enrollment year

moontrip:
Hi @d-callan and @danicahelb, a fix is now merged into the main branch so it would be reflected when the site is updated. Some screenshots can be seen at https://github.com/VEuPathDB/web-monorepo/pull/150#issue-1694609625

dmfalke:
The issue isn't quite resolved (see https://github.com/VEuPathDB/web-monorepo/pull/150#issuecomment-1534530355). I'm reopening the issue.

moontrip:
Hi @dmfalke & @d-callan 
First of all, thank you for updating the work @dmfalke. 
I am curious if it is okay for the cases where the displayMin does not exist. It is probably rare, but I could see that Amazonia ICEMR Brazil cohort has the edge case, and its histogram at Browse and Subset look like this. Dave also mentioned about a case https://github.com/VEuPathDB/web-monorepo/issues/147#issuecomment-1533306251. Fixing metadata is one option and the other one is to temporarily check both displayMin and rangeMin?

[...]

PR #150, opened 2023-05-03
remove-zero-baseline-consideration

moontrip:
This will resolve https://github.com/VEuPathDB/web-monorepo/issues/147. I have tested several cases, but one of the best and simplest way is to set the default value of the isNonZeroBaseline prop as true for. Actually, it is not used anywhere yet.

Here are various screenshots using year variable after applying this:

year variable 01

year variable 02

year variable 03

year variable 04

### DIFF
--- a/packages/libs/eda/src/lib/core/hooks/computeDefaultAxisRange.ts
+++ b/packages/libs/eda/src/lib/core/hooks/computeDefaultAxisRange.ts
@@ -24,7 +24,8 @@ export function useDefaultAxisRange(
   logScale?: boolean,
   axisRangeSpec = 'Full',
   // check non-zero baseline for continuous overlay variable
-  isNonZeroBaseline: boolean = false
+  // but not used in any Viz yet, so change the default to true
+  isNonZeroBaseline: boolean = true
 ): NumberOrDateRange | undefined {
   const defaultAxisRange = useMemo(() => {
     // Check here to make sure number ranges (min, minPos, max) came with number variables
--- a/packages/libs/eda/src/lib/core/hooks/computeDefaultAxisRange.ts
+++ b/packages/libs/eda/src/lib/core/hooks/computeDefaultAxisRange.ts
@@ -22,10 +22,7 @@ export function useDefaultAxisRange(
   max?: number | string,
   /** are we using a log scale */
   logScale?: boolean,
-  axisRangeSpec = 'Full',
-  // check non-zero baseline for continuous overlay variable
-  // but not used in any Viz yet, so change the default to true
-  isNonZeroBaseline: boolean = true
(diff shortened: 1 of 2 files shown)


packages/libs/eda/src/lib/core/hooks/computeDefaultAxisRange.ts
```diff
@@ -24,7 +24,8 @@ export function useDefaultAxisRange(
   logScale?: boolean,
   axisRangeSpec = 'Full',
   // check non-zero baseline for continuous overlay variable
-  isNonZeroBaseline: boolean = false
+  // but not used in any Viz yet, so change the default to true
+  isNonZeroBaseline: boolean = true
```
dmfalke, line 28: I don't really understand this parameter. Can you add a comment explaining what "non-zero baseline" means? Thanks!

dmfalke (changes requested):

packages/libs/eda/src/lib/core/hooks/computeDefaultAxisRange.ts
moontrip: @dmfalke That's a good question. The isNonZeroBaseline prop was intended to deal with setting a range of continuous variable for colormap legend: https://github.com/VEuPathDB/web-components/issues/454. Initially, the range was designed to be computed via the useDefaultAxisRange hook. However, I reminded myself that overall procedure was changed without using the hook, thus the prop is actually no longer used. Thus, I removed the prop and changed accordingly. To address the issue of year variable, I simply removed zero (0) case from the calculation of the min at numberDateDefaultAxisRange(). These changes are committed.

Thank you for inspiring me to recall this 👍

dmfalke: Thanks for the explanation! This looks great.

dmfalke (approved):

packages/libs/eda/src/lib/core/hooks/computeDefaultAxisRange.ts
moontrip: @dmfalke my pleasure 😄 Thank you for your reviews!

d-callan:
Does this now never force through 0? I'm not sure I'm reading it right..

dmfalke:
> Does this now never force through 0? I'm not sure I'm reading it right..

I think you're right. It seems to me, this can be fixed by changing this line:
https://github.com/VEuPathDB/web-monorepo/blob/e8f4d51eb12dfc120a17d7fc84a803ad9a3c9314/packages/libs/eda/src/lib/core/utils/default-axis-range.ts#L34

to
```
defaults.displayRangeMin ?? 0,
```

PR #152, opened 2023-05-04
Fix ortho useRef error

dmfalke:
This PR addresses the `useRef is not defined` error on group record pages. It also fixes an issue with a missing favicon file for local dev sites.


aurreco-uga (approved):
thanks!

PR #153, opened 2023-05-04
Fix default range min logic for numeric vars

### LINKED ISSUE #147: Map: (Urgent for demo) Dont force axis through 0 if the term has an annotated display range

dmfalke:
fixes #147

This PR further addresses issues seen with default ranges for continuous variables.


dmfalke:
cc @d-callan @moontrip

d-callan (approved):

PR #157, opened 2023-05-04
Strip HTML from document title

jernestmyers:
Minor nit: some ClinEpi studies include HTML in their `displayName` properties, which then gets displayed in the browser tab due to the `document.title` logic we use. This aims to fix that by stripping HTML from the string used for `document.title`.

Before:
image

After:
image


dmfalke (approved):
Nice!

PR #160, opened 2023-05-05
swap bar marker svg colors

asizemore:
Moved the bar colors around so they're looking a lot better now. Previously had the lightest on the right end which was terrible. Here's the new version:

Screen Shot 2023-05-05 at 4 51 16 PM

My latest concern - is the lightest bar too light? It's the same color as the lightest donut wedge, but that lightest donut wedge touches dark colors instead of being totally surrounded by white so it's not so jarring. It's possible i'm overthinking this... so let me know if it also bothers you.


adnauseum (approved):
Yo. You gotta cool it on the self hate here. Listen to the scathing comments: "they're looking a lot better now" or "which was terrible".

These icons are dope and look so good!!! 

Now, to address your latest concern, I don't think it's a huge accessibility issue, but the contrast ratio between the lightest bar and the white background fails to meet the WCAG spec. https://webaim.org/resources/contrastchecker/?fcolor=DBD9D9&bcolor=FFFFFF

It's not mission critical though! And that WCAG stuff **is designed for text more than parts of icons**. My advice: it could be a hair darker! 🪄

asizemore:
Lol fair. Also my comments didn't really specify the issues with the previous one at all! How will we remember why we changed the color order!??! 
Here we go:

The previous bar plot svg looked like this
Screen Shot 2023-05-08 at 7 52 55 AM

There are two reasons to update this bar
1. The colors go dark to light here but the actual markers should run light to dark (assuming continuous variable)
2. Dark to light with the light end dangling over the white background with no visual bounding box made the marker feel like it was just falling off or blending into the background which is not what we want - the plot should have a defined beginning and end that should be self evident.

First, i tried just swapping so that the bars go light to dark, just like the markers when a continuous var is chosen. That still felt off, so i ended up mimicking the categorical version of bar markers and swapped around the lighter and darker bars. 


And thank you for the advice on the colors! Yes i think a hair darker is good. Updated in the latest commit 😄

asizemore:
Aaanannddd here's the new version!

Screen Shot 2023-05-08 at 2 19 05 PM

PR #177, opened 2023-05-09
Change SAM's default zoom level to zero

### LINKED ISSUE #176: Default zoom level should be zero, not 4
Ideally, all data for a study or megastudy should appear in the user's initial viewport, no matter how big their screen is.

This is because the default behaviour of the map is to "fly to" the best-fit view of the marker data shown in the initial default viewport.

For some reason the default is 4, but zero should work better.

bobular:
Resolves #176 

Related to #144

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/appState.ts
+++ b/packages/libs/eda/src/lib/map/analysis/appState.ts
@@ -65,7 +65,7 @@ export type AppState = t.TypeOf<typeof AppState>;
 const defaultAppState: AppState = {
   viewport: {
     center: [0, 0],
-    zoom: 4,
+    zoom: 0,
   },
   mouseMode: 'default',
   activeMarkerConfigurationType: 'pie',


moontrip (commented):
Hi @bobular Actually I left a comment https://github.com/VEuPathDB/web-monorepo/issues/176#issuecomment-1540381056.

Another comment is that zoom: 0 is too small for me perhaps because I am using 4k monitor. 

I briefly zoom: 0 with the custom zoom control, but somehow zoom to data does not work in that case. It works from zoom: 1. I don't know why right now.

- zoom: 0
zoom-zero

bobular:
No 4k monitor here, but testing with zoom=0 in https://github.com/VEuPathDB/web-monorepo/pull/137 works fine for me (dfalke-b back end and SCORE S. mansoni).

Can you describe what happens on your screen?

bobular:
PS. I'm happy to close this without merging if you make the change in your branch.

moontrip:
> No 4k monitor here, but testing with zoom=0 in #137 works fine for me (dfalke-b back end and SCORE S. mansoni).
> 
> Can you describe what happens on your screen?

@bobular When I pressed zoom-to-data icon, it went to [0, 0] with zoom of 0 and did nothing since then. FlyTo works when zoom >= 1

bobular:
Converted to draft. Revisit and close/fix as necessary.

bobular:
Closed as won't fix

PR #178, opened 2023-05-09
Use minmax for varTree column in subsetting UI

jernestmyers:
Variable tree gets squished more aggressively in the map context. Switched the column width to use `minmax` to limit how small the width of the column gets.

Before:
image

After:
image


dmfalke (approved):
Nice! :rocket:

PR #180, opened 2023-05-09
Map: saved analyses and downloads

### LINKED ISSUE #89: Add Map app menu items to primary website header
We will need some menu items to link to different studies in the new Map App. We need to decide what those are.

Steve's suggestions:
```
Map: Full set of vector collections
Map: Updated Maine Tick collection
Map: Updated New York collection
```

### LINKED ISSUE #140: Standalone map: map state/analyses management panel
This is effectively a small "My Analyses" panel but it has some slightly different functionality.

It's a list of saved analyses for the current study (and no way to switch to other studies, study name should not be displayed).

Above the list is the editable name of the current analysis, and a "new analysis" button.

Clicking on an item in the list will switch to that saved analysis.
The list could also have delete buttons.

The "do you want to rename?" existing EDA functionality would be good to have on the "new analysis" button, but not essential right now. Make a new ticket if it's not trivial to do.

### LINKED ISSUE #171: Map: Download panel should show subset downloads, even if files aren't available
Currently, the panel shows an error message.

### LINKED ISSUE #173: Map: Study details adjustments
- [ ] font size is out of scale to rest of SAM
- [ ] offers unexpected WDK buttons up top (Add to Basket)

dmfalke:
closes #89
closes #140
closes #171
closes #173

This PR addresses several issues. Sorry about that. I can break it up into multiple PRs, if needed.

Of primary concern is the newly added **My Analyses** entry:
image

### Open questions

- [ ] Are we okay with the "My Analyses" verbiage?
- [ ] Do we need to offer some explanation of what that table is, why it's useful?
- [x] We are currently displaying the active analysis in the table. Should this be highlighted somehow, or filtered out?
- [ ] I haven't added **New Analysis** button, yet. I was wondering if that should be in the header, along with "make a copy" and "delete"?
- [ ] Currently, changes to marker configuration will not cause an unsaved analysis to become saved. Is this okay?

### Remaining TODOs

- [ ] Synchronize analysis list and active analysis. This might involve moving the cache to the AnalysisClient class. It might also involve adding a `subscribe()` method. This will keep changes to either the analysis or the list of analyses up-to-date.
- [ ] There are some issues with changing analysis in the table. The layout doesn't appear to be changing, etc. It might involve clearing app state if the analysis id changes. Or maybe use analysis id as the key for the MapAnalysisImpl component.
- [ ] Need to add loading indicator to the download panel, when counts are updating.
- [ ] Need to have a UX discussion about where to place analysis actions (copy, delete, etc)
- [ ] Close MyAnalyses section when a new one is selected
- [ ] This needs addressing `Showing 2 of 161 analyses` (the user can't get to the other 159)

### DIFF
--- a/packages/libs/eda/src/lib/map/MapVEu.scss
+++ b/packages/libs/eda/src/lib/map/MapVEu.scss
@@ -11,9 +11,19 @@
   .MapVEu {
     height: 100%;
 
+    .wdk-RecordContainer {
+      font-size: 0.9em;
+      margin-top: 1em;
+    }
+
     .wdk-RecordSidebarToggle,
     .wdk-RecordSidebar {
       background-color: #eaeaea;
+    }
+
+    .wdk-RecordActions,
+    .wdk-RecordHeading {
+      display: none;
     }
   }
 }
--- a/packages/libs/eda/src/lib/map/MapVeuContainer.tsx
+++ b/packages/libs/eda/src/lib/map/MapVeuContainer.tsx
@@ -44,10 +44,10 @@ export function MapVeuContainer(mapVeuContainerProps: Props) {
   return (
     <Switch>
       <Route
-        path={`${path}/:studyId/:analysisId`}
+        path={[`${path}/:studyId/new`, `${path}/:studyId/:analysisId`]}
         render={(
           routeProps: RouteComponentProps<{
-            analysisId: string;
+            analysisId?: string;
             studyId: string;
           }>
         ) => (
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -92,6 +92,7 @@ import {
 import { BarPlotMarkers, DonutMarkers } from './MarkerConfiguration/icons';
 import { AllAnalyses } from '../../workspace/AllAnalyses';
 import { getStudyId } from '@veupathdb/study-data-access/lib/shared/studies';
+import { isSavedAnalysis } from '../../core/utils/analysis';
 
 enum MapSideNavItemLabels {
   Download = 'Download',
@@ -730,8 +731,12 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
               }}
             >
               <AllAnalyses
-                // studyId={analysisState.analysis?.studyId!}
                 analysisClient={analysisClient}
+                activeAnalysisId={
+                  isSavedAnalysis(analysisState.analysis)
+                    ? analysisState.analysis.analysisId
+                    : undefined
+                }
                 subsettingClient={subsettingClient}
                 studyId={getStudyId(studyRecord)}
                 showLoginForm={showLoginForm}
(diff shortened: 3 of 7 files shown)


adnauseum:
> Do we need to offer some explanation of what that table is, why it's useful?

I think this is pretty intuitive! 

> We are currently displaying the active analysis in the table. Should this be highlighted somehow, or filtered out?

Wow, this is kinda tricky. I think we should keep it in the table. Some kind of active indicator would be good, but I don't know what. At first I think of was something like this:
image

jernestmyers:
> I haven't added New Analysis button, yet. I was wondering if that should be in the header, along with "make a copy" and "delete"?

I'd be concerned with cluttering up the map header. Perhaps it makes more sense in the My Analyses panel, but perhaps then we may want to change the menu title to something like "Manage Analyses".

jernestmyers:
Another instance of needing loading indicators: changing filters takes significant time to get reflected in the UI, so it's confusing when things do change 20 seconds later.

https://github.com/VEuPathDB/web-monorepo/assets/69446567/3f95e60c-27c0-4b2e-b603-400339d9791f

dmfalke:
> Also, I can put these into a separate ticket, but I encountered two very minor bugaboos.
> 
> Changing title of active analysis via My Analyses section doesn't reflect in the title.
> analysis_name.mp4
> 
> Looks like the links to the analysis are malformed?
> 
> Image

Thanks for catching these! I will fix ASAP.

dmfalke:
> > I haven't added New Analysis button, yet. I was wondering if that should be in the header, along with "make a copy" and "delete"?
> 
> I'd be concerned with cluttering up the map header. Perhaps it makes more sense in the My Analyses panel, but perhaps then we may want to change the menu title to something like "Manage Analyses".

I can make another issue for this. It's not really needed until the public release. We'll need to have a UX discussion, I suppose.

dmfalke:
> Another instance of needing loading indicators: changing filters takes significant time to get reflected in the UI, so it's confusing when things do change 20 seconds later.
> Screen.Recording.2023-05-10.at.5.20.45.PM.mov

Oh, yes. Not sure if this is needed for the demo. I will see how difficult it is to add, and assess.

jernestmyers:
Noticing another bug: switching between analyses isn't applying/rendering the correct zoom level.

dmfalke:
I pushed a change to show the active analysis. I think everything else might need to be done later. I will update the description to indicate what's left to do.

jernestmyers (approved):

adnauseum (approved):
🚀

bobular:
the `Showing 2 of 161 analyses` is not great, especially as the user has no way to get to the ones not listed

I'll add a TO DO checklist

bobular:
In the following, by "layout" do you mean viewport and zoom level?

> There are some issues with changing analysis in the table. The layout doesn't appear to be changing, etc. It might involve clearing app state if the analysis id changes. Or maybe use analysis id as the key for the MapAnalysisImpl component.

bobular:
I know downloads aren't the focus of this PR but I'm getting the following after adding columns to the download preview thingy, and it's very strange because there is no back end request (500 or otherwise) being made. So how can it be a "server error"? I am using dfalke-b.clinepidb.org back end.

image

Issue #169, opened 2023-05-09
Map: Filter panel tweaks


dmfalke:
- in megastudy the entity diagram too wide, so is partially offscreen.  this is a bummer.   we should discuss ways to fix this in an upcoming UX meeting. 
- the entity diagram is scrolled offscreen at the top.
- We use the word Filter to describe the filters, but Subset in heading.   We need to either:
  - use just one, always
  - or add a bold title in the Filter display, indicating that when you filter, you are creating a subset

jernestmyers:
The last two items were addressed in #188

bobular:
Not really a dealbreaker. Another general "shrinking" real-estate issue for the Map.

bobular:
Just added this to the regular EDA board.

This should be discussed as part of wider refactoring proposals. For example filters always in a modal.

PR #188, opened 2023-05-10
Adjust scroll behavior and verbiage in map's filtering UI

jernestmyers:
Addresses the 2nd and 3rd points of #169 
<br>
> (2nd point): the entity diagram is scrolled offscreen at the top.

**Solution**: The `FieldNode` component in the variable tree has some `scrollIntoView` logic. I disabled this logic in the map context.
<br><br>
> (3rd point): We use the word Filter to describe the filters, but Subset in heading. We need to either:
    - use just one, always
    - or add a bold title in the Filter display, indicating that when you filter, you are creating a subset

**Solution**: In the map header, the word 'Subset' has been replaced with 'Filtered' in hopes to better connect that value with the filter panel.

image

### DIFF
--- a/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
+++ b/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
@@ -198,6 +198,9 @@ interface VariableListProps {
   startExpanded?: boolean;
   asDropdown?: boolean;
   dropdownLabel?: string;
+  /**
+   * used to disable FieldNode's scrollIntoView property in map scope
+   */
   scope?: VariableScope;
   clearSelectionButton?: ReactNode;
 }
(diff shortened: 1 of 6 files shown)


packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
```diff
@@ -198,6 +198,9 @@ interface VariableListProps {
   startExpanded?: boolean;
   asDropdown?: boolean;
   dropdownLabel?: string;
+  /**
+   * used to disable FieldNode's scrollIntoView property in map scope
+   */
```
adnauseum, line 203: 🙇

adnauseum (approved):
🚀

adnauseum (approved):
🚀

PR #202, opened 2023-05-11
Make the SAM legend filter-aware

dmfalke:
### Description

This PR makes the legend in the standalone map (SAM) "filter-aware".

Previously, there was some logic to use the variable's vocabulary directly, with no ordering, if its cardinality was smaller than the color palette. Otherwise, the top 7 values of the _entire study_ was used for the legend.

In this PR, that logic has been replaced with always using the top values of the _subset_. If the number of values is larger than the color palette, then an "other" value is included; otherwise, all of the values are used without the "other" value.

In the function `getMostFrequentValues`, I'm passing in `vocabulary`, but this is only used for the "low-cardinality" check. I'm not entirely sure we need that, anymore. @bobular can you comment on the utility of this check, especially given the changes in the PR? Thanks in advance.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -305,7 +305,7 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
       return visualization.withOptions({
         hideFacetInputs: true,
         layoutComponent: FloatingLayout,
-        getOverlayVariable: (_) => selectedVariables,
+        getOverlayVariable: (_) => activeMarkerConfiguration.selectedVariable,
         getOverlayVariableHelp: () =>
           'The overlay variable can be selected via the top-right panel.',
       });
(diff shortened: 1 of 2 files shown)


dmfalke:
> In the function `getMostFrequentValues`, I'm passing in `vocabulary`, but this is only used for the "low-cardinality" check. I'm not entirely sure we need that, anymore. @bobular can you comment on the utility of this check, especially given the changes in the PR? Thanks in advance.

After some additional testing, I have removed this, possibly naively. With the code I removed in place, if the subset filters causes the distribution endpoint to return fewer values than the overlay variable's vocabulary, the error would get thrown, and no markers/legend would be displayed.

I tested this with:
```
Site: https://feature.vectorbase.org
Study: Mega Test 1
Filter: species (top 8 values selected; sort by count, and choose the top 8)
Marker overlay: Sample type
```

Please let me know if this bad.

jernestmyers (approved):
I didn't encounter any bugs in the testing I did.

dmfalke:
> I didn't encounter any bugs in the testing I did.

Thank you!

Issue #197, opened 2023-05-10
Simplify zoom/pan toolbar


asizemore:
Discussed in [styling](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/270663681/Styling+Group+Meeting+May+10+2023) 05/10/23.

- [ ] Remove the globe button
- [ ] Set a max zoom out level that equals what we would have gotten from hitting the globe button. When a user hits max zoom out, the zoom out button (the minus) should become gray
- [ ] Replace the fly to data icon with the [Undo](https://github.com/VEuPathDB/CoreUI/blob/main/src/components/icons/Undo.tsx) icon

PR #206, opened 2023-05-11
styling and adding features on zoom control, change flyTo condition

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/197 and change flyTo condition to improve zoom to data behavior

- [x] Remove the globe button
- [x] Set a max zoom out level that equals what we would have gotten from hitting the globe button. 
- [ ] Disable and grey out zoom buttons when min or max zoom levels are reached (not our `defaultAppState` map zoom level)
- [x] Replace the fly to data icon with the [Undo](https://github.com/VEuPathDB/CoreUI/blob/main/src/components/icons/Undo.tsx) icon
- [x] Change flyTo condition to Improve zoom to data behavior

A screenshot is attached
zoomIcons


asizemore:
Looking at wwarn - when i use the Reset button, I get first zoomed out to the whole world, then markers render with some markers at the very top of the map, then a second render and those markers at the top of the map disappear. 
Is this expected?
Screen Shot 2023-05-15 at 7 42 28 AM

moontrip:
Good morning @asizemore Thank you for your tests and reviews. I didn't check WWARN case but the thing you observed may happen due to a couple of reasons. The internal procedure of Reset zoom (zoom to data) is to move viewport to [0,0] and zoom level 1, load all markers, then flyTo function occurs: flyTo will lead to the smallest map bound with zoom level that contains all markers: e.g., if all markers are located at Africa, then it will zoom into Africa area. 

So, during this process, it may happen that viewport (map center and zoom level) and map tile are already changed, while loading all markers may take some time. So, some markers may be off screen or not shown during data loading, which may be the case what you observed.

I will check WWARN case too.

bobular:
The WWARN thing is very strange. 

There are 6 points at (88.2667, 27.2) - which are almost certainly data provider bugs.

However, they only seem to render at the widest zoom level, when the viewport sent to the back end clearly includes it.
```
{
	"viewport": {
		"latitude": {
			"xMax": 89.4822408119778,
			"xMin": -89.1746315294642
		},
		"longitude": {
			"left": -111.79687498999998,
			"right": -111.79687499999997
		}
	}
}
```

Zoom in one leaflet level (same geohash level) and the viewport latitude stops at 86.77. Just shy of the 88.2 where the data is.

```
{
	"viewport": {
		"latitude": {
			"xMax": 86.77799674310461,
			"xMin": -67.06743335108298
		},
		"longitude": {
			"left": -129.02343749,
			"right": -129.0234375
		}
	}
}
```

It might be a bug in leaflet. I don't think we should spend a lot of time looking into it, unless we see problems like this nearer to the equator!

bobular (changes requested):
Looks good - just one change please.

I say ignore the WWARN-related issue for now.

I'll make a ticket for it so we don't forget.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -850,7 +850,16 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
   // This makes sure that the user sees the global location of the data before the flyTo happens.
   useEffect(() => {
     if (pending) {
-      setWillFlyTo(isEqual(appState.viewport, defaultAppState.viewport));
+      // set a safe margin (epsilon) to perform flyTo correctly due to an issue of map resolution etc.
+      // not necessarily need to use defaultAppState.viewport.center [0, 0] here but used it just in case
+      const epsilon = 2.0;
+      const isWillFlyTo =
+        appState.viewport.zoom === defaultAppState.viewport.zoom &&
+        Math.abs(appState.viewport.center[0]) <=
```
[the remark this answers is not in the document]
moontrip: @bobular Thank you for your comments 👍 I will change it accordingly after doing filter-aware stuff.

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -509,19 +495,15 @@ function CustomZoomControl(props: CustomZoomControlProps) {
           role="button"
           aria-label="Zoom out"
           onClick={zoomOut}
+          style={{
+            color:
+              map.getZoom() === props.defaultViewport?.zoom
```
bobular, line 500: Sorry, a couple of other things came up.

Just greying it out is OK but not ideal.

If the default zoom level was 3, as below, then the button would still function and zoom out to level 2. 

```
export const defaultAppState: AppState = {
  viewport: {
    center: [0, 0],
    zoom: 3,
  },
```

bobular: So I think better to disable the button functionally as well as aesthetically.

Also, we should have the same logic for the zoom **in** button.

I don't think the defaultAppState zoom level should be used for either limit.  We should use the Leaflet `map` object's minzoom and maxzoom properties (assuming these are readable).

bobular: Finally, these Leaflet buttons are "leaky". You can drag or double click on them and it is the same as dragging/double clicking elsewhere in the map. Could also cause problems if there are markers under the buttons.  Could you prevent click events passing through please? (I don't know how, but I note that our custom floaters, like the legend, behave OK.)

bobular (changes requested):
Saw a few more things!

packages/libs/components/src/map/MapVEuMap.tsx
moontrip: @bobular Thank you for your comments and suggestions 👍 I will think about the first two. For the last, actually it is leaflet's default behavior. That is, even the default leaflet's zoom control has the same functionality: you can test it at https://leafletjs.com/examples/quick-start/. If you still want it, I will think about how to override it.

moontrip: @bobular I have addressed all of your feedbacks and they seem to work fine. Can you please check them out? Thanks!

bobular (approved):
Approved assuming my suggestion is incorporated.

Working pretty good (maybe we can punt the not-functionally disabled button issue until it becomes a problem in practice).

I don't really like the default styling of `leaflet-disabled` - the size change of the - and + doesn't look good to me, but we can keep it for now.  See if the styling team object!

moontrip:
@bobular Thank you for your test and reviews The leaflet-disabled class is the default one from leaflet and yes I noticed that they changed the size of the +/-. Perhaps it is acceptable for now as you mentioned. 😃 map.getMaxZoom() seems to work fine 👍

PR #208, opened 2023-05-11
Move legend to right edge and below pointer controls

### LINKED ISSUE #207: Map: move legend to top
It can go just below the pointer mode buttons.

jernestmyers:
Resolves #207 

image

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -972,8 +972,8 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
 
                 <FloatingDiv
                   style={{
-                    top: 350,
-                    right: 50,
+                    top: 250,
+                    right: 8,
                   }}
                 >
                   {legendItems.length > 0 && (


moontrip:
@jernestmyers Just for your information. First, since one of zoom control in your screenshot (global zoom) will be gone (https://github.com/VEuPathDB/web-monorepo/pull/206), the height of the zoom control would be shorter than current screenshot. Second, based on what Bob said in the ticket, it looks like he wants to place legend something like below screenshot, though it is not legend :) - pointer mode that Bob pointed out is the horizontal two icons with arrow and magnifying glass. So, please check with Bob.

235556275-c86a081b-4e13-47b4-b2d8-94888fca3157

jernestmyers:
> @jernestmyers Just for your information. First, since one of zoom control in your screenshot (global zoom) will be gone (#206), the height of the zoom control would be shorter than current screenshot. Second, based on what Bob said in the ticket, it looks like he wants to place legend something like below screenshot, though it is not legend :) - pointer mode that Bob pointed out is the horizontal two icons with arrow and magnifying glass. So, please check with Bob.
> 
> 235556275-c86a081b-4e13-47b4-b2d8-94888fca3157

Thanks for the feedback @moontrip . I probably should have posted in mapveu-ux but I did seek input on the placement in ui-infra. We preferred the right edge and more separation from the other controls. To that end, it's likely not an issue if/when those controls shorten. Ultimately, I think we want the legend to be draggable, at which point we could adjust its default position. 🤷

moontrip:
@jernestmyers Thank you for your detailed information! 👍

adnauseum:
I messed around with an implementation of the draggable panel. I was wondering how easy it'd be to slip into this PR, but there's some considerations we should..... take into.. consideration.

https://github.com/VEuPathDB/web-monorepo/assets/24446573/52fd41d6-e029-457f-8049-b7f1266b69cb

adnauseum (approved):
🚀 👏 Brilliant!

PR #211, opened 2023-05-12
209: prevent interactions with filter from re-rendering the map or exploding the application

### LINKED ISSUE #209: Prevent interactions with filter from re-rendering the map and exploding the application.
https://github.com/VEuPathDB/web-monorepo/assets/24446573/b71bcaaa-c972-4ac5-b019-54a32ee0128f

https://github.com/VEuPathDB/web-monorepo/assets/24446573/f84ebb3b-c470-4862-a24f-40cbab958ecc

☝ 
```
Reduce of empty array with no initial value
TypeError: Reduce of empty array with no initial value
```

We're attempting to `reduce` an empty array without providing an initial value. If `props.data` is an empty array, calling `reduce` on it without an initial value raises this error.

Pictures say more than words.

Image

adnauseum:
Resolves #209.

**Before**:

# Screencast of the map unmounting. ♻️ 
https://github.com/VEuPathDB/web-monorepo/assets/24446573/b71bcaaa-c972-4ac5-b019-54a32ee0128f

# Screencast of the map exploding from a `TypeError` 💥 .
https://github.com/VEuPathDB/web-monorepo/assets/24446573/f84ebb3b-c470-4862-a24f-40cbab958ecc

☝ Here's the stack trace:
```
Reduce of empty array with no initial value
TypeError: Reduce of empty array with no initial value
```

Root cause: we're attempting to `reduce` an empty array without providing an initial value. If `props.data` is an empty array, calling `reduce` on it without an initial value raises this error. The array is empty only very rarely. Perhaps only when the user's thrashing around inputs like a punk. 👊 

Pictures say more than words. Here's the problem:
Image

### DIFF
--- a/packages/libs/components/src/map/DonutMarker.tsx
+++ b/packages/libs/components/src/map/DonutMarker.tsx
@@ -237,7 +237,7 @@ function donutMarkerSVGIcon(props: DonutMarkerStandaloneProps): {
         .map((o) => o.value)
         .reduce((a, c) => {
           return a + c;
-        });
+        }, 0);
 
   // for display, convert large value with k (e.g., 12345 -> 12k): return original value if less than a criterion
   const sumLabel = props.markerLabel ?? String(fullPieValue);
--- a/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
@@ -277,12 +277,13 @@ export function HistogramFilter(props: Props) {
   const updateUIState = useCallback(
     (newUiState: Partial<UIState>) => {
       // if (uiState.binWidth === newUiState.binWidth) return;
-      analysisState.setVariableUISettings({
+      analysisState.setVariableUISettings((currentState) => ({
+        ...currentState,
         [uiStateKey]: {
           ...uiState,
           ...newUiState,
         },
-      });
+      }));
     },
     [analysisState, uiStateKey, uiState]
   );


dmfalke (approved):
LGTM!

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -277,12 +277,13 @@ export function HistogramFilter(props: Props) {
   const updateUIState = useCallback(
     (newUiState: Partial<UIState>) => {
       // if (uiState.binWidth === newUiState.binWidth) return;
-      analysisState.setVariableUISettings({
+      analysisState.setVariableUISettings((currentState) => ({
```
bobular, line 280: Was this basically broken for 2 years?
Ah, yes, it was! In the regular EDA, the subsetting panel could only remember histogram settings (log scale, bin width etc) for one variable. If you made changes to a second variable, the settings of the first would be lost.

bobular (commented):
Just saw this and had a look.

packages/libs/components/src/map/DonutMarker.tsx
```diff
@@ -237,7 +237,7 @@ function donutMarkerSVGIcon(props: DonutMarkerStandaloneProps): {
         .map((o) => o.value)
         .reduce((a, c) => {
           return a + c;
-        });
+        }, 0);
```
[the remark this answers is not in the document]
adnauseum: Nice!

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
dmfalke: Quite the scandal!

PR #213, opened 2023-05-12
Fix header issue in eda subset download table preview

dmfalke:
### Description

Use "height: auto" as default for table header and remove css targeting generated classname. Note that this issue is only happening on feature/qa/prod sites.

ETA: I also added a loading indicator in the site menu entry for the maps.

### Screenshots

**Before**
image

**After**
image

### DIFF
--- a/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/homepage/VEuPathDBHomePage.tsx
+++ b/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/homepage/VEuPathDBHomePage.tsx
@@ -600,7 +600,13 @@ const useHeaderMenuItems = (
           metadata: {
             include: useEda ? [VectorBase] : [],
           },
-          items: mapMenuItems ?? [],
+          items: mapMenuItems ?? [
+            {
+              key: 'maps-loading',
+              type: 'custom',
+              display: <Loading radius={4} />,
+            },
+          ],
         },
         {
           key: 'pubcrawler',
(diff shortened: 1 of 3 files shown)


packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/homepage/VEuPathDBHomePage.tsx
```diff
@@ -600,7 +600,13 @@ const useHeaderMenuItems = (
...
             include: useEda ? [VectorBase] : [],
           },
-          items: mapMenuItems ?? [],
+          items: mapMenuItems ?? [
+            {
+              key: 'maps-loading',
+              type: 'custom',
+              display: <Loading radius={4} />,
+            },
+          ],
```
jernestmyers, line 609: For clarity, is this loading indicator displayed only when the map is clicked from the site header?

jernestmyers (approved):
Looks good!

packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/homepage/VEuPathDBHomePage.tsx
dmfalke: Sorry, no. The list of studies is fetched from the backend. In some cases, that can take several seconds, and if the user expands the "Interactive maps" menu item, it will be empty. This change renders a loading indicator.

PR #217, opened 2023-05-12
SAM - fix floater overlay order

bobular:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/218

Also moved the `__UNSELECTED__` token stuff to a more central location

Tested for categorical and continuous.

Continuous doesn't work yet for scatterplot - the back end response may be malformed. See https://github.com/VEuPathDB/EdaNewIssues/issues/642


bobular:
Thanks @dmfalke and also for the two-parameter generic that solves it without overloads!

dmfalke (approved):

PR #219, opened 2023-05-15
Force order of filter values for overlay variable to match legend

### LINKED ISSUE #218: Bug: SAM legend and plot colors can be reversed, depending on how filters are added
Looking at the WWARN data on SAM...

I first filtered `Genotype or haplotype assessed` = `pfcrt k76` to look at the wildtype. I got this plot, which looks good:
image

I then went back to change my filters for `Genotype or haplotype assessed` and added `pfcrt 76t` so that i could compare the wildtype to the mutant. I got this plot, which had the colors of the legend and the plot switched:
image

I then cleared all filters and re-added filters for `Genotype or haplotype assessed` ...this time filtering both `pfcrt k76` and `pfcrt 76t` at the SAME time. I got this plot, which once again has the colors correct:
image

Please make sure the legend colors match to the appropriate line in the graph, regardless of how the filters are ordered, thanks!

dmfalke:
Fixes #218

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -4,6 +4,7 @@ import {
   AnalysisState,
   DEFAULT_ANALYSIS_NAME,
   EntityDiagram,
+  Filter,
   PromiseResult,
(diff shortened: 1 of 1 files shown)


dmfalke:
This is addressed in #217

PR #221, opened 2023-05-15
EDA: Make count requests in parallel

dmfalke:
This PR addresses an issue where requests to the count endpoint are done serially. This causes severe lag in the UI, especially when there is a large number of entities in the study.


adnauseum (approved):
🥇 Thank goodness for `Promise.all()`!

PR #225, opened 2023-05-16
Remove hardcoded apps filtering in SAM and only display apps with plugins

### LINKED ISSUE #161: Change logic for determining which visualization apps are shown/usable
As @dmfalke suggested on [Slack](https://epvb.slack.com/archives/C012ZK4Q5CZ/p1682445721644859?thread_ts=1682432486.469829&cid=C012ZK4Q5CZ)

> I'm not sure we need to worry about adding context. The way things work right now it, if an "app" doesn't have a client plugin defined, it is rendered as unavailable. Maybe, instead, we should just not show it at all. If we do that, then in the EDA workspace, and in the standalone map, we can provide client plugins for only the apps we want to show up in each.

Does this also mean we deprecate the "projects" attribute in the `apps` endpoint response (or stop filtering on it in the client)?

bobular:
Note merge target is `sam-wire-floaters`

Partly addresses https://github.com/VEuPathDB/web-monorepo/issues/161

Removes a hardcoded app filter from MapAnalysis and only shows apps in `StartPage` (SAM+Mbio) that have client plugins defined.

Does not address what to do with the `projects` prop in the `apps` response.

At some point after we have removed the "single app mode" we will want to be able to configure different flavours of EDA (and potentially SAM) to use different apps.

So I'm not sure if/when we close #161, but @asizemore has confirmed it is OK to roll out this `StartPage` to MBio.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/StartPage.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/StartPage.tsx
@@ -54,8 +54,9 @@ export function StartPage(props: Props) {
           }}
         >
           {/* orderBy renders available apps ahead of those in development */}
-          {orderBy(apps, [(app) => (plugins[app.name] ? 1 : 0)], ['desc']).map(
-            (app) => (
+          {apps
+            .filter((app) => plugins[app.name] != null)
+            .map((app) => (
               <div
                 style={{
                   display: 'grid',
@@ -224,8 +225,7 @@ export function StartPage(props: Props) {
                   })}
                 </div>
               </div>
-            )
-          )}
+            ))}
         </div>
       </div>
     )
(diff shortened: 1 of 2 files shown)


asizemore:
Tested with the mbio backend and the start page looks good!

dmfalke (approved):
I tested with clinepi, mbio, and SAM. All looks good!

bobular:
Sorry guys. Before I merge, I want to change

https://github.com/VEuPathDB/web-monorepo/blob/778fdd4f66247cce7923e59041ee03a57c25f797/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx#L138

to
```
  const analysisState = useAnalysis(props.analysisId);
```
I don't think it breaks anything but would like a second opinion.

dmfalke:
> Sorry guys. Before I merge, I want to change
> 
> https://github.com/VEuPathDB/web-monorepo/blob/778fdd4f66247cce7923e59041ee03a57c25f797/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx#L138
> 
> to
> 
> ```
>   const analysisState = useAnalysis(props.analysisId);
> ```
> 
> I don't think it breaks anything but would like a second opinion.

I agree, this is the right thing to do!

PR #226, opened 2023-05-17
filter-aware axis range for Vizs

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1698. Changes are made for following Vizs

- Histogram X axis
- Boxplot Y axis
- Lineplot X and Y axes
- Scatterplot X and Y axes
- Scatterplot continuous overlay variable (colormap)

Here is an example screenshot for a case of Scatterplot continuous overlay variable (colormap): BMI-for-age z-core is filtered as [-10, 50]

### DIFF
--- a/packages/libs/eda/src/lib/core/utils/filter-axis-range.ts
+++ b/packages/libs/eda/src/lib/core/utils/filter-axis-range.ts
@@ -1,0 +1,24 @@
+import { Filter } from '../types/filter';
+import { VariableDescriptor } from '../types/variable';
+
+export function filterMinMax(
+  filters: Filter[] | undefined,
+  variable: VariableDescriptor | undefined
+) {
+  const minMaxArray =
+    filters != null && variable != null
+      ? filters
+          ?.map((value) => {
+            if (value.type === 'numberRange' || value.type === 'dateRange')
+              return value.entityId === variable?.entityId &&
+                value.variableId === variable?.variableId
+                ? value
+                : undefined;
+          })
+          .filter((data) => data != null)
+      : undefined;
+
+  return minMaxArray != null && minMaxArray.length > 0
+    ? { min: minMaxArray[0]?.min, max: minMaxArray[0]?.max }
+    : undefined;
+}
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/BoxplotVisualization.tsx
@@ -102,6 +102,8 @@ import { useDeepValue } from '../../../hooks/immutability';
 // reset to defaults button
 import { ResetButtonCoreUI } from '../../ResetButton';
 
+import { filterMinMax } from '../../../utils/filter-axis-range';
+
 type BoxplotData = { series: BoxplotSeries };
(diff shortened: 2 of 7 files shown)


packages/libs/eda/src/lib/core/utils/filter-axis-range.ts
```diff
@@ -0,0 +1,24 @@
...
+import { VariableDescriptor } from '../types/variable';
+
+export function filterMinMax(
+  filters: Filter[] | undefined,
+  variable: VariableDescriptor | undefined
+) {
+  const minMaxArray =
+    filters != null && variable != null
+      ? filters
+          ?.map((value) => {
```
bobular, line 11: I prefer `filters.map((filter) => ...`
(rather than `(value) => ...`)

bobular (commented):
This is great, and works as specified (ignore my comments in Slack about zeroes).

However, @dmfalke's suggestion in Slack is a good one:

> If we implement this by augmenting the variable's metadata such that it's display range is based on the filter, would we need to do anything else?

See how Dave handled the categorical filter-awareness here in `useStudyEntities(filters)`

https://github.com/VEuPathDB/web-monorepo/pull/139/files#diff-c97f01b67eb5e0321353d3d5fb9d4d8183b2e584239f2dbefe6d8ed27f8827fdR96

We could similar here: based on filter values, augment the study metadata's `distributionDefaults.rangeMin` and `rangeMax`.  Then `useDefaultAxisRange` etc and the individual visualizations could stay unchanged.

moontrip:
@bobular Thank you for your reviews and tests (still ongoing I suppose 😄). I got the point about augmenting metadata so I will try to do that. 👍

moontrip:
@bobular forgot to remind you that I refactored codes to simplifying. It looks fine for me but it would be great if you can test this PR again. Thanks!

packages/libs/eda/src/lib/core/hooks/workspace.ts
```diff
@@ -116,12 +116,70 @@ export function useStudyEntities(filters?: Filter[]) {
...
+
+                    // augment variable metadata including filter-aware axis range
+                    if (
+                      variable.type === 'number' ||
+                      variable.type === 'integer'
+                    )
+                      return {
+                        ...variable,
+                        vocabulary,
+                        distinctValuesCount: vocabulary?.length ?? 0,
```
bobular, line 143: Nice!
I don't have time for a full review or test right now, but I wonder if we need to bother augmenting `distinctValuesCount` for range filters?  I don't think the `vocabulary` exists on `type: 'number'` variables, and I don't think we want it to be zero (could that "constrain it away" from some variable pickers?).

We don't know how many values are left after filtering. Would it be useful to know that only 7 distinct values are left? I guess so, it would make it available as a barplot x-axis variable. BUT is that really worth the hassle of finding out how many values pass the filter?  I don't think so.  I think it would be save to leave `distinctValuesCount` unchanged.

@dmfalke second opinion?

dmfalke: I agree that we should leave that value unchanged. I don't think it will impact the primary use case, and we can't easily find the correct value.

moontrip:
Thank you for your review and test @bobular 👍

PR #143, opened 2023-05-02
Wire new organised floating visualization plugins to SAM

bobular:
Resolves #45 and #86 

This currently only works with a local back end with these versions. Let me know if you need help setting it up.
```
SUBSETTING_TAG=qa
COMPUTE_TAG=qa
MERGING_TAG=qa
DATA_TAG=branch-sam-floater-groups
RSERVE_TAG=qa
USER_TAG=qa
SERVICE_TAG=qa
DAS_TAG=qa
```

- [ ] make `useStandaloneVizPlugins` hook to create customised plugins for SAM and add `plugins` prop with fallback to uncustomised statically defined plugins to any component that used the static plugins
- [ ] re-use `StartPage` (MBio's new viz picker) in place of `NewVisualizationPickerGrouped`
- [ ] retire `NewVisualizationPickerGrouped`
- [ ] add props to control layout of `StartPage` in SAM menu context
- [ ] proper handling of computations and visualizations (replace `computations[0]` with something appropriate) see also https://github.com/VEuPathDB/web-eda/issues/1495 
- [ ] get floating visualisations working (probably one checkbox per viz)

PR #232, opened 2023-05-18
142: Add loading spinner to map legend

adnauseum:
**Before**: 

**After**:

image

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import PlotLegend from '@veupathdb/components/lib/components/plotControls/PlotLegend';
 import { LegendItemsProps } from '@veupathdb/components/lib/components/plotControls/PlotListLegend';
 import { Loading } from '@veupathdb/wdk-client/lib/Components';
--- a/packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
@@ -1,33 +1,40 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PlotLegend from '@veupathdb/components/lib/components/plotControls/PlotLegend';
 import { LegendItemsProps } from '@veupathdb/components/lib/components/plotControls/PlotListLegend';
+import { Loading } from '@veupathdb/wdk-client/lib/Components';
 
 interface Props {
   legendItems: LegendItemsProps[];
+  isLoading: boolean;
   title?: string;
   showCheckbox?: boolean;
 }
 
 export function MapLegend(props: Props) {
-  const { legendItems, title = 'Legend', showCheckbox } = props;
+  const { legendItems, isLoading, title = 'Legend', showCheckbox } = props;
+
   return (
     <>
       <div>
         <strong>{title}</strong>
       </div>
-      <PlotLegend
-        type="list"
-        legendItems={legendItems}
-        showOverlayLegend
-        containerStyles={{
-          border: 'none',
-          boxShadow: 'none',
-          padding: 0,
-          width: 'auto',
-          maxWidth: 400,
-        }}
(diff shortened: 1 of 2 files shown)


packages/libs/components/src/components/Spinner.tsx
```diff
@@ -1,6 +1,10 @@
 import CircularProgress from '@material-ui/core/CircularProgress';
 
-export default function Spinner() {
+interface Props {
+  size?: string | number;
+}
+
+export default function Spinner({ size }: Props) {
```
bobular, line 7: Shouldn't `size` default to 50 for backwards compatibility?

bobular (changes requested):
Works great but let's see how it looks after the merge from `main`.  I did warn about this in the parent issue! ;-) 

> Don't tackle this until https://github.com/VEuPathDB/web-monorepo/pull/143 is complete! Changes to default overlay behaviour are happening.

packages/libs/components/src/components/Spinner.tsx
adnauseum: Whoops, it totally did until I accidentally deleted it! Thanks for catching that! Restored in: https://github.com/VEuPathDB/web-monorepo/pull/232/commits/d0a319a27ddd17dfc0aafc2eb2aea71b5d20d63f

adnauseum:
> Works great but let's see how it looks after the merge from `main`. I did warn about this in the parent issue! ;-)
> 
> > Don't tackle this until #143 is complete! Changes to default overlay behaviour are happening.

Lol I meant to branch off of #143, but I forgot. I'm definitely not (merge) conflict averse, so I'm gonna dive in 👿

bobular:
Good luck! (I have no idea what the conflicts are.)

I don't think there's an easy way to get the pending status of the `getDefaultOverlayConfig` async function - not without wrapping it in a `usePromise` (and I don't remember why I didn't).

However, things could change when Dave tackles the `useEffect` pandemic, so probably best not to make any assumptions now.

Might be good to wrap all the `legendIsLoading` logic in a little hook (same source file) just to keep it neat and tidy.

adnauseum:
> Works great but let's see how it looks after the merge from `main`. I did warn about this in the parent issue! ;-)
> 
> > Don't tackle this until #143 is complete! Changes to default overlay behaviour are happening.

Speaking of that, did the legend break during #143?

adnauseum:
> Good luck! (I have no idea what the conflicts are.)
> 
> I don't think there's an easy way to get the pending status of the `getDefaultOverlayConfig` async function - not without wrapping it in a `usePromise` (and I don't remember why I didn't).
> 
> However, things could change when Dave tackles the `useEffect` pandemic, so probably best not to make any assumptions now.
> 
> Might be good to wrap all the `legendIsLoading` logic in a little hook (same source file) just to keep it neat and tidy.

The merge was no biggie. Turns out that the legend loading state is as easy as `legendItems.length === 0` 👍

packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
```diff
@@ -1,33 +1,49 @@
...
 }
 
 export function MapLegend(props: Props) {
-  const { legendItems, title = 'Legend', showCheckbox } = props;
+  const { legendItems, isLoading, title = 'Legend', showCheckbox } = props;
+  console.log(
+    '🚀 ~ file: MapLegend.tsx:14 ~ MapLegend ~ isLoading:',
+    isLoading
+  );
+  console.log(
```
bobular, line 18: Remember to remove these guys :-)

bobular (approved):
Nice simplification!

Works fine on the true categorical+continuous variables (after I broke things being addressed in https://github.com/VEuPathDB/web-monorepo/pull/255)

packages/libs/eda/src/lib/map/analysis/MapLegend.tsx
adnauseum: Oh dear! I put those in there because I thought I broke something and was confuzzled! But it was something else (#255).

Removed in: https://github.com/VEuPathDB/web-monorepo/pull/232/commits/d14593eb0f00c6faf06af461f100d9ea57ff9444

PR #240, opened 2023-05-22
Wire up a `workspace/maps` route

### LINKED ISSUE #212: SAM: Add link to "My interactive maps" under "Workspace"
### Description

There are two parts to this issue:
1. Update the routes
2. Update the menu items

### Update the routes

- [ ] Replace `/maps` with `/workspace/maps`
- [ ] Add a redirect from `/maps` to `/workspace/maps` for backward compatibility
- [ ] Update `packages/libs/eda/src/lib/map/MapVeuContainer.tsx` to render `AllAnalyses`, instead of `StudyList`.

### Update the menu items

_This only applies to genomics-site_

- [ ] Update the existing map links to use `/workspace/maps`
- [ ] Add a link to **My interactive maps** under the **Workspace** menu heading

jernestmyers:
Resolves #212

New menu item:
image

Result of clicking the `My interactive maps` menu item (routes to `workspace/maps`):
image

### DIFF
--- a/packages/libs/web-common/src/routes.jsx
+++ b/packages/libs/web-common/src/routes.jsx
@@ -71,11 +71,19 @@ export const wrapRoutes = (wdkRoutes) => [
   },
 
   {
-    path: '/maps',
+    path: '/workspace/maps',
     exact: false,
     isFullscreen: true,
     rootClassNameModifier: 'MapVEu',
     component: EdaMapController,
+  },
+
+  {
+    path: '/maps',
+    exact: false,
+    isFullscreen: true,
+    rootClassNameModifier: 'MapVEu',
+    component: () => <Redirect to="/workspace/maps" />,
   },
 
   {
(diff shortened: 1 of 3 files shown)


packages/libs/web-common/src/routes.jsx
```diff
@@ -27,7 +27,7 @@ import { EdaMapController } from './controllers/EdaMapController';
 export const STATIC_ROUTE_PATH = '/static-content';
 
 export function makeEdaRoute(studyId) {
-  return '/workspace/analyses' + (studyId ? `/${studyId}` : '');
+  return '/workspace/maps/analyses' + (studyId ? `/${studyId}` : '');
```
dmfalke, line 30: This should be reverted

dmfalke (commented):
Something I just noticed is that `/workspace/maps` is rendered without  the site header/footer. I think we want that. One way to do it is to add two route entries in `routes.jsx`:

```
  {
    path: '/workspace/maps',
    exact: true,
    component: EdaMapController,
  },

  {
    path: '/workspace/maps',
    exact: false,
    isFullscreen: true,
    rootClassNameModifier: 'MapVEu',
    component: EdaMapController,
  },
```

The first one will match the exact route `/workspace/maps`, and the second will match all other routes that start with `/workspace/maps`.

Issue #205, opened 2023-05-11
Change SAM floating plot overlay <=> global marker overlay variable interaction behaviour


bobular:
Currently this is managed with an effect (`useProvidedOptionalVariable`) that watches for changes in the global overlay variable and makes changes to the currently active viz's config.

This is causing some unwanted data requests. For example, when the viz is first rendered, a request is made for the config, then the config changes and a new one is made.

### New approach

Similar to how new filter-aware variable constraints are now handled (these can change and invalidate a viz config) we can simply not display the viz and message the user to say they should change something.  Avoid effects completely if possible!

See `assertValidInputVariables` as introduced by https://github.com/VEuPathDB/web-monorepo/pull/139

dmfalke:
Is this still an issue?

PR #251, opened 2023-05-23
SAM: Remove some `useEffect` uses

dmfalke:
This PR gets rid of some `useEffect` hook uses. Some related code cleanup is also addressed.

### DIFF
--- a/packages/libs/eda/src/lib/core/hooks/workspace.ts
+++ b/packages/libs/eda/src/lib/core/hooks/workspace.ts
@@ -25,6 +25,12 @@ import { ComputeClient } from '../api/ComputeClient';
 import { DownloadClient } from '../api';
 import { Filter } from '../types/filter';
 import { mapStructure } from '@veupathdb/wdk-client/lib/Utils/TreeUtils';
+import {
+  useFeaturedFieldsFromTree,
+  useFieldTree,
+  useFlattenedFields,
+} from '../components/variableTrees/hooks';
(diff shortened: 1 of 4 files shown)


packages/libs/eda/src/lib/core/hooks/workspace.ts
```diff
@@ -154,3 +160,49 @@ function defaultMakeVariableLink({
     ? `/variables/${entityId}`
     : `/variables`;
 }
+
+/**
+ * TODO: This is pasted directly `DefaultVariableRedirect`. Cover this hook by some
+ * kind of test and simplify its logic.
+ */
+export function useGetDefaultVariableDescriptorCallback() {
```
bobular, line 168: While we're moving it. I think we could probably lose the `Callback` at the end of the hook name?

It already has `Get` in the name, suggesting it returns a function.

dmfalke:
Setting this as "ready for review". I expect some discussion about the changes :smile:

dmfalke:
The most controversial thing are the changes to marker configuration: removal of `overlayConfig`, and additional of `selectedValues`

bobular:
That didn't freak me out in the slightest. Well tbh it did to start with... 🤣

bobular (approved):
Just revisited this - did some testing etc.

Are we planning to remove the effect in `useProvidedOptionalVariable` (used by floating viz).

Ah yes, we are. But not in this issue/PR.  It has its own ticket: https://github.com/VEuPathDB/web-monorepo/issues/205

PR #256, opened 2023-05-24
Add cypress to monorepo

adnauseum:
# Component testing

https://github.com/VEuPathDB/web-monorepo/assets/24446573/5dc181ff-0bc4-46db-b34a-1922effd64b3


adnauseum:
This will be kept for posterity!

Issue #249, opened 2023-05-22
Map: Bubble size legend component


d-callan:
bubble markers need a legend to indicate what their relative size represents. they should look something like this:
![bubble_legend](https://github.com/VEuPathDB/web-monorepo/assets/53306535/d4a263bc-2d5c-4f18-aa79-366bf1357118)
 
note there should be a dashed line connected each circle to a value as well

Issue #95, opened 2023-04-21
Bubble marker react component


steve-fischer-200:
Create svg marker similar to donut.  Props include size and color.  In future support gradients. 

Probably don't need a magnify mode.

d-callan:
also worth noting they scale continuously in terms of size. so well have some size bubble we know to be the largest we can make, and some size we know to be the smallest. then we can also have any size in between, if the data requires it. the data will not be binned in a manner that would let us use one size bubble for all values in a specified range.

PR #257, opened 2023-05-25
Bubble markers map type

moontrip:
(Connor is new lead on this PR)
Resolves https://github.com/VEuPathDB/web-monorepo/issues/263
Resolves https://github.com/VEuPathDB/web-monorepo/issues/95
Resolves https://github.com/VEuPathDB/web-monorepo/issues/249


chowington:
image

moontrip:
The screenshot looks great, @chowington 👍 Seems like center label helps to alleviate overlapped bubbles like bottom left bubbles in your screenshot.

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
```diff
@@ -316,30 +368,50 @@ export function useStandaloneMapMarkers(
                 },
               ];
 
+        // here's what Bob pointed out
         const count =
           vocabulary != null // if there's an overlay (all expected use cases)
             ? overlayValues.reduce((sum, { count }) => (sum = sum + count), 0)
```
d-callan, line 374: just curious, why not just use entityCount always?

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
```diff
@@ -316,30 +368,50 @@ export function useStandaloneMapMarkers(
...
             : entityCount; // fallback if not
 
+        const bubbleData = [
+          {
+            label: reorderedData[0].label,
+            value: count,
+            color:
+              'color' in reorderedData[0] ? reorderedData[0].color : undefined,
```
d-callan, line 382: is this a temp thing until i have the backend put together? not sure i understand..

asizemore:
https://github.com/VEuPathDB/web-monorepo/issues/311 for the bubble icon

For the overlapping bubbles, it's common in bubble charts to have the stroke be whatever color and then the fill be the same color but maybe 0.7 opacity like in the following

Screen Shot 2023-06-22 at 5 35 56 AM

And sorry to leave the discussion early at the dev meeting yesterday! Did y'all come to a conclusion about the bubble sizing wrt different zoom levels and the legend?

bobular (commented):
I just want to put these comments in before I get distracted by other work. Will return for a proper review!

bobular:
@chowington are you planning to use `dependentAxisRange` in the bubble markers?
Probably not needed if there's a `valueToSizeMapper` function passed around?
 
If not, could you remove it please.

Also a switch statement here would be lovely please:

```
  const markers = useMemo(
    () =>
      markersData?.map((markerProps) =>
        markerType === 'pie' ? (
          <DonutMarkerComponent {...markerProps} />
        ) : markerType === 'bubble' ? (
          <BubbleMarkerComponent {...(markerProps as BubbleMarkerProps)} />
        ) : (
          <ChartMarkerComponent {...markerProps} />
        )
      ) || [],
    [markersData, markerType]
  );
```

packages/libs/components/src/map/BubbleMarker.tsx
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/**
+ * this is a SVG bubble marker icon
+ */
+export default function BubbleMarker(props: BubbleMarkerProps) {
+  const { html: svgHTML, size } = bubbleMarkerSVGIcon(props);
+
+  // set icon as divIcon
+  const SVGBubbleIcon = L.divIcon({
```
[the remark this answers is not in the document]
chowington: Yep, still need a nice icon for the bubble marker menu item!

packages/libs/components/src/stories/BubbleMarkers.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  BubbleMarkerStandalone,
+} from '../map/BubbleMarker';
+
+export default {
+  title: 'Map/Bubble Markers',
+} as Meta;
+
+const valueToSizeMapper = (value: number) => {
+  // Area scales directly with value
+  const constant = 100;
```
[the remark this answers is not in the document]
chowington: No, this is just the scaling factor/"slope" to convert a value to a size in pixels

packages/libs/components/src/stories/plotControls/PlotLegend.stories.tsx
```diff
@@ -527,6 +527,43 @@ export const GradientPlotLegend = () => {
   );
 };
 
+export const BubbleMarkerLegend = () => {
+  const maxValue = 100;
```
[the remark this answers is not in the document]
chowington: Since this story is just the legend without any actual markers, I need to set a fictional max marker value for the bubble sizing function

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
chowington: Not sure what in particular you're talking about here, but yeah most likely a temporary thing, lol

packages/libs/components/src/stories/BubbleMarkers.stories.tsx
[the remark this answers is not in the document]
d-callan: i still dont understand..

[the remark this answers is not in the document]
chowington: It doesn't make sense to make the circle size exactly equal to the value, like a value of 500 -> 500px circle, so the constant is what you multiple the value by to get the circle size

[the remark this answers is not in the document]
chowington: I've changed it to scalingFactor locally, with I think is more clear

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
```diff
@@ -252,11 +268,47 @@ export function useStandaloneMapMarkers(
...
+                    0
+                  )
+                : mapElement.entityCount;
+            return count;
+          })
+        )
+      : 0;
+
+    const bubbleValueToSizeMapper = (value: number) => {
+      const largestCircleArea = 9000;
```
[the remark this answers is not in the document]
chowington: I eyeballed it 😉

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
chowington: Dunno, I just left this as it was!

chowington:
> @chowington are you planning to use `dependentAxisRange` in the bubble markers? Probably not needed if there's a `valueToSizeMapper` function passed around?
> 
> If not, could you remove it please.
> 
> Also a switch statement here would be lovely please:
> 
> ```
>   const markers = useMemo(
>     () =>
>       markersData?.map((markerProps) =>
>         markerType === 'pie' ? (
>           <DonutMarkerComponent {...markerProps} />
>         ) : markerType === 'bubble' ? (
>           <BubbleMarkerComponent {...(markerProps as BubbleMarkerProps)} />
>         ) : (
>           <ChartMarkerComponent {...markerProps} />
>         )
>       ) || [],
>     [markersData, markerType]
>   );
> ```

I was using dependentAxisRange before moving to the size function, so yes, I can remove it now

chowington:
> Thinking more about accommodating the new back end.
> 
> In `useStandaloneMapMarkers` (should be renamed maybe to `useStandaloneMapMarkerData`?):
> 
> I think renaming this type `StandaloneMapMarkersResponse` into `DonutMarkersResponse | ChartMarkersResponse` with those two types actually being identical/aliases (until we tackle the tech debt mentioned above) - this will help with readability (I much prefer it to `DonutAndChartMarkersResponse`). And something similar with the request types.
> 
> Then we can move ahead with `DonutMarkersResponse | ChartMarkersResponse | BubbleMarkersResponse`.
> 
> The reduce in the same hook that returns `{ valueMax, valueMinPos, countSum }` will likely break for the bubble response, but we only need to do it for the chart markers (hence my question above) - so we can start being a bit more selective about what we do for the different marker types.

This is pretty much what I've been doing so far, yes!

bobular:
Hi @chowington  Having a look now.  A few things I noticed:

* bubbles seem to work great!
* I like the two separate legends but not when they shrink/expand between each variable update
  - seems to happen when panning the map too - isn't the legend data supposed to be valid for the entire map and all zoom levels?
* legend has disappeared from donut (and I assume barplot) marker mode
* back end problem with binary variables?
   - SCORE S. mansoni CRT -> Schistosoma mansoni, by microscopy -> `Error: 400 Bad Request {"status":"bad-request","message":"{\"keyedErrors\":{},\"validationLevel\":\"RUNNABLE\",\"validationStatus\":\"FAILED\",\"errors\":[\"Incorrect overlay configuration type for categorical var: BINARY\"]}"}`
   - I've Slacked the back end people and you about it
* the smallest bubbles are smaller than the number inside them (SCORE S. mansoni as above) 
image
* or are numbers just a temporary thing?  improved visibility of small markers is going to be important.
* there's a (x) button by the main "color" variable picker but it doesn't work and isn't needed - I think Jeremy addressed this but can't find the ticket right now - don't know if it's merged
* do we want a default value pre-selected for the numerator so that markers "just work" when you enter the mode? It would make it consistent with the other modes, but assumes people notice they can change the num/denom values without having been forced to select a numerator.
* that's all for now - will have another play tomorrow and look at the code.

chowington:
@bobular In regards to the numbers inside the bubbles, they were originally intended only as a sanity check for me, and were not part of the original mockup. I think we can go either way:

- If we keep them, we can "truncate" the numbers as we do with the donut marker numbers, and increase the minimum bubble size to a size that can fit 4 characters (the most I believe you'd need for a reasonable truncation algorithm)
- If we remove them, we can easily add a popup showing *just the number* (maybe with a "Count:" label). We wouldn't want to show an enlarged version of the bubble for obvious reasons.

Now that I'm thinking about it, even in the first situation, we'd want to have the popup so users could see the precise count.

What do y'all think?

bobular:
Hi @chowington - I'm thinking for consistency we should have a pop-up. It doesn't need to show a larger circle (in fact this would be a bad thing) but it could definitely show the count and the colour-coded value.

Would it be practical (is there enough space) to describe the number in full, e.g.
```
Count of Samples: 1,234
Fraction of infected/(infected+uninfected+not done) in Schistosoma infection by microscopy: 0.23
```
I don't think so!

chowington:
> there's a (x) button by the main "color" variable picker but it doesn't work and isn't needed - I think Jeremy addressed this but can't find the ticket right now - don't know if it's merged

@jernestmyers Does this sound familiar to you?

jernestmyers:
> > there's a (x) button by the main "color" variable picker but it doesn't work and isn't needed - I think Jeremy addressed this but can't find the ticket right now - don't know if it's merged
> 
> @jernestmyers Does this sound familiar to you?

Yes! There are actually two merged PRs of late that give us more freedom in how we do/don't render the "clear selection" button:
- https://github.com/VEuPathDB/web-monorepo/pull/371
- https://github.com/VEuPathDB/web-monorepo/pull/396/files#diff-fc0752b79dcba305bff4ee47bac712209bfc5d992776b7684186a05cf2592f41

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -385,17 +402,35 @@ function MapAnalysisImpl(props: ImplProps) {
...
+        numeratorValues:
+          activeMarkerConfiguration &&
+          'numeratorValues' in activeMarkerConfiguration
+            ? activeMarkerConfiguration.numeratorValues
+            : undefined,
+        denominatorValues:
+          activeMarkerConfiguration &&
+          'denominatorValues' in activeMarkerConfiguration
+            ? activeMarkerConfiguration.denominatorValues
+            : undefined,
```
chowington, line 424: I do this sort of conditional property accession quite a few times. Anyone know of more succinct way to do this that both grabs the value and narrows the type? For instance, Lodash's `get` function is a great way to grab the value conditionally, but it doesn't narrow the type.

chowington:
Looks like I can't remove people as reviews if they've already given a review--I'm assigning @bobular to be the primary reviewer if that's alright with you, Bob!

chowington:
So honestly there's not much I can think of that requires specific attention/decision. Maybe some minor points:

- I removed the numbers inside the bubbles since adding the popup, as they were not part of the mockup, and it would be a headache to make them fit inside the smaller bubbles well.
- The default gradient is the brownish color; I don't particularly like that for the map, as it can be hard to distinguish the brown from the map background. That is the default we use for the scatter plot though. I may look into changing it unless others think it's fine.
- There is still the issue with binary variables failing at the backend.

chowington:
Actually, I'll tag @asizemore for the gradient topic, as she put a lot of thought into choosing these gradients!

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
chowington: Okay so technically `_.get()` does work for this particular example, but there are a couple of others like this that it doesn't work for, because TS thinks the expression can only lead to a value of `undefined`

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -492,15 +526,17 @@ function MapAnalysisImpl(props: ImplProps) {
...
       );
     }
-  }, [previewMarkerData]);
+  }, [activeMarkerConfiguration, markerType, previewMarkerData]);
 
   const markers = useMemo(
     () =>
       markersData?.map((markerProps) =>
         markerType === 'pie' ? (
-          <DonutMarkerComponent {...markerProps} />
```
chowington, line 501: Remove type assertions, replace by prop filtering

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -674,7 +718,9 @@ function MapAnalysisImpl(props: ImplProps) {
                     }
                     toggleStarredVariable={toggleStarredVariable}
                     constraints={markerVariableConstraints}
-                    overlayConfiguration={activeOverlayConfig.value}
+                    overlayConfiguration={
+                      activeOverlayConfig.value as OverlayConfig
+                    }
```
chowington, line 723: Look into this assertion too

asizemore:
So exciting to see this all come together!!! I'd be happy to tackle the gradient issue. I'll make it a new issue and assign myself

dmfalke (commented):
@chowington is looks like you've addressed the major things we discussed. I'm ready to approve this! Is there anything else you're working on, in this PR?

chowington:
@dmfalke I'm just merging in main and testing, but I've done otherwise 👍

dmfalke (approved):
:rocket:

PR #260, opened 2023-05-25
Implement `Map Type` UI

### LINKED ISSUE #243: Map: Introduce 'Map type' menu entry in place of 'Markers' and 'Plot'
this has been discussed in eda ux meetings, that the relationship between the markers and plot menu items is unclear. We've agreed that the markers and plot menu items will be replaced w a single menu item 'Map type'. map type will be expandable and list the different map types we currently offer, in the top tier of the menu directly. after selecting a map type, a panel expands allowing for configuration of markers and supporting plots. see ex: 
map_menu

the supporting plots tab will list both the currently configured supporting plots and the types of plots we have available to configure, exact details/ styling tbd (provided by @asizemore when ready). see this tentative ex: 
supporting_plots

depends on https://github.com/VEuPathDB/web-monorepo/issues/244

jernestmyers:
Resolves #243

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -99,6 +99,7 @@ enum MapSideNavItemLabels {
   Share = 'Share',
   StudyDetails = 'View Study Details',
   MyAnalyses = 'My Analyses',
+  MapType = 'Map Type',
 }
 
 type SideNavigationItemConfigurationObject = {
@@ -107,6 +108,16 @@ type SideNavigationItemConfigurationObject = {
   icon: ReactNode;
   renderSideNavigationPanel: (apps: ComputationAppOverview[]) => ReactNode;
   onToggleSideMenuItem?: (isActive: boolean) => void;
+  isExpandable?: boolean;
+  isExpanded?: boolean;
+  subMenuConfig?: SubMenuItems[];
+};
+
+type SubMenuItems = {
+  labelText: string;
+  icon?: ReactNode;
+  onClick: () => void;
+  isActive: boolean;
 };
 
 function getSideNavItemIndexByLabel(
@@ -119,6 +130,33 @@ function getSideNavItemIndexByLabel(
 const mapStyle: React.CSSProperties = {
   zIndex: 1,
   pointerEvents: 'auto',
+};
+
+/**
+ * The following code and styles are for demonstration purposes
+ * at this point. After #1671 is merged, we can implement these
+ * menu buttons and their associated panels for real.
+ */
+const buttonStyles: React.CSSProperties = {
+  alignItems: 'center',
+  background: 'transparent',
+  borderColor: 'transparent',
+  display: 'flex',
+  fontSize: 16,
+  justifyContent: 'flex-start',
+  margin: 0,
+  padding: 0,
+  width: '100%',
+};
+const iconStyles: React.CSSProperties = {
+  alignItems: 'center',
+  display: 'flex',
+  height: 25,
+  justifyContent: 'center',
+  width: 25,
+};
+const labelStyles: React.CSSProperties = {
+  marginLeft: '0.5em',
 };
 
 export const defaultAnimation = {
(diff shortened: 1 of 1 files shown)


packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/MapTypeConfigurationMenu.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { H5 } from '@veupathdb/coreui';
+import { MarkerConfiguration } from '../appState';
+import TabbedDisplay, {
+  TabbedDisplayProps,
+} from '@veupathdb/coreui/dist/components/grids/TabbedDisplay';
+
+export interface MarkerConfigurationOption {
+  displayName: string;
+  icon?: JSX.Element;
+  renderConfigurationMenu: JSX.Element;
```
dmfalke, line 11: Rename this `configurationMenu`. Reserve the `render` prefix for render functions.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -122,6 +136,33 @@ const mapStyle: React.CSSProperties = {
...
+ * The following code and styles are for demonstration purposes
+ * at this point. After #1671 is merged, we can implement these
+ * menu buttons and their associated panels for real.
+ */
+const buttonStyles: React.CSSProperties = {
+  alignItems: 'center',
+  background: 'transparent',
+  borderColor: 'transparent',
+  display: 'flex',
+  fontSize: 16,
```
dmfalke, line 149: I know this predates this PR, but this should really be a relative font size. It might be best to make a follow-up issue for this.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -452,99 +493,120 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
...
 
+          const markerConfigurationObjects: MarkerConfigurationOption[] = [
+            {
+              type: 'pie',
+              displayName: 'Donuts',
+              icon: <DonutMarkers style={{ height: 30 }} />,
+              renderConfigurationMenu:
+                activeMarkerConfiguration?.type === 'pie' ? (
+                  <PieMarkerConfigurationMenu
+                    inputs={[{ name: 'overlay', label: 'Overlay' }]}
```
dmfalke, line 533: `name` should still be `overlayVariable`

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -878,11 +914,28 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
...
+          } else {
+            // "normalize" the activeSideMenuIndex to resolve the parent's index
+            const indexOfParentContainer = Math.floor(
+              activeSideMenuIndex / 10 - 1
+            );
+            return sideNavigationButtonConfigurationObjects[
+              indexOfParentContainer
+            ].renderSideNavigationPanel(apps);
+          }
+        }
```
dmfalke, line 938: I'm not crazy about how this is being done. Instead of using indexes, I think we should give each menu and submenu item a unique id, and use that to keep track of which menu item is selected. If a top-level menu item has a submenu, then we should never set it as active. How does this sound?

dmfalke (commented):
I've reviewed the code and left some comments and suggestions. Let me know what you think. I will test the code, shortly.

dmfalke:
> I think @asizemore can provide the updated icons for the menu
> 
> image

Looking at this screenshot, the meaning of the blue dot isn't obvious. That might improve with the new icons, since the dot is currently the same size as the marker icon.

But, another idea is to use something like an mui chip to indicate which is active. I don't think this styling is optimal (could use some sort of a "glowing" appearance):

image

asizemore:
New icons sent to @jernestmyers via slack!

I do like that active chip... very nice. I'd be happy either way. While the little dot is small, it's also tough for the user to ever get into a position where they need to know what marker is active but they have no markers in the viewport.

dmfalke:
> New icons sent to @jernestmyers via slack!
> 
> 
> 
> I do like that active chip... very nice. I'd be happy either way. While the little dot is small, it's also tough for the user to ever get into a position where they need to know what marker is active but they have no markers in the viewport. 

I agree that it's probably obvious which is active. The other thing about the dot, that I'm just realizing, is that it commonly means that there is a update.

I'm not in love with the label, but I think it might be better than the dot.

dmfalke (approved):
LGTM :rocket:

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
jernestmyers: Here's the follow up issue for this: https://github.com/VEuPathDB/web-monorepo/issues/274

PR #261, opened 2023-05-25
fix typo in score for ortho layout

jbrestel:


steve-fischer-200 (approved):
I can't really assess this without looking more broadly at the datastructure, etc, but i am sure you've fixed the bug!

asizemore:
This PR addresses an issue found with the cytoscape network layout on ortho ([example](https://orthomcl.org/orthomcl/app/cluster-graph/OG6_107162)). The original issue was that the edge scores were not being used correctly in the layout.

Once this first issue was fixed, some example cluster networks still did not group as expected. @jbrestel and I did some testing and found that the [cola](https://orthomcl.org/orthomcl/app/cluster-graph/OG6_107162)
 layout did a better job overall of laying out the cluster networks in a reasonable manner.

Note that this new layout will _not_ work for all networks. Layout algorithms always contain biases towards networks with different properties. This one seems to do well with mid-sized networks with flat-ish edge weight distributions. Additionally, the cluster networks themselves are also highly heterogeneous with differently skewed edge weight distributions, different sizes, and overall architecture, and so finding one layout algorithm to fix them all would be quite unlikely. 

When we look to create network visualizations in the future, we should keep the above in mind and strongly consider offering the user a choice of layout for their network.

PR #238, opened 2023-05-20
Sam - enable proportion markers

bobular:
Resolves #203 

Branched from `sam-wire-floaters` but I see that it also merges to `main` without problems.

Also removed an unused prop in the barplot config.

PR #262, opened 2023-05-25
Map: Fix the counts under the bar chart markers in proportion mode

bobular:
This wasn't properly fixed in https://github.com/VEuPathDB/web-monorepo/pull/238 - sorry!


bobular:
This doesn't currently handle no overlay at all.
But we don't allow that anyway.  But I might add a fallback just in case.

dmfalke (approved):

PR #264, opened 2023-05-26
Remove fullscreen apps relate code from EDA

### LINKED ISSUE #231: Remove full screen map (FSM) code from repo
This may seem like minor technical debt but we are regularly having to modify this code so it compiles.

I suggest we keep the minimap component because I have heard people say it might be useful.

Note that we also use "full screen" to refer to the full size EDA visualizations in comparison to their thumbnails (and not to be confused with the full screen modals that pop up from faceted visualizations!).

Probably want to start by tracing the `enableFullScreenApps` boolean config variable.

dmfalke:
fixes #231


packages/libs/eda/src/lib/core/components/fullScreenApps/MiniMap.tsx
bobular, line 1: I was thinking maybe we keep the minimap code around for a rainy day?

But looking at the code, there's nothing really special here. As long as we keep `tinyLeafletZoomLevelToGeohashLevel` in `@veupathdb/components/lib/map/utils/leaflet-geohash` we'll be OK. (Which we are keeping.)

bobular (approved):

PR #267, opened 2023-05-31
Fix infinite recursion when faceting

bobular:
No related issue. Just a bug fix.


moontrip (approved):
It works, @bobular Thank you for your fast fix 👍

PR #270, opened 2023-05-31
Remove overlayVariable update `useEffect` from floating visualizations

bobular:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/205 

The main thing that is not clear from the code is that for floating visualizations, `vizConfig.overlayVariable` is used basically as a boolean - if it is non-null, then overlay is on. The actual overlay variable comes from `options?.getOverlayVariable()` which gets loaded into `providedOverlayVariable` and `providedOverlayVariableDescriptor`.

Task list

- [x] barplot
- [x] figure out warning message for dependency-order based constraint violations
- [x] scatterplot
- [x] boxplot
- [x] lineplot
- [x] histogram 
- [x] consult crows and remove snackbar context?? answer = no
- [x] remove the `useProvidedOptionalVariable` hook source code


packages/libs/eda/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -297,33 +295,13 @@ function BarplotViz(props: VisualizationProps<Options>) {
 
   const selectedVariables = useDeepValue({
     xAxisVariable: vizConfig.xAxisVariable,
-    overlayVariable: vizConfig.overlayVariable,
+    overlayVariable: vizConfig.overlayVariable
```
bobular, line 298: I will add a comment to explain this

bobular:
Now I've merged `main` I see there is an extra back end request from a floating viz when switching map types.  The first request is with the overlay variable from the old map type, then the correct request is made.

I'll look into it tomorrow.

dmfalke:
> Now I've merged `main` I see there is an extra back end request from a floating viz when switching map types. The first request is with the overlay variable from the old map type, then the correct request is made.
> 
> I'll look into it tomorrow.

I quickly tried to reproduce and could not, fwiw.

dmfalke:
I'm having trouble testing this. Lots of 500s with various overlay variables. Can you provide some tips for testing?

I'm off tomorrow, but I can document configurations that cause a 500 early next week.

asizemore:
the apostrophe issue is already [here](https://github.com/VEuPathDB/EdaDataService/issues/274) i think?

packages/libs/eda/src/lib/core/utils/visualization.ts
```diff
@@ -259,11 +273,19 @@ export function assertValidInputVariables(
...
-        .map((input) => input.label)
-        .join(', ')}`
-    );
+    if (dataElementDependencyOrder != null) {
+      throw new Error(
+        `The variables are no longer valid due to entity constraints. Try disabling or changing the ${
+          invalidInputs[invalidInputs.length - 1].label
+        } variable.`
+      );
+    } else {
```
dmfalke, line 282: This logic seems flawed. For example, this condition will be met for a variable of an acceptable entity, but with too many values (such as when a filter is removed from a categorical variable with more than 8 values).

I think we will have to extract some logic from `disabledVariablesForInput` to be this specific.

dmfalke (commented):
The code looks good, and things seem to work, in my limited testing. Please see my comment below, in case I'm missing something.

packages/libs/eda/src/lib/core/utils/visualization.ts
dmfalke: > this condition will be met for a variable of an acceptable entity, but with too many values (such as when a filter is removed from a categorical variable with more than 8 values).

Oh, I guess this isn't an issue in the SAM, since we get the top 7+other values.

bobular: I tried to handle "non-entity constraints" first by calling the function from itself without the `dataElementDependencyOrder` - so the user *should* get a more helpful message if one of those constraints breaks first?

dmfalke: Oh, I see! I should have read the comments :flushed:.

dmfalke (approved):

PR #271, opened 2023-05-31
Show plot controls in floating visualizations

### LINKED ISSUE #189: Map: Floating plots need to have axes controls (bin widths, axes ranges, etc)
there is an issue https://github.com/VEuPathDB/web-monorepo/issues/99 which i dont honestly know how it relates. it *sounds* like that one is about the variable selection though, so im opening this as well.

This ticket is about adding controls that we have in the EDA for things like changing the bin width of histograms and line plots, changing the axes ranges and log scaling etc. These controls dont currently exist for the floating plots, but i see as a high priority. Without these, its very easy to build useless floaters and not be able to fix them.

@bobular suggests this will require some refactoring of the existing vizs?

chowington:
Fixes #189


bobular (approved):
Works great. I was expecting real-estate problems but it's fine for now!

Issue #239, opened 2023-05-21
Map, Trivial: ChartMarkers.tsx needs a default value for `reduce()`


bobular:
There's the same issue in ChartMarker.tsx, I'll use this comment to open a new issue.

_Originally posted by @bobular in https://github.com/VEuPathDB/web-monorepo/pull/211#discussion_r1199739519_

PR #272, opened 2023-05-31
Add default value for reduce()

chowington:
Fixes https://github.com/VEuPathDB/web-monorepo/issues/239


bobular (approved):
Tricky review - but I think I nailed it!

PR #277, opened 2023-06-01
Fix logic for rendering initial map menu

jernestmyers:
This is a quick follow up to #243. Currently, a page refresh when `appState` has an activeVizId and an activeMarkerType will render a map type's configuration menu but the submenu isn't highlighted properly. This fixes that logic while also using an `enum` to clean up some code, such that changes to a label's text could be made in one place now.


packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -792,7 +789,14 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
   }
 
   const intialActiveSideMenuId: string | undefined = (() => {
-    if (appState.activeVisualizationId) return MapSideNavItemLabels.MapType;
+    if (
+      appState.activeVisualizationId &&
+      appState.activeMarkerConfigurationType
+    )
```
dmfalke, line 795: I'm wondering if we should add a condition here to check that `MarkerTypeLabels[appState.activeMarkerConfigurationType]` is defined.

dmfalke (approved):
This looks good. I made a comment, but it's probably fine to ignore it.

PR #278, opened 2023-06-02
Fix build problem by making sure the typescript compiler knows import-export tokens are types

bobular:
Original compile issues with `yarn nx start @veupathdb/eda`

```
Compiled with problems:
ERROR in ../coreui/dist/assets/icons/index.js 70:0-69
export 'IconProps' (reexported as 'IconProps') was not found in './types' (module has no exports)
ERROR in ../coreui/dist/components/theming/index.js 3:0-34
export 'UITheme' (reexported as 'UITheme') was not found in './types' (module has no exports)
```

Looks like the compiler needs a hint about the reexport of a type.

Do we solve this here, or in the webpack/ts-config somehow? That's above my pay grade ;-) 

From my googling, I think `import type ...` and `export type ...` are typescript >= 3.8.x.

We are on 4.x so safe to proceed?

### DIFF
--- a/packages/libs/coreui/src/assets/icons/index.tsx
+++ b/packages/libs/coreui/src/assets/icons/index.tsx
@@ -1,10 +1,10 @@
-import React from "react";
-import { IconProps } from "./types";
-import Arrow from "./Arrow";
-import CaretDown from "./CaretDown";
-import CaretUp from "./CaretUp";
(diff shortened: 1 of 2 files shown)


jernestmyers (approved):
Works for me!

I tested the following environs:
- the `genomics-site` dev server
- the `eda` dev server
- coreui's storybook

Issue #182, opened 2023-05-09
Mosaic plot x-axis label has moved


jernestmyers:
It appears the x-axis label for mosaic plots is now above the graph. This affects the placement of the "required inputs" prompt, as shown in the screenshot.

![Image](https://github.com/VEuPathDB/web-monorepo/assets/69446567/b3916d00-a557-4027-a870-f00da029f9be)

I propose we fix the regression and render the x-axis label beneath the plot.

PR #280, opened 2023-06-03
Fix position of x axis title on mosaic plot

chowington:
Fixes https://github.com/VEuPathDB/web-monorepo/issues/182

Because the dummy x axis tick values were empty strings, the title had no labels to "automargin" against, so it was just positioned relative to the top of the plot. Changing the dummy labels to a single space gives it something to "automargin" against.

Removed tick angle because we don't actually have plotly-native tick labels.


jernestmyers (commented):
I'm still seeing this problem on the `clinepi-site` dev server. I'm not seeing it on `genomics-site` because the axis labels don't appear to render until all data is ready to go.

image

jernestmyers (approved):
Looks good! Thanks for fixing this 💪

Issue #279, opened 2023-06-02
Map - low-cardinality variables should not be sorted by frequency


bobular:
Do we want to bypass the frequency-based sorting for low-cardinality variables? My guess is yes...

Here is a six-value variable from **SCORE S. mansoni**

As a marker overlay
![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/845b4a3b-b007-469b-b583-5d340ec91ca0)

In subsetting/filtering
![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/3a44bb7e-855c-4c93-96b3-d0813ebf706d)

d-callan:
i think once we have advanced marker config showing things based on their frequency becomes the work of the markers configuration menu and never the legend

bobular:
Still an issue. Not a show stopper.

Agree with Danielle - the marker configuration somehow needs to freeze the order for low-cardinality vars (and disable table sorting?)


![Image](https://github.com/VEuPathDB/web-monorepo/assets/308639/d8da6c1d-9b58-4a44-9976-09c05b184686)

dmfalke:
> Still an issue. Not a show stopper.
> 
> Agree with Danielle - the marker configuration somehow needs to freeze the order for low-cardinality vars (and disable table sorting?)
> 
> ![Image](https://user-images.githubusercontent.com/308639/261623630-d8da6c1d-9b58-4a44-9976-09c05b184686.png)

I understood @d-callan's comment to be that the legend should not be sorted by frequency, but the marker preview table should continue to do so. I think I agree with this, just to keep the table consistent. I think we just need to make sure that the **Values** column is sorted based on the order from the variable's `vocabulary` property. Thoughts??

Issue #106, opened 2023-04-24
Map: Refactor: Introduce MapType front end plugin


dmfalke:
In the future, we will have non-marker map types. When that time comes, we will want to move marker-based logic out of `MapAnalysisImpl`, since not all map types will have markers.

Instead, we should introduce a `MapType` plugin:

- Each `MapType` defines a plugin to render a "config panel" and a "map type layer"
- The "config panel" is what appears in the left-hand menu
- The "map type layer" is rendered on top of the map, and includes markers, legend, floater, etc
- The MapAnalysis component will simply render a <ActiveMapTypeLayer {...activeMapTypeConfig}/>, replacing all of the existing marker and legend logic
- It's likely that all of the marker-based map types will share a `MarkerMapTypeLayer` component

d-callan:
to add a bit more detail to this, bubbles will have two parts to their legend. the first will indicate size and likely be a set of concentric circles, the second will indicate color and at least for whats been discussed so far will require a gradient legend.

d-callan:
see https://github.com/VEuPathDB/web-monorepo/issues/249

dmfalke:
I propose we introduce `MapTypePlugin`, which will include a component to render the configuration panel, and a component to render a "map layer" (which will include a legend, markers, floating vizs, etc).

This approach will make it possible to have non-marker map types (such as animation, time-based view, etc).

The map itself would continue to be rendered independently.


@bobular I would love you feedback on this proposal.

dmfalke:
I'm going to rename this issue and move it back to "TODO". I also don't think it's needed for NIAID QO

d-callan:
Just thinking if we want to be really flexible and future proof, then rendering the base map itself might be the purview of the 'map type'.

dmfalke:
> Just thinking if we want to be really flexible and future proof, then rendering the base map itself might be the purview of the 'map type'. 

I had a similar thought. I'm not opposed to that. Can you think of use cases? Maybe if a map type wants to use a different base layer?

d-callan:
I was thinking of choropleth and cartograms, etc

dmfalke:
Cool, makes sense. I'm going to leave this here for future reference: https://leafletjs.com/examples/choropleth/

bobular:
Even though this isn't urgent any more I like the idea of future proofing. Dave's proposal looks good!

Choropleths still go on top of a regular map, so I don't think the base layers need to be part of the plugin scheme.

[...]

PR #285, opened 2023-06-07
Implement advanced map marker configuration menu

### LINKED ISSUE #190: Map: Update 'Markers' menu entry to allow more advanced user configuration
Additional tweaks from [2023-06-15 Map UX meeting](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/300613633/2023-06-15+Map+UX):

- [x] try to use the same layout as regular visualizations for the marker config panel:
      - variable picker at top
         - can we give this a `max-width`?
         - not discussed at meeting, but if it's at the very top, does it really need a "Color:" heading? (For continuous vars, is the color gradient really a color? maybe)
      - plot next (in this case it's the marker preview)
      - controls
      - value picker distributions (checkbox table or histogram)
- [x] Heading for preview is now "Summary marker (all filtered data)" (plus possible tooltip to explain that it's visible plus any other filtered data outside the viewport?)
- [x] Controls subheadings
      - chart marker: Marker X-axis controls (continuous only: contains "binning method"), Marker Y-axis controls (contains: count/proportion radio button and logscale toggle).  The extra "Marker" in the headings might be overkill but let's see it first. 
      - donut marker: Doesn't have an X-axis. We just need to have a binning method selector for continuous variables. Do we just use a "Donut controls" heading? Bit of a grey area...
- [x] Binning method selector implementation notes:  the `ContinousVariableMetadataResponse` contains three types of binRanges: `equalInterval`, `quantile` (actually decile) and `standardDeviation`.  We are currently hardcoded to use `equalInterval` somewhere. Should be simple to use the other two types based on the radio button-controlled state.
- [x] For categorical variables, put a box around the checkbox table (and add a scroll bar) - see [figma](https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?type=design&node-id=1026%3A40876&t=qHZatybYd01dHdXe-1) - color up to you!
- [x] We didn't decide on a search box in the checkbox table - exclude for now.
- [x] Add a radio button control inside the box under the table, "Show counts for :white_circle: filtered :black_circle: visible" (categorical only)
- [x] I'm 95% sure we decided this radio button will NOT affect the summary marker (that's why we put the box around the table)
- [x] #313
- [x] Continuous reference distribution
      - try labelling it "Raw distribution of <Variable name> over all filtered data" (tooltip could mention "equal-binned" ??)
      - color uniform #333 (add this to HistogramProps: `colorPalette: Array(8).fill('#333')`)
      - reduce margins (`spacingOptions` should hopefully do it)

More categorical tweaks

- [ ] debounce with 2s delay
- [x] confirmation of previous plans to deal with >7 selections
      - allow any number of checkbox selections
      - but only use the last valid combination of checkboxes for the overlayConfig
      - _overlay_ a warning banner on top of the Summary marker (aka preview marker) to say "maximum 7 selections allowed, please deselect X values" (summary marker should always be visible, hence the scroll bar for the checkbox table) 
- [x] investigate issue with summary marker containing stale data after filters have been changed (Bob can demo)

General menu-related
- [x] implement the decision of this thread https://epvb.slack.com/archives/CMQ8K9XRS/p1686863706963029 ("Map Type" menu verbiage)

mockups exist in [figma](https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?type=design&node-id=0-1&t=oTp9KkfnBRYMg6CP-0).. ex screenshots as well

categorical
image

continuous
image

**NOTE:** I believe some changes were discussed after these mockups were made, but idk if the mockups were updated to reflect them. For ex, The histogram shown when configuring a continuous variable will always reflect the subset, rather than offering the visible vs filtered toggle.

@asizemore @bobular do you know of any other changes discussed which are important to note here?

jernestmyers:
Resolves #190

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -2,6 +2,7 @@ import { ReactNode, useCallback, useEffect, useMemo, useState } from 'react';
 
 import {
   AnalysisState,
+  CategoricalVariableDataShape,
   DEFAULT_ANALYSIS_NAME,
   EntityDiagram,
   OverlayConfig,
@@ -257,17 +258,26 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
   // get the default overlay config.
   const activeOverlayConfig = usePromise(
     useCallback(async (): Promise<OverlayConfig | undefined> => {
-      // TODO Use `selectedValues` to generate the overlay config. Something like this:
-      // if (activeMarkerConfiguration?.selectedValues) {
-      //   return {
-      //     overlayType: CategoryVariableDataShape.is(overlayVariable?.dataShape) ? 'categorical' : 'continuous',
-      //     overlayVariable: {
-      //       variableId: overlayVariable.id,
-      //       entityId: overlayEntity.id,
-      //     },
-      //     overlayValues: activeMarkerConfiguration.selectedValues
-      //   } as OverlayConfig
-      // }
+      // Use `selectedValues` to generate the overlay config
+      if (
+        activeMarkerConfiguration?.selectedValues &&
+        activeMarkerConfiguration?.type === 'barplot'
+      ) {
+        return {
+          overlayType: CategoricalVariableDataShape.is(
+            overlayVariable?.dataShape
+          )
+            ? 'categorical'
+            : 'continuous',
+          overlayVariable: {
+            variableId: overlayVariable?.id,
+            entityId: overlayEntity?.id,
+          },
+          overlayValues: activeMarkerConfiguration.selectedValues,
+          allValuesSorted: activeMarkerConfiguration.allValues,
+        } as OverlayConfig;
+      }
+
       return getDefaultOverlayConfig({
         studyId,
         filters,
@@ -283,6 +293,8 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
       overlayVariable,
       studyId,
       subsettingClient,
+      activeMarkerConfiguration?.selectedValues,
+      activeMarkerConfiguration?.type,
     ])
   );
 
@@ -510,6 +522,7 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
                     toggleStarredVariable={toggleStarredVariable}
                     configuration={activeMarkerConfiguration}
                     constraints={markerVariableConstraints}
+                    overlayConfiguration={activeOverlayConfig.value}
                   />
                 ) : (
                   <></>
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -261,7 +261,7 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
       // Use `selectedValues` to generate the overlay config
       if (
         activeMarkerConfiguration?.selectedValues &&
-        activeMarkerConfiguration?.type === 'barplot'
+        activeMarkerConfiguration?.allValues
       ) {
         return {
           overlayType: CategoricalVariableDataShape.is(
@@ -294,7 +294,7 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
       studyId,
       subsettingClient,
       activeMarkerConfiguration?.selectedValues,
-      activeMarkerConfiguration?.type,
+      activeMarkerConfiguration?.allValues,
     ])
   );
 
@@ -497,6 +497,7 @@ function MapAnalysisImpl(props: Props & CompleteAppState) {
                     }
                     toggleStarredVariable={toggleStarredVariable}
                     constraints={markerVariableConstraints}
+                    overlayConfiguration={activeOverlayConfig.value}
                   />
                 ) : (
                   <></>
--- a/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
@@ -5,7 +5,7 @@ import {
 import RadioButtonGroup from '@veupathdb/components/lib/components/widgets/RadioButtonGroup';
 import { VariableDescriptor } from '../../../core/types/variable';
 import { VariablesByInputName } from '../../../core/utils/data-element-constraints';
-import { AllValuesDefinitions, OverlayConfig } from '../../../core';
+import { AllValuesDefinition, OverlayConfig } from '../../../core';
 import { CategoricalMarkerConfigurationTable } from './CategoricalMarkerConfigurationTable';
 
 interface MarkerConfiguration<T extends string> {
@@ -17,7 +17,7 @@ export interface BarPlotMarkerConfiguration
   selectedVariable: VariableDescriptor;
   selectedValues: OverlayConfig['overlayValues'] | undefined;
   selectedPlotMode: 'count' | 'proportion';
-  allValues: AllValuesDefinitions | undefined;
+  allValues: AllValuesDefinition[] | undefined;
 }
 
 interface Props
@@ -99,6 +99,7 @@ export function BarPlotMarkerConfigurationMenu({
         <CategoricalMarkerConfigurationTable
           overlayConfiguration={overlayConfiguration}
           configuration={configuration}
+          // @ts-ignore
           onChange={onChange}
         />
       )}
--- a/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
@@ -7,6 +7,7 @@ import { VariableDescriptor } from '../../../core/types/variable';
 import { VariablesByInputName } from '../../../core/utils/data-element-constraints';
 import { AllValuesDefinition, OverlayConfig } from '../../../core';
 import { CategoricalMarkerConfigurationTable } from './CategoricalMarkerConfigurationTable';
+import { MarkerPreview } from './MarkerPreview';
 
 interface MarkerConfiguration<T extends string> {
   type: T;
@@ -62,6 +63,7 @@ export function BarPlotMarkerConfigurationMenu({
 
   return (
     <div>
+      <MarkerPreview data={overlayConfiguration} />
       <RadioButtonGroup
         containerStyles={{
           marginTop: 20,
--- a/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
@@ -1,4 +1,3 @@
-import { H6 } from '@veupathdb/coreui';
 import {
   InputVariables,
   Props as InputVariablesProps,
@@ -6,7 +5,8 @@ import {
 import RadioButtonGroup from '@veupathdb/components/lib/components/widgets/RadioButtonGroup';
 import { VariableDescriptor } from '../../../core/types/variable';
 import { VariablesByInputName } from '../../../core/utils/data-element-constraints';
-import { BinDefinitions } from '../../../core';
+import { AllValuesDefinitions, OverlayConfig } from '../../../core';
+import { CategoricalMarkerConfigurationTable } from './CategoricalMarkerConfigurationTable';
 
 interface MarkerConfiguration<T extends string> {
   type: T;
@@ -15,8 +15,9 @@ interface MarkerConfiguration<T extends string> {
 export interface BarPlotMarkerConfiguration
   extends MarkerConfiguration<'barplot'> {
   selectedVariable: VariableDescriptor;
-  selectedValues: BinDefinitions | undefined;
+  selectedValues: OverlayConfig['overlayValues'] | undefined;
   selectedPlotMode: 'count' | 'proportion';
+  allValues: AllValuesDefinitions | undefined;
 }
 
 interface Props
@@ -26,6 +27,7 @@ interface Props
   > {
   onChange: (configuration: BarPlotMarkerConfiguration) => void;
   configuration: BarPlotMarkerConfiguration;
+  overlayConfiguration: OverlayConfig | undefined;
 }
 
 export function BarPlotMarkerConfigurationMenu({
@@ -35,6 +37,7 @@ export function BarPlotMarkerConfigurationMenu({
   toggleStarredVariable,
   configuration,
   constraints,
+  overlayConfiguration,
 }: Props) {
   function handleInputVariablesOnChange(selection: VariablesByInputName) {
     if (!selection.overlayVariable) {
@@ -59,6 +62,18 @@ export function BarPlotMarkerConfigurationMenu({
 
   return (
     <div>
+      <RadioButtonGroup
+        containerStyles={{
+          marginTop: 20,
+        }}
+        label="Y-axis"
+        selectedOption={configuration.selectedPlotMode || 'count'}
+        options={['count', 'proportion']}
+        optionLabels={['Count', 'Proportion']}
+        buttonColor={'primary'}
+        margins={['0em', '0', '0', '1em']}
+        onOptionSelected={handlePlotModeSelection}
+      />
       <p
         style={{
           paddingLeft: 7,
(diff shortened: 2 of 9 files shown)


packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
```diff
@@ -26,16 +37,73 @@ interface Props
...
   onChange: (configuration: BarPlotMarkerConfiguration) => void;
   configuration: BarPlotMarkerConfiguration;
+  overlayConfiguration: OverlayConfig | undefined;
+  overlayVariable: Variable | undefined;
+  subsettingClient: SubsettingClient;
+  studyId: string;
+  filters: Filter[] | undefined;
 }
 
+// TODO: generalize this and PieMarkerConfigMenu into MarkerConfigurationMenu. Lots of code repitition...
```
adnauseum, line 47: https://blog.codinghorror.com/rule-of-three/

jernestmyers: Perhaps I'll change it to a "maybe do" 😆

adnauseum: 🙃  
Just wanted to derp by the water color 👋

bobular:
Reviewing this now.

https://github.com/VEuPathDB/web-monorepo/issues/279 seems important as an "up next"?

image

bobular (commented):
I think the big nut to crack is the checkbox column heading behaviour in `CategoricalMarkerConfigurationTable`

At the moment it seems to behave erratically. With SCORE S. mansoni "occupation" it will create an endless loop (or at least for the number of items).  If "Peasant" is unselected first, then the very last value "Transporter" gets selected.

With low cardinality variables it seems to deselect the last value only.

Some options:

1. somehow hack the Mesa/DataTable component to disable the heading checkbox (I looked already there's no easy prop to do this)
2. keep the heading checkbox but make the first click **deselect** all (also requires some hacking I think) - and when it is clicked again, don't show endless warnings.
3. gate off the value selection behind [:heavy_check_mark:  apply] [:x: cancel] buttons, and leave the table implementation alone

I think something is a bit off with the `handleSelection` and `handleDeselection` handlers but @dmfalke probably understands quicker.  They are called in a `forEach` which explains the multiple alerts.

```
  selectAllRows() {
    const { rows, options, eventHandlers } = this.props;
    const { isRowSelected } = options;
    const { onRowSelect, onMultipleRowSelect } = eventHandlers;
    const unselectedRows = rows.filter((row) => !isRowSelected(row));
    if (onMultipleRowSelect) return onMultipleRowSelect(unselectedRows);
    return unselectedRows.forEach(onRowSelect);
  }
```

bobular:
Unless perhaps we can use the `onMultipleRowSelect` and `onMultipleRowDeselect` handlers?

However these are only applied to unselected and selected rows respectively. I think we'd need our putative `toggleSelectAll` handler to know about selected and unselected rows.

jernestmyers:
> Unless perhaps we can use the `onMultipleRowSelect` and `onMultipleRowDeselect` handlers?
> 
> However these are only applied to unselected and selected rows respectively. I think we'd need our putative `toggleSelectAll` handler to know about selected and unselected rows.

The latest push makes use of `onMultipleRowSelect` and `onMultipleRowDeselect`. More specifically...

- onMultipleRowSelect is currently selecting all values, which makes the legend and the markers look quite strange
- onMultipleRowDeselect sets `selectedValue` to be only `__UNSELECTED__`, which renders as `All other values` in the legend; this handler is also now triggered when the indeterminate-state checkbox is clicked in the header.

Decisions that remain:
How to handle selecting too many values?
- Should we add intermediate state that can be used to show the boxes as checked while we continue to aggregate any "excess" selections as `__UNSELECTED__`? We would still need to communicate to the use that, "Hey, we can only select X unique values. All others will be aggregated as `All other values`."
- 🤷

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -333,9 +351,72 @@ function MapAnalysisImpl(props: ImplProps) {
...
+            ? { color: prevData[index].color }
+            : 'color' in data
+            ? { color: data.color }
+            : {}),
+        })),
+      initialDataObject
+    );
+    if (markerType === 'pie') {
+      return (
+        <DonutMarkerStandalone
```
bobular, line 399: Ideally we need to add something like `markerLabel: kFormatter(finalData.reduce(...sum the values...))` to the props here, so that the number formatting is the same (without `markerLabel`, the marker components fall back to doing a sum reduce of the values and render the number as a plain old number).

But this isn't very nice. It would be better I think to get rid of the `markerLabel` prop from the marker components pass a formatting function to the markers instead.  (I added `markerLabel` - my bad!)  But we could do this as technical debt if you prefer.

Aiming for this (hacked in dev tools)
image

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -333,9 +351,72 @@ function MapAnalysisImpl(props: ImplProps) {
...
+          data={finalData}
+          {...sharedStandaloneMarkerProperties}
+        />
+      );
+    } else {
+      return (
+        <ChartMarkerStandalone
+          data={finalData}
```
bobular, line 407: and `mFormatter` here

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
```diff
@@ -80,18 +167,93 @@ export function BarPlotMarkerConfigurationMenu({
...
+        <span style={{ fontWeight: 'bold' }}>
+          Summary marker (all filtered data)
+        </span>
+        {overlayConfiguration?.overlayType === 'categorical' ? (
+          <MarkerPreview data={overlayConfiguration} mapType="barplot" />
+        ) : (
+          continuousMarkerPreview
+        )}
+      </div>
+      <LabelledGroup label="Marker X-axis controls">
```
bobular, line 182: The controls get very wide - on my screen the menu takes up more than half so I can't see the map's spinner any more.
I'm no CSS expert but adding this
`containerStyles={{ display: 'flex' }}`
to the `LabelledGroup`s (and the next one) at least makes them wrap down the screen. I'd recommend that as a temporary measure before addressing the styling of the controls, headings etc.

PS. The LabelledGroup component is all my fault - not sure how we would style the labels within them (e.g. "Binning method")

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
```diff
@@ -68,6 +149,71 @@ export function PieMarkerConfigurationMenu({
...
+        <span style={{ fontWeight: 'bold' }}>
+          Summary marker (all filtered data)
+        </span>
+        {overlayConfiguration?.overlayType === 'categorical' ? (
+          <MarkerPreview data={overlayConfiguration} mapType="pie" />
+        ) : (
+          continuousMarkerPreview
+        )}
+      </div>
+      <LabelledGroup label="Donut marker controls">
```
bobular, line 163: Need to make this render only for continuous vars?

bobular (commented):
Excellent progress - looking good.

High cardinality categoricals don't seem ready for review yet, so I've left them, but made some other comments which I'd better submit now before they go stale.

Have a great weekend!

bobular:
Sorry, meant to add one more comment:

The logscale mode for the preview bar marker doesn't seem to be right.

Here are no-logs markers - the preview looks similar to the three map markers:
image

Here are logged markers - now they don't look at all similar:
image

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
jernestmyers: Maybe? Unclear it we'd rather disable them or remove them. Removing them has the benefit of uncluttering the UI but conversely results in page jumps and less consistency. Currently, I'm disabling binning controls in both marker config menus for categorical vars.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -272,28 +283,110 @@ function MapAnalysisImpl(props: ImplProps) {
     [markerConfigurations, setMarkerConfigurations]
   );
 
+  const filtersIncludingViewport = useMemo(() => {
```
jernestmyers, line 286: **NOTE**: No logic/code has changed for `filtersIncludingViewport` but I had to move its initialization above `allVisibleCategoricalValues`.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -333,9 +428,73 @@ function MapAnalysisImpl(props: ImplProps) {
...
+            activeMarkerConfiguration &&
+            'dependentAxisLogScale' in activeMarkerConfiguration
+              ? activeMarkerConfiguration.dependentAxisLogScale
+              : false
+          }
+          {...sharedStandaloneMarkerProperties}
+        />
+      );
+    }
+  }, [previewMarkerData]);
```
bobular, line 496: It may be possible to move the logic for generating `continuousMarkerPreview` inside the `*MarkerConfigurationMenu` components (as a utility function, I imagine). It looks like all but one (`geoConfig`) of the props needed to make `previewMarkerData` are already passed to the `*MarkerConfigurationMenu` components, so it wouldn't involve a lot of "prop bloat". In fact we'd be removing one prop (`continuousMarkerPreview`) and adding `geoConfig`, so breaking even!

It would clean things up a bit at this outer level.

packages/libs/eda/src/lib/map/analysis/appState.ts
```diff
@@ -4,6 +4,7 @@ import * as t from 'io-ts';
 import { isEqual } from 'lodash';
 import { useCallback, useEffect } from 'react';
 import {
+  AllValuesDefinition,
```
bobular, line 7: My IDE says not used :point_up:

packages/libs/eda/src/lib/map/analysis/appState.ts
```diff
@@ -28,12 +29,35 @@ export const MarkerConfiguration = t.intersection([
   t.union([
     t.type({
       type: t.literal('barplot'),
-      selectedValues: t.union([BinDefinitions, t.undefined]), // user-specified selection
+      selectedValues: t.union([BinDefinitions, t.array(t.string), t.undefined]), // user-specified selection
```
bobular, line 32: I'm not quite sure what's going on here. Doesn't the `pie` version below also need `BinDefinitions`?  Or maybe neither need it (nope - tried that and it doesn't compile).  I think I wrote the original types, so I should know. Need lunch though, so going to hit "send"!

bobular (approved):
This is looking and working great.

If you choose not to follow any of my suggestions please make tickets for them.

I'll come back laters to the io-ts query I made in appState.ts

packages/libs/eda/src/lib/map/analysis/appState.ts
bobular: Well, I still don't get it, but I'll have to move on.

Both barplot and pie markers can display continuous variables, and so both should need to be able to store the `BinDefinitions` in the config. Or at least they would if completely custom user-defined bins were allowed (no plans for this, don't worry).

But yeah, right now the `markerConfigurations` are not storing `selectedValues` for continuous variables. See what I grabbed from dev tools below.  Both pie and barchart configurations have `undefined` `selectedValues`:

```
[
  {
    "type": "pie",
    "selectedVariable": {
      "entityId": "EUPATH_0043226",
      "variableId": "EUPATH_0030018"
    },
    "selectedCountsOption": "filtered",
    "binningMethod": "quantile"
  },
  {
    "type": "barplot",
    "selectedPlotMode": "count",
    "selectedVariable": {
      "entityId": "EUPATH_0043226",
      "variableId": "EUPATH_0030017"
    },
    "dependentAxisLogScale": true,
    "selectedCountsOption": "filtered",
    "binningMethod": "quantile"
  }
]
```

bobular: So I'm pretty sure the `BinDefinitions` shouldn't be needed in line 32 above, but will require some other changes to make it compile.

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/CategoricalMarkerConfigurationTable.tsx
```diff
@@ -0,0 +1,244 @@
...
+       * We want "All other values" label to be at the end, so if it exists:
+       *  - delete it from the Set
+       *  - add it back to the end of the Set
+       */
+      if (nextSelections.has(UNSELECTED_TOKEN)) {
+        nextSelections.delete(UNSELECTED_TOKEN);
+        nextSelections.add(UNSELECTED_TOKEN);
+      }
+      const newSelectedValues: string[] = [];
+      nextSelections.forEach((v) => newSelectedValues.push(v));
```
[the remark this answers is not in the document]
jernestmyers: Nope, this would actually be _**a lot**_ easier to read! 😄

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
dmfalke: I wonder if this should wait for https://github.com/VEuPathDB/web-monorepo/issues/106?

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
```diff
@@ -26,16 +38,105 @@ interface Props
...
+    new Set(
+      overlayConfiguration?.overlayType === 'categorical'
+        ? overlayConfiguration?.overlayValues
+        : undefined
+    )
+  );
+  useEffect(() => {
+    if (overlayConfiguration?.overlayType !== 'categorical') return;
+    setUncontrolledSelections(new Set(overlayConfiguration?.overlayValues));
+  }, [overlayConfiguration?.overlayValues, overlayConfiguration?.overlayType]);
```
dmfalke, line 91: You can remove the `useEffect` here. See https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes.

```suggestion
  if (overlayConfiguration?.overlayType === 'categorical') {
    setUncontrolledSelections(new Set(overlayConfiguration?.overlayValues));
  }
```

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
```diff
@@ -1,35 +1,138 @@
...
+    new Set(
+      overlayConfiguration?.overlayType === 'categorical'
+        ? overlayConfiguration?.overlayValues
+        : undefined
+    )
+  );
+  useEffect(() => {
+    if (overlayConfiguration?.overlayType !== 'categorical') return;
+    setUncontrolledSelections(new Set(overlayConfiguration?.overlayValues));
+  }, [overlayConfiguration?.overlayValues, overlayConfiguration?.overlayType]);
```
dmfalke, line 87:
```suggestion
  if (overlayConfiguration?.overlayType === 'categorical') {
    setUncontrolledSelections(new Set(overlayConfiguration?.overlayValues));
  }
```

packages/libs/eda/src/lib/map/analysis/utils/categoricalValues.ts
```diff
@@ -0,0 +1,72 @@
...
+    const unfilteredDistributionResponse =
+      await subsettingClient.getDistribution(
+        studyId,
+        overlayEntity.id,
+        overlayVariable.id,
+        {
+          valueSpec: 'count',
+          filters: [],
+        }
+      );
```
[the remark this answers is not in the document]
dmfalke: I don't have any suggestions right now, but I will think about it some more.

dmfalke (changes requested):
This is all looking really good. I reviewed the code and made a couple of suggestions. Feel free to comment if you disagree.

I will continue to think about the best way caching calls to the distribution service, but I think we should make it a follow-up task.

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
jernestmyers: This suggestion results in an infinite loop, which is why I think I reached for the `useEffect`. But it does appear that the "Better" suggestion from that section of the React docs works. Pretty clunky looking though!
```js
  const [previousOverlays, setPreviousOverlays] = useState(overlayConfiguration?.overlayValues);
  if (
      previousOverlays !== overlayConfiguration?.overlayValues && 
      overlayConfiguration?.overlayType === 'categorical'
    ) {
    setUncontrolledSelections(new Set(overlayConfiguration?.overlayValues));
    setPreviousOverlays(overlayConfiguration.overlayValues);
  }
```
Thoughts on this?

dmfalke: Ah, yes, I see. This is more verbose, but also more efficient. Since this logic will be used in at least two places, maybe extract it into a hook. You can also add the initialization of `uncontrolledSelections` in the hook and call it `useUncontrolledSelections`. That will make it easier for the reader to ignore the implementation details, if they want to.

dmfalke (approved):

bobular:
🎉

PR #288, opened 2023-06-07
Check that the current study has geoconfigs

dmfalke:
This PR adds a check for geoconfigs in the `MapAnalysis` component. If none are found, then it renders an error message indicating as much:
image

It's not the prettiest thing, but I don't expect this to be an issue in production.


bobular (approved):
Nice! I'd have thrown an error or something way too complicated.

Tested and works for me!

PR #289, opened 2023-06-07
Upgrade to nx@16

dmfalke:
This PR upgrades `nx` to the latest version (16.3.2).

I've tested with the following commands, all of which work:
- `yarn nx start @veupathdb/clinepi-site`
- `yarn nx start @veupathdb/eda`
- `yarn nx bundle:dev @veupathdb/clinepi-site`

Note that `migrations.json` should be deleted, but I kept it based on [this comment](https://nx.dev/core-features/automate-updating-dependencies#step-3-cleaning-up):
> Note: You may want to keep the migrations.json until every branch that was created before the migration has been merged. Leaving the migrations.json in place allows devs to run nx migrate --run-migrations to apply the same migration process to their newly merged code as well.


dmfalke:
I updated coreui to put generated code into `lib`, rather than `dist`. This makes all projects similar, which allows us to set the `targetDefaults.outputs` property in `nx.json`, which appears to improve cache detection.

jernestmyers (approved):
LGTM

Issue #131, opened 2023-04-28
Map: Timeline + marginal histogram floating visualization


d-callan:
This is needed for the tick QO, deadline sept.

This ticket is to build a floating visualization for the standalone map, using the component built in #130. The line and histo should share xaxis controls and the xaxis should be identical for both. This means the xaxis of the lineplot MUST be binned, where a typical line plot would have the option. 

The backend strategy, as of now, is to continue to use the two existing endpoint for line and histo separately. if necessary, a backend ticket will be filed to ensure consistent bins. Given that  a) the existing histogram supports semantic zooming and the line plot doesnt and  b) this histogram being marginal means it will only represent counts for non-NULL values of the line plot's yaxis variable.. I think the best strategy for requesting data would be to build the line plot request as usual, and from the response parse the default bin width. Then make a second request to histogram using that bin width, while also adding a filter for all values (non-NULLs) of the yaxis variable of the line plot. I suppose though if the lineplot yaxis already has been filtered on by the user, then no need to add the filter though.

If any of that ^ is unclear or we dont think will work for some reason, let me know.

EDIT: after reading the above back, I'd like to clarify the histo request. It should take the existing filters on the yaxis variable of the line plot and add to them wherever necessary to exclude values which were also excluded from the denominator. And it should use both the bin width and viewport the line plot used.

bobular:
Makes sense to me :+1: 

So this would be a new visualisation type (from the client/user perspective)?

I think that's probably a good idea. The alternative is to have the marginal plot optional and only an option when binning is on for continuous/date vars. I think that's probably more cognitive overload than having a new plot type.

d-callan:
Agree. I intentionally didn't make a ticket to add the marginal histo to the existing line plot viz, for some of the reasons you list.

d-callan:
outside of binning being required i think the specs for this are the same as the floating lineplot. in particular, the variable inputs are the same.

Issue #130, opened 2023-04-28
Map: Timeline + marginal histogram component


d-callan:
This is needed for the tick QO, deadline sept.

Current plan is to add a floating chart which is primarily a timeline (lineplot w temporal xaxis) with a marginal histogram. This ticket is to build the component, not the visualization. It likely requires building a new lineplot component in visx with the same specs as the existing plotly component, with the exception that it has a linked histogram sharing its xaxis. Mockups with the exact specs will be available soon.

asizemore:
duplicate of #96 ?

d-callan:
an ex of what were aiming for:

![timeline_specs](https://user-images.githubusercontent.com/53306535/236018165-457219dd-dada-4020-95d9-95cdab628849.png)


* the axis should be placed in between the line and histo
* histogram is upright
* both line and histo should be colored by overlay, the height of the histo reflects counts of non-NULL values for all variables relevant to the line plot
* Ideally, this will have a dedicated tooltip. if time doesnt allow that we can do the same as we did for line for now. ill make a separate ticket for the tooltip and reference it here.
* no need to consider semantic zoom

d-callan:
talking w @asizemore today, we realized itd make a lot of sense for the histo to be optional in the component and required for the visualization

moontrip:
I have tested something but I wonder if we can go for multiple plots with two y-axes (left and right) in a single figure using plotly, instead of separate two plots: attached a screenshot: didn't make a component but tested at Codepen. In this case, it would probably be faster to implement than Visx and an additional advantage is that we can use tooltip relatively easy as also shown in the screenshot: tooltip contents can be adjusted as we did for other Vizs. 

![two y-axis01](https://github.com/VEuPathDB/web-monorepo/assets/12802305/b8871a96-273a-4a25-9b44-bb3337f54d56)

I am also checking/learning Visx approach as well.

moontrip:
Thank you for your sharing your thought @d-callan. If so, as far as I understand, basically we need to make two separate plots, one is line/scatter and the other is histogram anyway. So what if we just use plotly to plot two such different types of plots at up and down?

d-callan:
We also need the histo to respond to the x-axis controls as well, without actually having an xaxis.

moontrip:
> @moontrip do you know if plotly allows having a tooltip that spans multiple plots? Eventually we'd want the tooltip to span both the line and histogram parts of this visualization

[...]

Issue #151, opened 2023-05-03
Visualizations: dedicated tooltip for timeline + histo component


d-callan:
See an ex here:

![timeline_tooltip](https://user-images.githubusercontent.com/53306535/236020422-32c19a1a-b1de-48ea-b2d4-0e3e3b941538.png)

* x and y values and CIs for the points in the line go in the line plot tooltip, as usual.
* n value goes below near the histogram bin associated w the same point in the line, rather w the line tooltip as previously done.
* a dashed line appears connecting the point in the line w the associated histo bin
* all three (two tooltips and line) appear when you hover over either the point in the line or the histo bin

PR #290, opened 2023-06-08
130 lineplot marginal histogram

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/130 and https://github.com/VEuPathDB/web-monorepo/issues/131. Screenshots will be added in the following comment: tested many cases and all seem to work properly.

Two new props are added to Lineplot component, `showMarginalHistogram` and `marginalHistogramSize`. The former decides to display marginal histogram in the Lineplot Viz and the latter is for the size of marginal histogram that has default value of 0.2 (20 %): Lineplot will take 80 %. Perhaps we need to adjust `showMarginalHistogram` prop to distinguish Lineplot Viz from Lineplot with marginal histogram Viz.

Though, I am not quite sure where to place marginal histogram, top or bottom: a mockup suggested placing marginal histogram at the bottom of Lineplot, but examples we have seen and discussed in the meeting showed top area: personally, it looks better at top area.


moontrip:
- ### GEMS1 with overlay using defalut binWidth
marginal histogram30


- ### same to above but unchecked 24-59 months in the legend
marginal histogram31


- ### same to above but with a facet variable
marginal histogram32


- ### Same to above but binWidth is increased and custom axis ranges are applied to check
marginal histogram4


- ### testing filter for Age variable
marginal histogram51


- ### Lineplot 0/0 case: example is taken from https://github.com/VEuPathDB/web-eda/issues/1082
marginal histogram1

dmfalke:
> Though, I am not quite sure where to place marginal histogram, top or bottom: a mockup suggested placing marginal histogram at the bottom of Lineplot, but examples we have seen and discussed in the meeting showed top area: personally, it looks better at top area.

I think one reason to put it on the bottom is so that is it closer to the x-axis ticks. Maybe @d-callan or @asizemore can weigh in?

d-callan:
It looks nice above. Can we duplicate the x-axis at the top of the line plot/ below the histo easily?

moontrip:
> It looks nice above. Can we duplicate the x-axis at the top of the line plot/ below the histo easily?

Thank you for your comment @d-callan 👍 I will try to see if I can add the x-axis.

moontrip:
@d-callan Here are some test results. Hope this will be useful for making a decision.
I have firstly checked a possibility of a tooltip described at https://github.com/VEuPathDB/web-monorepo/issues/151, but what I could achieve so far is to display a vertical line to cover up both plots. Tooltip contents are separately made depending on where a hover is located. Please see below two screenshots. Actually lineplot's tooltip contains the number of count (n).

- hovering histogram
marginal histogram xaxis2 label04

- hovering lineplot
marginal histogram xaxis2 label03

I have also tested whether we can have xaxis tick labels at the bottom of histogram. I could manage to find a way to do that, but one disadvantage is that above tooltip cannot be used with this tick labels. It is because additional xaxis tick labels require additional definition of xaxis, saying xaxis2. Anyway please see below screenshots

- placing xaxis tick labels
marginal histogram xaxis2 label01

- Above may be okay but there is a possibility that lineplot's line/marker may be overlapped with those axis tick labels in some cases: though it is most likely okay because we have set that lineplot (and scatterplot) has some internal margin at the yaxis min and max area to avoid clipping of markers around those area. 
Anyway, I found a way to avoid such a case by slightly decrease Lineplot size, like below screenshot. Total plot sizes between above and below plots are identical, but lineplot size below has smaller size than the above, as you may notice.
marginal histogram xaxis2 label02

d-callan:
Thanks @moontrip. If we can't guarantee that both x axes are aligned then I favor the tooltip w dashed line. I think it's still ok to put the histo above. Nice job 👍

moontrip:
> Thanks @moontrip. If we can't guarantee that both x axes are aligned then I favor the tooltip w dashed line. I think it's still ok to put the histo above. Nice job 👍

@d-callan Thank you for your comment! After testing something more, I have one update concerning the positional difference of tick labels between xaxis and xaxis2, as seen earlier. I did not test many cases yet but such a difference is probably not an issue when drawing the plot via Viz because we have set the axis range: in the storybook examples above, I did not consider the range. Here is an example at Viz with the same data used at the storybook and the tick label positions seem to be identical each other. If you prefer aligned xaxis label, then I will check more cases carefully and implement it in this PR: if this is the case, one additional question is whether you want to have a gap in between lineplot and marginal histogram like below screenshot or without the gap.

marginal histogram xaxis2 label05

d-callan:
Can we see the regular tooltips? Remind me what they look like if we go for the duplicated x axis?

moontrip:
> Can we see the regular tooltips? Remind me what they look like if we go for the duplicated x axis?

@d-callan Yes, we will get regular tooltips that can be seen at either Histogram Viz (hovering histogram) or Lineplot Viz (hovering lineplot) if we use two individual x-axis. Below is an example for hovering Lineplot Viz's marker.

marginal histogram xaxis2 label07

bobular:
Reviewing now - but I suggest that linked tooltips be "future growth" i.e. another ticket.

bobular:
Do we prefer a tiny bit of spacing between the plots?

image

d-callan:
I think the standard tooltip w duplicate x axis is fine, and that we should start without the gap between the two plots and see how much of an issue it really is in practice.

@bobular it's already a second ticket.

packages/libs/components/src/plots/LinePlot.tsx
```diff
@@ -270,41 +296,49 @@ const LinePlot = makePlotlyPlotComponent('LinePlot', (props: LinePlotProps) => {
...
-            // use <br> instead of \n for line break
-            return (
-              `x: ${binLabel ?? x}<br>y: ${yText}${CI}` +
-              (N !== undefined ? '<br>n: ' + N : '') +
-              (xtra ? '<br>' + xtra : '')
-            );
-          }),
-        })),
+
+        // need to exclude marginal histogram
```
bobular, line 300: a better comment would be

```
// don't process tooltips for marginal histogram
```

But I am confused - where is `series.mode` set to non-null?  Seems like the consumer `LineplotVisualization` where the value is `'lines+markers'`.  This is a bit opaque.  Is there any way we can make it more self-explanatory?

I think at the minimum some very clear comments should be added. It seems like a specification for the props passed to this component is that if `showMarginalHistogram` is true, then a data series containing the histogram counts with undefined `mode`, must be provided.  Perhaps even validate the props to make sure this is the case.

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2348,28 +2370,107 @@ function processInputData(
...
+                  el.overlayVariableDetails.value,
+                  overlayVariable
+                )
+              : el.seriesType !== 'zeroOverZero'
+              ? 'Data'
+              : '') +
+            (el.seriesType === 'zeroOverZero'
+              ? (overlayVariable !== undefined ? ', ' : '') +
+                'Undefined Y (denominator of 0)'
+              : ''),
```
bobular, line 2436: Can this section be DRYed up?

moontrip:
> I think the standard tooltip w duplicate x axis is fine, and that we should start without the gap between the two plots and see how much of an issue it really is in practice.
> 
> @bobular it's already a second ticket.

@d-callan Sounds good. I will make the change accordingly. 👍

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2348,28 +2370,107 @@ function processInputData(
...
+        el.binSampleSize &&
+        el.binSampleSize.length > 0
+      ) {
+        // compute binSampleSize one
+        const binSampleSize = categoricalMode
+          ? el.binSampleSize.map((val) =>
+              val == null
+                ? null
+                : (val as BinSampleSizeProportion).numeratorN /
+                  (val as BinSampleSizeProportion).denominatorN
```
bobular, line 2388: Just `denominatorN` here.  No fraction calculation.

bobular (changes requested):
All looking good, but I'd like to test again and agree on some styling issues before we merge.

bobular:
The slightly thicker top line (with no gap) bothers me in the storybook examples, but that doesn't seem to be a problem in the EDA. So I agree - no gap!

How about an axis label for the histogram?  "N"?  "Counts"?

d-callan:
Marginal plots often don't have an axis label, but I could go either way I guess..

moontrip:
> I have to say, it's a thing of beauty
> 
> image

@bobular Thank you for your feedbacks earlier. I will address them. By the way, do you prefer x-axis tick labels for marginal histogram or without them like your screenshot?

d-callan:
Yes. If we don't have the axes duplicated I think the dotted line make it much more clear which values in the line plot are associated w a bin in the histo

dmfalke (commented):
This is all looking great. Maybe next week we can discuss how to properly wire this up, using the `withOptions` functionality to reuse `LineplotVisualization` for the timeline plot.

moontrip:
@dmfalke I have defined `showMarginalHistogram` at visualizationPlugins (withOptions), https://github.com/VEuPathDB/web-monorepo/pull/290/commits/c3b51c90d04888dafe92c775d95f81891c115c6d. Technically, we don't necessarily do this for general lineplot Viz because I have set the default value as `false` at `LineplotVisualization` component if `options.showMarginalHistogram` is undefined, however, it may be better to explicitly define it for clarity.

I think that we may also use this new viz (line + marginal histogram) for standalone map's Viz. Although I did define `showMarginalHistogram` in there as well, it would require some additional work as I made a note in there (`standaloneVizPlugins.ts`).

packages/libs/eda/src/lib/map/analysis/hooks/standaloneVizPlugins.ts
```diff
@@ -40,6 +40,10 @@ export function useStandaloneVizPlugins({
       return visualization.withOptions({
         hideFacetInputs: true, // will also enable table-only mode for mosaic
         hideShowMissingnessToggle: true,
+        // TODO: need to distinguish lineplot from lineplot with marginal histogram?
+        // perhaps need to make another function only for lineplot with marginal histogram
+        // or define xxx.withOptions({}) directly at corresponding visualizationPlugins below
```
bobular, line 45: Yes I think we need to add `showMarginalHistogram` for just the appropriate lineplot(s).

bobular (changes requested):
Marginal histograms are currently off everywhere.

I think we'll need a bit of back end work to define a timeline plot, right? It will need special constraints on the x axis.

moontrip:
@bobular Thank you for your comments 👍 Yes, I disabled lineplot with marginal histogram for at pass app but perhaps we need to enable it for the time being (testing etc.). I will do that.

PR #294, opened 2023-06-09
Append statement about multi-valued variables not being supported.

### LINKED ISSUE #259: Map and general: almost-unconstrained variables do not "explain" to users the default constraint against multi-valued vars
At the moment the back end `apps` response contains the machine-readable constraints definition and a `description` prop that is presented to the user when they mouse-over a greyed out variable in the variable picker (`InputVariables`).

For example, the x-axis variable for a standalone-map scatterplot has this description:
> Variable must be a number or date and be the same or a child entity as the variable the map markers are painted with, if any.

Since (we will soon) closing https://github.com/VEuPathDB/EdaNewIssues/issues/635 we have an issue with the standalone map `map-markers` endpoint which has no `description` for the overlay variable - just the very basics:

```
{
	"overlayVariable": {
		"isRequired": true,
		"minNumVars": 1,
		"maxNumVars": 1
	}
}
```

The front end falls back to a default message: `'This variable cannot be used with this plot and other variable selections.'`

However, we should probably say that they are greyed out because we don't currently handle multi-valued variables.

Should we add a description to the back end response, or should we change the default message to mention multi-valued variables?

Opinions from @d-callan and @dmfalke would be gratefully received.

dmfalke:
closes #259 

This PR adds verbiage about multi-valued variables not being supported:

image

image

image

### DIFF
--- a/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
+++ b/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
@@ -958,8 +958,9 @@ const FieldNode = ({
         isMultiPick
           ? ''
           : isDisabled
-          ? (customDisabledVariableMessage ??
-              'This variable cannot be used with this plot and other variable selections.') +
+          ? (customDisabledVariableMessage
+              ? customDisabledVariableMessage
+              : 'This variable cannot be used with this plot and other variable selections.') +
             ' Muti-valued variables are not currently supported.'
           : 'Select this variable.'
       }
--- a/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
+++ b/packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
@@ -958,8 +958,9 @@ const FieldNode = ({
         isMultiPick
           ? ''
           : isDisabled
-          ? customDisabledVariableMessage ??
-            'This variable cannot be used with this plot and other variable selections.'
+          ? (customDisabledVariableMessage ??
+              'This variable cannot be used with this plot and other variable selections.') +
+            ' Muti-valued variables are not currently supported.'
           : 'Select this variable.'
       }
       className={


packages/libs/eda/src/lib/core/components/variableTrees/VariableList.tsx
```diff
@@ -958,8 +958,9 @@ const FieldNode = ({
         isMultiPick
           ? ''
           : isDisabled
-          ? customDisabledVariableMessage ??
-            'This variable cannot be used with this plot and other variable selections.'
+          ? (customDisabledVariableMessage ??
```
adnauseum, line 961: **question**: I might be doing something wrong, but I think the string will interpolate the boolean from the lefthand side of the `??`

I.e., 

image

dmfalke: Why, hello, stranger!

The type of `customDisabledVariableMessage` is `customDisabledVariableMessage: string | undefined`, so it shouldn't ever be `true` or `false`.

I was hoping that typescript would not compile if we changed that to `customDisabledVariableMessage: string | boolean | undefined`, but [it doesn't appear to](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF0QV00gtgCICWaUARhCMAGqzEVUCyIaZA5iAFzyYzEp28AD7xySJFSgoR8bClAAzAdQDcAWABQW5OiwAPeAF54ACjC4MBEmUrU6-RiBZsoneAH4PW+L7---AHIAFQALUngAN3onRGkUJCxyBFxqeAB3YgxQ+GyIgAcIRPhpYHhE0JAYKJi7XhAqMAxiVDQAOkCASngAah8AgcD4JmxmgFpoiGw06Mc7NBK4eASsCxg4FAwIAE9ebHz8pBgMag7VIA).

Thus, I will change this to a ternary expression. Thanks for the review, @adnauseum!

dmfalke: It's worth mentioning that, even with the ternary expression, if `customDisabledVariableMessage` is `true`, then the resulting message will be

```
true Muti-valued variables are not currently supported.
```

adnauseum: Good call! 🫡 

And hello 🙂. Hope all has been well, especially with the summer kids!

bobular (approved):
LGTM!

Hi @adnauseum - good to hear from you!

PR #297, opened 2023-06-13
Add differential abundance app and volcano viz

### LINKED ISSUE #282: volcano - add visualization
See mockups [here](https://www.figma.com/file/0Z6cCnTNg1yAaFGEOXQZeg/Microbiome-apps?type=design&node-id=2058-16835&t=lvs24bYs9Dn7iqzn-0)

For now, no table view, just the volcano. 

Viz inputs:
- log2 fold change threshold value (says where to draw vertical dashed lines)
- p value threshold value (says where to draw horizontal dashed line)
- no strata vars
- x and y are provided by computation

Screen Shot 2023-06-05 at 6 50 45 AM

### LINKED ISSUE #283: differential abundance - add frontend plugin
Compute Inputs:
- collection 
- variable for the group comparison. Let's start with categorical only.
- two variable groups, created with selectors just like we have for line plot

See mockups here
https://www.figma.com/file/0Z6cCnTNg1yAaFGEOXQZeg/Microbiome-apps?type=design&node-id=2058-16835&t=lvs24bYs9Dn7iqzn-0

x and y axis are provided by computation

asizemore:
Resolves #283 and resolves #282 
Adding the differential abundance plugin + volcano plot visualization. 

Current best screenshot 
Screen Shot 2023-07-14 at 7 01 04 AM

To Do for this PR

- [x] add containerStyles to volcano
- [x] input threshold values control (p value only)
- [x] thumbnail not working? (look to the map! the mapveu map in components)
- [x] Add loading spinner
- [x] Pass groups (or at least something!) to the x axis annotations

Questions in this PR:
1. How do we want to handle the plotRef type? It seems specific to plotly. My lazy typing is working, but it's definitely not the best appraoch.
2. If i have two vizs of the same 

The following tasks I've put in new issues.

- [ ] Axes range controls (add to visualization, component is ready)
- [ ] Add opacity slider to controls
- [ ] Set threshold by pvalue or adjusted p value #359 
- [ ] Continuous variables https://github.com/VEuPathDB/web-monorepo/issues/360
- [ ] Check to ensure no values overlap in groupA and groupB
- [ ] Add swap button to group A group B
- [ ] Add empty vizx plot. Otherwise the panel looks really weird before the viz data is ready. 
- [ ] Add legend
- [ ] Add axis truncation indicators (remember use props+data to derive this instead of state!)
- [ ] update numberInputRange to take increment prop (right now changes by 1, but we need it to change by 0.001 or so)
- [ ] Layout inputs and such nicer (may require update to LabelledGroup)

Any tasks above that do not have an associated issue next to them are included in this parent issue: #358 

To run this branch, use the qa backend. Also, _only_ use categorical variables please :) We're planning to add continuous vars in the future but i'm not sure yet if they're required for initial release.


asizemore:
Alrighty @jernestmyers and @dmfalke ready for your review!
It's not perfect but i had to stop somewhere. It's gotten quite out of hand! 

Looking forward to your feedback!

jernestmyers:
Just a heads up: I ran into the same compiler issues that the failed workflow encountered.
```
Error: src/lib/core/components/computations/plugins/differentialabundance.tsx(7,42): error TS2307: Cannot find module '../../visualizations/implementations/VolcanoPlotVisualization' or its corresponding type declarations.
Error: src/lib/core/components/computations/plugins/differentialabundance.tsx(13,26): error TS2307: Cannot find module '@veupathdb/coreui/dist/components/inputs/SingleSelect' or its corresponding type declarations.
```

But it's easy enough to fix, as follows:
``js
// line 7 should be cased with lowercase "plot"
import { volcanoPlotVisualization } from '../../visualizations/implementations/VolcanoplotVisualization';

// line 12 should nav to coreui's lib directory
import SingleSelect from '@veupathdb/coreui/lib/components/inputs/SingleSelect';
```

File: packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx [hunk 1]
```diff
@@ -0,0 +1,273 @@
...
+    computation,
+    Computation
+  );
+  const { configuration } = computation.descriptor;
+  const collectionVariable =
+    'collectionVariable' in configuration
+      ? configuration.collectionVariable
+      : undefined;
+  const comparatorVariable =
+    'comparatorVariable' in configuration
```
jernestmyers, line 83: I noticed the description was always rendering "Not selected" and its due to this line. We should be checking for `'comparator' in configuration`. Then, we'll also need to get the variable's display name so we're not just displaying the `variableId` further down, which won't mean anything to our users! Could be done with the `useFindEntityAndVariable` hook, but there may be something else out there to do so.

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -0,0 +1,273 @@
...
+            toggleStarredVariable={toggleStarredVariable}
+            entityId={configuration?.comparator?.variable?.entityId}
+            variableId={configuration?.comparator?.variable?.variableId}
+            variableLinkConfig={{
+              type: 'button',
+              onClick: (variable) => {
+                changeConfigHandler('comparator', {
+                  variable: variable as VariableDescriptor,
+                  groupA: configuration?.comparator?.groupA ?? null,
+                  groupB: configuration?.comparator?.groupB ?? null,
```
jernestmyers, line 240: I think we need to reset the group values when the comparator var changes. In my screenshot, I had selected group values based on `City, village or region` and then switched to `Delivery mode` and the newly irrelevant group values persisted.

image

I also inadvertently made _**a lot**_ of computations when testing this work, so I think we may need to further investigate how/why/when we make a new computation. Frankly, I don't much recall how I derived that logic and would be happy to help investigate.

[Warning: dizzying screencast of all the computations!](https://github.com/VEuPathDB/web-monorepo/assets/69446567/0dc17925-87f5-4f67-b0cd-8d4bb33e6a61)

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -255,11 +287,12 @@ function VolcanoPlot(props: VolcanoPlotProps) {
...
         {/* The data itself */}
         {/* Wrapping in a group in order to change the opacity of points. The GlyphSeries is somehow
-            a bunch of glyphs which are <circles> so there should be a way to pass opacity
-            down to those elements, but I haven't found it yet */}
+          a bunch of glyphs which are <circles> so there should be a way to pass opacity
+          down to those elements, but I haven't found it yet */}
         <Group opacity={markerBodyOpacity ?? 1}>
           <GlyphSeries
             dataKey={'data'} // unique key
+            //@ts-ignore
```
jernestmyers, line 295: I think this can be removed.

jernestmyers (commented):
Looking really nice so far- very cool viz!

I didn't quite finish reviewing everything, but I wanted to leave some comments before the weekend since you get such early starts (before I'll even be awake Monday morning)! 😄

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
asizemore: > I think we need to reset the group values when the comparator var changes

Oh yes good catch! Done!

asizemore: Wow that's a lot of computations!! You've been busy!
I agree the how/when/why of making a new computation is worth a think. One part I'm torn over is if the variable values should contribute to making a new computation section or not. I'm planning to demo this at the mbio meeting this week and can ask dan what he thinks

dmfalke:
> Alrighty @jernestmyers and @dmfalke ready for your review! It's not perfect but i had to stop somewhere. It's gotten quite out of hand!
> 
> Looking forward to your feedback!

I will spend most of today reviewing this, fyi

asizemore:
thank you both! It's already been helpful to just have a fresh set of eyes on this work. Jeremy has already caught a bunch of bugs i missed!

dmfalke (changes requested):
Overall, this looks really great. I made a few change suggestions. I also noticed that the generated tooltip has a different range than the source plot. Not sure if this is an artifact of the screenshot tool, or visx. Note that this screenshot was taken in storybook... maybe it isn't an issue in the application, since the source plot dimensions are more constrained?

image

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
dmfalke: There is something very strange going on with this. Every time I open/close either the **Group A** or **Group B** selector, the number of visualizations doubles.

dmfalke: > There is something very strange going on with this. Every time I open/close either the **Group A** or **Group B** selector, the number of visualizations doubles.

But this only happens if no new values are selected... very strange!

dmfalke: Also, all of these vizs have the same id :face_with_head_bandage:

dmfalke: I think there is a bug in this part of the code: https://github.com/VEuPathDB/web-monorepo/blob/8ba293186841a84e2406064b8e4fb08e7729c11f/packages/libs/eda/src/lib/core/components/computations/Utils.ts#L125-L151.

I think we need to check, earlier in this function, if the new compute config is indeed different than the existing compute config, and bail if it isn't.

We are getting here because the variable drop down will call the onchange handler, even if the value is the same. This causes the code to enter a path where it thinks it's moving a viz from one computation to another, but that is not what is happening.

dmfalke: Try applying this diff:

```diff
diff --git a/packages/libs/eda/src/lib/core/components/computations/Utils.ts b/packages/libs/eda/src/lib/core/components/computations/Utils.ts
index 371b17be3d..9984f186ee 100644
--- a/packages/libs/eda/src/lib/core/components/computations/Utils.ts
+++ b/packages/libs/eda/src/lib/core/components/computations/Utils.ts
@@ -98,6 +98,9 @@ function handleConfigurationChanges<ConfigType>(
   url: string,
   history: RouterChildContext['router']['history']
 ) {
+
+  if (isEqual(computation.descriptor.configuration, updatedConfiguration)) return;
+  
   // when a config value changes:
   // 1. remove viz from current computation
   // 2. check if the newConfig exists
```

asizemore:
Yes, the funky screenshot is an artifact of doing it in storybook without the nice dimensions constraints (I learned about it [here](https://github.com/VEuPathDB/web-monorepo/blob/4ee9a3b74b1851f65ee80a3054454f56952f7b7b/packages/libs/components/src/stories/MapRefRelated.stories.tsx#L171), since i mostly followed the map work). It all works great in the application though!

 
Screen Shot 2023-07-19 at 5 56 37 AM

asizemore:
Screen Shot 2023-07-19 at 6 00 20 AM
Wait a second... i've got that ref in the wrong spot! I'm capturing the whole nice panel that the plot sits on!

Fixed in the latest commit

asizemore:
Thank you both for your feedback! All ready for another look when you have a moment!

packages/libs/eda/src/lib/core/api/DataClient/types.ts
```diff
@@ -15,6 +15,7 @@ import {
   keyof,
   boolean,
   literal,
+  undefined,
```
jernestmyers, line 18: nit: can be removed with the new `partial` implementation

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+export type DifferentialAbundanceConfig = t.TypeOf<
+  typeof DifferentialAbundanceConfig
+>;
+
+const Comparator = t.type({
+  variable: VariableDescriptor,
+  groupA: t.array(t.string),
+  groupB: t.array(t.string),
+});
```
jernestmyers, line 46: I noticed the "Generate" button is still enabled when Group A and Group B values aren't selected. I think what's happening is that they are being set to empty arrays when a comparator var is selected, so it passes the `isConfigurationValid` check that [is done in the viz container component](https://github.com/VEuPathDB/web-monorepo/blob/496354de2ea8e6c850978e1f528ca6409ef71257/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx#L548).

If we set `groupA` and `groupB` to null instead of an empty array, we can get the disabling as desired. But this results in a TS error and some other strange behavior: the first selection made in the picker gets removed when first rendered. Doesn't seem to happen thereafter. 😵

dmfalke (approved):
LGTM!

jernestmyers (commented):
Looking really nice! 

Most of my comments are nits, but I am observing a `ResizeObserver` error that I'm looking into. How to reproduce:

1. Generate a valid computation
2. Wait for viz to generate
3. Change any of the `Comparator` vars and you'll get the `ResizeObserver` error

I've yet to pinpoint the exact cause, but I'll update this comment if/when I do!

asizemore:
@jernestmyers about the resize observer issue - do you have an idea of what this issue means? From my googling i didn't find anything other than basically exactly what the error says (which makes no sense to me because I'm not using ResizeObserver that i know of 😄 ) and that the error is benign so the site won't break. There's a ResizeObserver in the DraggablePanel but nothing that i see that would affect this app 🤔

jernestmyers:
> @jernestmyers about the resize observer issue - do you have an idea of what this issue means? From my googling i didn't find anything other than basically exactly what the error says (which makes no sense to me because I'm not using ResizeObserver that i know of 😄 ) and that the error is benign so the site won't break. There's a ResizeObserver in the DraggablePanel but nothing that i see that would affect this app 🤔

So we've seen a different `ResizeObserver` error in genomics record pages and [we just ignored it](https://github.com/VEuPathDB/web-monorepo/blob/496354de2ea8e6c850978e1f528ca6409ef71257/packages/libs/wdk-client/src/StoreModules/UnhandledErrorStoreModule.ts#L41-L45). It's possible we may just want to do the same with this error. It's also possible this error is only [in dev servers as a result of `webpack-dev-server`](https://github.com/webpack/webpack-dev-server/issues/4771) and won't bubble up in production.

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -0,0 +1,276 @@
...
+  createDefaultConfiguration: () => undefined,
+  isConfigurationValid: DifferentialAbundanceConfig.is,
+  visualizationPlugins: {
+    volcanoplot: volcanoPlotVisualization, // Must match name in data service and in visualization.tsx
+  },
+};
+
+function DifferentialAbundanceConfigDescriptionComponent({
+  computation,
+  analysisState,
```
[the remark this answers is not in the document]
asizemore: Good idea! Resolved in latest commit

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
asizemore: aha yes! Okay i will work on this now! Thank you!

And if you find one more console log or unused import in this PR your Friday beer is on me!

asizemore: (dont look yet though i just found one)

asizemore: alright @jernestmyers all fixed! I ended up defining a custom `isConfigurationValid` function. That was a good clue!

jernestmyers: Ah yes, good idea! Looks great.

asizemore:
Cool. Good to know! I guess let's see if it happens when deployed, and then if it does we can make a new issue for it. 

Thanks again for your patience with the many rounds of feedback!

PR #298, opened 2023-06-13
Use checkmark instead of chip to indicate active mapType

jernestmyers:
Replaces the `Chip` component with `active` string content with a checkmark per concerns about A) the `Chip` component looking like a button and B) an unclear parent-child hierarchy. The checkmark gives us better indentation that also better conveys the hierarchy.

Before:
image

After:
image


bobular (approved):
Very nice.

PR #302, opened 2023-06-14
Remove use of eda_study_id and url-based cache for permissions

dmfalke:
### Description

This PR removes the use of the WDK dataset record attribute `eda_study_id` to determine which datasets have an associated EDA study.

Instead, the following logic is used:
1. Get all WDK datasets
2. Get all EDA studies
3. Get permissions object
4. Use permissions object to map EDA study IDs to WDK dataset IDs, and create a Set of WDK dataset IDs that are in EDA
5. Filter WDK datasets based on membership in the Set above.

### Concerns

In VectorBase, the "mega study" dataset is composed of many smaller datasets. These individual datasets are loaded into EDA, and therefore are present in the Set of dataset IDs in EDA. This causes all of them to appear in the menu of "Interactive maps". We need a way to flag these smaller datasets as "hidden", probably ideally as a part of the EDA study information.

The current plan is to add a new question to the WDK model that will return datasets for which we will include a map link in the menu.

Until that question is added, `AllDatasets` is being used.

### Testing

This branch can be tested on both https://dfalke-b.clinepidb.org and https://dfalke-b.vectorbase.org.


dmfalke:
I have _assigned_ this PR to @bobular and @jbrestel to raise awareness. No action is needed, but feel free to comment!

bobular:
I tried to understand this but came up with a runtime error upon opening any EDA or map "My Analyses" page. dfalke-b back end. 

```
ERROR
Context has not been initialized: unknown context.
useRefinedContext@https://localhost:3000/static/js/bundle.js:563030:13
useSubsettingClient@https://localhost:3000/static/js/bundle.js:35579:115
AllAnalyses@https://localhost:3000/static/js/bundle.js:24601:87
renderWithHooks@https://localhost:3000/static/js/bundle.js:445375:31
mountIndeterminateComponent@https://localhost:3000/static/js/bundle.js:447858:17
```

bobular:
What's the overall aim here?

To get rid of database tables? To get rid of eda_study_id from the wdk model (which I don't full understand)?

It seems that deep down, the permissions request is using the StudyIdDatasetId table:

https://github.com/VEuPathDB/service-dataset-access/blob/master/src/main/java/org/veupathdb/service/access/service/dataset/DatasetRepo.java#L40

Though I don't fully understand what's going on in the service either, though it all looks very elegant.

bobular:
@jbrestel suggests providing a new WDK query called something like `MapStudiesForToolbar` that will provide all you need.  Any problems with that?

dmfalke:
> @jbrestel suggests providing a new WDK query called something like `MapStudiesForToolbar` that will provide all you need. Any problems with that?

Came here to add a similar comment :slightly_smiling_face:. This is great.

aurreco-uga (approved):

PR #305, opened 2023-06-16
pin yarn version for volta and delete migrations

asizemore:
After the migration to nx@16, was receiving the following error
```
Volta error: No Yarn version found in this project.

Use `volta pin yarn` to select a version (see `volta help pin` for more info).
```

Addressed by adding a pinned version of yarn, though we're not sure why volta is complaining at all...


asizemore:
tested by running the eda ✅

dmfalke (approved):

PR #308, opened 2023-06-20
add axis ranges prop to volcano

### LINKED ISSUE #72: Volcano: let user override default axis range
The volcano currently has `independentAxisRange` and `dependentAxisRange` as props but they're not used yet. If set, let them override the default axes calculated by the plot.

asizemore:
Resolves #72 

Updates the volcano component so that the props `independentAxisRange` and `dependentAxisRange` update the plot axes appropriately. Check out the volcano story to play with it!


packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -150,8 +150,21 @@ function VolcanoPlot(props: VolcanoPlotProps) {
           so use caution when ordering the children (ex. draw axes before data).  */}
       <XYChart
         height={height ?? 300}
-        xScale={{ type: 'linear', domain: [xMin, xMax] }}
-        yScale={{ type: 'linear', domain: [yMin, yMax], zero: false }}
+        xScale={{
+          type: 'linear',
+          domain: independentAxisRange
```
jernestmyers, line 155: Will `xMin` and `xMax` ever be used if `independentAxisRange` is passed in? Similar question re: `yMin` and `yMax` wrt to `dependentAxisRange`.

I ask because I wonder if in the future we should refactor the `domain` logic such that we only calculate what is needed.

jernestmyers (approved):
Does the trick, so I'm going to approve.

I have a vague memory of you suggesting that the data consumed by these volcano plots is quite large, so I do think we should save ourselves the computational costs of determining `xMin`, `xMax`, `yMin`, and `yMax` if we in fact won't use them in the presence of either axis range!

packages/libs/components/src/plots/VolcanoPlot.tsx
asizemore: good question. It looks like no right now but actually yes! I'm planning to have the component decide if it should show truncation indicators, so it will need to know if the set axis range is inside or outside the data range. I'll add a comment!

asizemore: ooh okay no you're right we don't need to calculate xmin or xmax if we have the range passed in. We already do the data range part elsewhere. Good catch!

PR #309, opened 2023-06-21
Add group annotations volcano

### LINKED ISSUE #70: Volcano: add "up in {group a}" or similar to x axis
Following the volcano [mockups](https://www.figma.com/file/0Z6cCnTNg1yAaFGEOXQZeg/Microbiome-apps?node-id=2059-18150&t=ql9WLMfIvMzuUYBo-0), there should be an "up in {group a}" on the left of the x axis and "up in {group b}" on the right of the xaxis. 

I think using `Annotation` will be simplest. Could anchor it to invisible points at (minX, minY) or (maxX, minY)?

asizemore:
Resolves #70 

Adds a prop to volcano to allow passing in annotations for the x axis. These annotations should always live at the two edges of the x axis, and be approximately in line with the x axis label (log 2 fold change). 

Mockups (the annotation in question here would be those days of the week at the bottom):
Screen Shot 2023-06-21 at 7 15 03 AM

Storybook (up in group a/b):
Screen Shot 2023-06-21 at 7 15 39 AM

Notes:
- I tried following the [documentation](https://airbnb.io/visx/docs/annotation#Label) for annotations but never got only `Label` or `CircleSubject` even to work. My annotations only worked as expected when using `AnnotationLabel` or `AnnotationCircleSubject`. Just an fyi for anyone doing similar things.
- These annotations are placed based on a static dy pixel value that looked good. Ideally we'd get the position of the x axis label and then use that, or use the logic for placing the x axis label when placing annotations. My thought was that i have no idea how these annotations will or will not get overlapped, hidden, cropped, moved, etc so to implement the cheapest solution until that becomes a problem. The x axis label should always be the same (it's hard coded in the volcano component) so this basic approach should get us going at least for a while? Happy to be convinced otherwise though!
- The `AnnotationLabel` also has a title prop that renders in bold above the subtitle. Just a future reference if we want to change how these annotations look or maybe put "Up in" in bold and then the list of variable values as the subtitle... For a future ux discussion :)


docker/makeSupportedBrowsersScript.js
```diff
@@ -19,8 +19,8 @@ const isModern = ${r}.test(userAgent);
...
-  if (err) throw err;
-});
-
-else process.stdout.write(program);
\ No newline at end of file
+if (outputFileName)
+  fs.writeFile(outputFileName, program, 'utf8', (err) => {
+    if (err) throw err;
+  });
+else process.stdout.write(program);
```
jernestmyers, line 26: Huh, I'm curious how these formatting changes came about? Obviously trivial, just curious!

asizemore: no idea! lol i assumed i must have opened it by accident at some point. I thought about reverting but since it looked like only formatting changes i didn't worry about it haha

Issue #245, opened 2023-05-22
Map: ez time filter


d-callan:
This has been discussed in dataviz. requirements, partly subject to some change/ finalization, are these:
1. have a variable input
2. means to indicate for which values that variable does or doesnt have data visible on the map (using the width of the slider)
    - is this a separate component?
    - needs legend as well
3. has text inputs but also two thumbs on the slider, and either can define range start and ends
4. either thumb can move fully independently
5. arrows to shift the current bin end to the next bin start and vice versa while maintaining the bin width
6. if you click the blue box you can move the entire thing to a new part of the slider while maintaining your bin width

see tentative ex (with obvious room for ux improvements) here. 
![ez_time_filter](https://github.com/VEuPathDB/web-monorepo/assets/53306535/bf5821dd-05ed-48cb-8178-c4e0fc825aa5)

d-callan:
worth noting, we may choose to phase these features in. their exact priority order isnt determined yet.

d-callan:
we have updated mockups 


![Image](https://github.com/VEuPathDB/web-monorepo/assets/53306535/a54f3a32-5a8b-4f69-a661-1c26158d8713)

moontrip:
> whoever works on this, just give me a ping when you need the icon!

@asizemore Thank you for your kind consideration. I think that the icon is not that urgent as I will be able to use any other icon for the time being - either way, I didn't even start SAM part either :)

PR #320, opened 2023-06-25
Ez time filter component and story

moontrip:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/245. 

It is now wired up to the user analysis saved state and the slider is working smoothly except for a known Firefox bug (see #480). The ez time filter now applies to markers and floaters. But not downloads. I added an on/off toggle, feedback welcome on this.

Remaining subtasks - in no particular order, and many could be done in parallel

- [x] play with it and figure out any major UX issues
- [x] grey out most of component when `timeSliderActive` is `false` (e.g. toggled off)
- [x] take it out of the draggable and attach underneath the header and 
      - create a super slim minimised mode
      - expand/collapse button (or click on the minimised mode to expand)
      - user interactions should only occur in the maximised mode, but the following basic information should be communicated in minimised mode:
           1. the variable name (plain text, ellipses as required)
           2. the distribution (no x-axis tick labels) and blue selected region if active (grey if inactive?)
- [ ] test with proper date variables and remove the number/integer/ordinal variable support (so it will only work with actual date variables, not year numbers, ordinals or the like)
- [ ] confirm that "show counts for filtered/visible" in the donut/barplot marker categorical variable config tables should include the ez time filter 
- [ ] add step left/right buttons (or leave for a future ticket)
- [ ] add date inputs to the date range display at top of widget (or make new ticket)
- [ ] sort out z-axis issues w.r.t. mouse-over marker enlargements (see screenshot)
image

<details>
  <summary>DK's original notes mainly about the slider component (click to expand)</summary>

It is primarily based on Visx Brush component, thus I added @visx/brush at web-components's package.json/dependencies. Also made a story file to test. 

Some notes are: a) the plot area is made of a kind of density plot for now. It is because it looks good and the mockup one will probably need much more work; b) an example data was taken from Lineplot data (GEMS1 Case Control; x: Enrollment date; y: Weight) for demo purpose; c) the selected range (or filtered area) is expressed as diagonal lines so that background plot (density plot) can be seen; d) initial position of the filter/selection is set to be whole range; e) the brush handle, which is for mouse drag to slide, is currently shown as small rectangles. It is a typical shape that can be found in other tools so I leave it as it: it seems like one can customize it though.

Here is a demo video:

https://github.com/VEuPathDB/web-monorepo/assets/12802305/5d122de3-3be1-4756-aa63-5bc2d2ec00c0

</details>


moontrip:
Update 1: made a time filter at SAM https://github.com/VEuPathDB/web-monorepo/pull/320/commits/8440f05c1defbd13751de1ce387b7c16336b215d . A demo video is attached in the following. Though, there are several things to do for working, including unknowns.

a) Currently, a temporary test data is used to show a kind of density plot. Not quite sure what data should be used for this purpose.
b) Although InputVariables component is attached, it is not functional. Similar to a), not quite sure what should be used for that variable tree.
c) since the range of the time filter is changed by dragging each end, I think that a Submit button needs to be made to add the selected range to the filter.
d) the icon and the position of the close/open button are temporary, which can be adjusted later. 

https://github.com/VEuPathDB/web-monorepo/assets/12802305/e5409f69-50cf-4d35-aad7-6c3751869048

d-callan:
a) there is [this](https://github.com/VEuPathDB/EdaDataService/issues/281) backend ticket, which more or less aims to establish a new endpoint dedicated to providing the data needed for this. the ticket describes providing a response indicating only the presence or absence of data, rather than how much data
b) any variable annotated as temporal should be available in that variable picker.. i suppose we could treat the new endpoint (from the above link) like a more typical viz and add a constraints pattern to it

moontrip:
> a) there is [this](https://github.com/VEuPathDB/EdaDataService/issues/281) backend ticket, which more or less aims to establish a new endpoint dedicated to providing the data needed for this. the ticket describes providing a response indicating only the presence or absence of data, rather than how much data b) any variable annotated as temporal should be available in that variable picker.. i suppose we could treat the new endpoint (from the above link) like a more typical viz and add a constraints pattern to it c) i dont love a submit button, can we debounce or something? (not sure i fully understand the implications of what im suggesting). also, ideally the text indicating the selected range above the slider could also serve as an input.
> 
> how much more work do we think it is to make the bar from the mockups?

Thank you for your detailed information, @d-callan 👍 Honestly, it is not clear how to wire data-related one to the time filter component until I see the response from the backend, so I will wait for the backend to be ready. 

As for the submit button, perhaps a kind of debounce may work without submit button as you mentioned. This may not be that clear to a user whether data request is made or not, so perhaps a loading  icon or something like that needs to be displayed not to change the range. Will think about it. In addition, you also mentioned about selected range as an input, which I guess you meant a kind of input form to change values. As far as I know, the sliding bar, which is made of Visx's brush component, is uncontrolled component. Thus it is very hard to manage it from parent component like map. Not quite sure if the following would work without testing, but this may be feasible by moving the range part (and reset icon) into the time filter component (EzTimeFilter.tsx). 

For your last comment, "make the bar", I guess you mean thin and thick horizontal bar style at mockup instead of current density plot. I thought that relevant data/response from the backend would include the format of x=time & y=value format, so in this case, it was arguably the easiest way to use Visx's xychart package and AreaClosed (like density plot) component so that x-axis tick & its label and y-axis values are shown with minimum efforts. From my understanding, there is no specific Visx component to display thin and thick horizontal bar style. This means that if we want to have such a bar style, then: 1) they should be made via SVGs manually; 2) this will also mean that x-axis should also be made via SVGs manually. That is, this would involve significant works to be implemented: basically this means that we need to make a specific plot component via d3 only.

asizemore:
On the subject of the bars, what if we just used histo data with all bars having height 1 if there is data and 0 if no data?
So if we get the binned data, for each bin we translate the counts to 0 or 1. Count becomes 1 if the original counts were 1 or higher, and stays at 0 if the original counts for that bin was 0. Then, we do a `BarSeries` instead of the `AreaClosed`?

That could mess with the axis a bit, but maybe worth a shot @moontrip ? 

The other idea i came up with was doing basically ticks at every point where we have data. So we'd use `LineSeries` and use the x value of the data for the x1, x2 of each line, and then like 5 and 0 for the y1 and y2. 

Anyways, curious if either of the above might be an easier way to get closer to the mockups!

d-callan:
thanks @asizemore. i do have a fairly strong preference to show presence/ absence of data as opposed to 'how much' data. if we can find a way to achieve that without a huge cost, that would be ideal.

moontrip:
Thank you for sharing your thoughts @asizemore 👍 Yes your suggestions can be relatively easier approach than mockup one. Looks like bar approach is more similar to the mockup. Will test it.

moontrip:
Hi @d-callan & @asizemore 
I have tested scaled values that represent with (1) or without data (0). In Visx, in general scaleBand is used for Bar plot's x-axis scaling instead of scaleTime (for, e.g., lineplot/time series, etc.). The scaleBand automatically compute bandWidth based on data and plot size, however, it cannot be used for our purpose directly because slider filter does not recognize the x-axis values which is time-based, not bin/categorical. But I found that scaleTime can also be used for Bar plot by computing bandWidth manually, which resolves the aforementioned issue. The first screenshot is a demo version. 

I have also tested lineplot (xychart) case with 0 and 1 values (2nd screenshot), but I guess you prefer the bar instead of having slopes between data and no data in the lineplot.

As a side note, I tested debouncing functionality to possibly be used for submitting selected range, and I could implement it with slider action.

ezTimeFilter-bar

ezTimeFilter-density

d-callan:
Thanks 👍 think you can make it shorter now? Possibly add a custom line annotation across the center of those bars?

moontrip:
Thank you for your comments @d-callan 👍 Can you please confirm my guess in the following? a) so making it shorter means the height of bars (i.e., plot height)? like 2/3 of current height? b) simply adding a horizontal center line in the bar plot?

d-callan:
my issue w the line is primarily that it leaves it a bit ambiguous whether there is or isnt data for July 2009

moontrip:
@d-callan, @asizemore, (and @bobular)

I have addressed feedbacks, center line in the horizontal direction and shorter height. In addition, I figured out how to make the time filter plot be closer to the original mockup, https://github.com/VEuPathDB/web-monorepo/issues/245. This is applied to both components' storybook and SAM, which a screenshot taken from SAM is attached FYI.

ez timefilter 03

moontrip:
> Looks very nice DK!

Thanks @bobular 😃

moontrip:
> beautiful! i wonder now if its possible to have the label above serve also as an input?

Glad you like it @d-callan :) As for the label/input style, although I will need to investigate more, at this stage there are a couple of issues that need to be addressed. The biggest issue is that the slider, which is made of Visx brush, is uncontrolled component within my knowledge. Thus, there seems to be no explicit way to change the slider position by parent component (e.g., via props): in other words, even if an input range is made and the range values are changed, the slider position won't be changed, which is not good for UI/UX perspective.

bobular (commented):
I have a suggestion how to make the `Brush` behave like it is controlled. It seems to work for me (I can control it from dev tools), but would like @dmfalke's view on this too!

First, set the initial position from EzTimeFilter's `selectedRange` prop

```
  const initialBrushPosition = useMemo(
    () => ({
      start: { x: xBrushScale(new Date(selectedRange.start)) },
      end: { x: xBrushScale(new Date(selectedRange.end)) },
    }),
    [selectedRange, xBrushScale]
  );
```

Then add a key to the `Brush` component

```
          <Brush
	    key={selectedRange.start + "/" + selectedRange.end}
```

bobular:
ChatGPT3.5 seemed to reluctantly agree that the `key` approach was OK.
Here are the docs for `Brush` - as DK says - it seems totally uncontrollable.
https://airbnb.io/visx/docs/brush

moontrip:
@bobular Thank you for your examination 👍 I saw something similar approach somewhere: IIRC, we also used similar one at our Plotly (for thumbnail I suppose). Will check.

packages/libs/components/src/stories/plotControls/EzTimeFilter.stories.tsx
```diff
@@ -0,0 +1,360 @@
...
+        </div>
+        {/* button to reset selectedRange */}
+        <div
+          style={{
+            marginLeft: 'auto',
+            marginTop: '-0.1em',
+            paddingRight: '1.5em',
+          }}
+        >
+          <a
```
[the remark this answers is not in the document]
moontrip: So, my (initial) intention to place the reset at parent level was for positioning. But I think that it may be better to place the reset and selected range label into the ez time filter component, and then adjust InputVariable component, which should be at parent component, via CSS positioning. 

Yes, I also noticed that <a> tag causes it at storybook. Will handle it, either change to button or something else.

bobular:
Hi @moontrip - did you run into problems making (faking) the controlled behaviour of the `<Brush>`?

Or have you not been able to get back to this due to time constraints?

moontrip:
> Hi @moontrip - did you run into problems making (faking) the controlled behaviour of the `<Brush>`?
> 
> Or have you not been able to get back to this due to time constraints?

Hi @bobular 👋 I think that it may work but as you said, I didn't have time to test it due to time constraints. Actually I did use key approach for other purpose, attach/detach functionality of the filter to the top area and it seems to work fine: didn't commit the latest works yet as it needs some refactoring as well as I am dealing with other ticket (truncation one you commented alot 😄).

dmfalke:
> Maybe @dmfalke could comment on my suggested use of the `key` prop?

This will work, but it could be expensive. Changing the value of the `key` prop causes the component to unmount and mount again. I wonder if we can base the key value off of some state that is incremented only when the input changes? Something like this, roughly speaking:

```typescript
function EZFilter(props: Props) {
  // Track how many times the input has changed by user 
  const [inputChangeCounter, setInputChangeCounter] = useState(0);
  // Generate brush key based on inputChangeCounter
  const brushKey = `iteration-${inputChangeCounter}`;

  // ...

  return (
    <>
      <DateInputs onChange={range => {
        // Update range value in analysis state
        updateRange(range);
        // Increment counter
        setInputChangeCounter(c => c + 1);
      }}/>
      <Brush key={brushKey}/>
    </>
}
```

moontrip:
Made lots of changes for ez time slider concerning UI and UX feedbacks. Summary is:

### changes
- changed the  box style of the selected range: no hatched lines, use the same color with HistogramFilter, etc.
- reduced some margins to be smaller overall
- used dash instead of tilde between dates
- removed no data line in the middle
- used darker gray (#333 used in floaters for no overlay)
- removed x (close) in the top right
- removed Reset button: removed forwardref
- changed the name from Time Filter to Time Slider
- static drawer mode (expand/shrink button): no interaction can be made under drawer mode (minimized); reset to initial value under drawer mode

### todo
- data request: distribution
- post processing response for the time slider data format
- applying selected range to filters
- adjusting input variable selector
 

A captured video is attached in the following:
https://github.com/VEuPathDB/web-monorepo/assets/12802305/4e018aac-7c82-4060-9834-abcaac3275b6

moontrip:
@bobular I think that major functionality of the ez time slider now works, even if you want to change some behaviors :) For your convenience, I have tried to add in-line comments as detailed as possible. Some notes are also provided in the following. In addition, For your better understanding of overall behavior of the current ez time slider, I made a video capture in the end of this comment. Thanks again for your kind consideration and willingness to help this work be finished. :thumb

It is worth noting that it is hard for me to find appropriate study to have time variable, thus for test purpose, I chose SCORE S. mansoni Cluster Randomized Trial study for the implementation of the EZ time slider. But technically there is no date variable at the study, I used loose conditions that teporarily allow integer variables as well: for those parts, I made comments, "DKDK TODO", with actual codes (commented out for now) that should be used for Production level. Also, note that due to such loose conditions, some plot's axis label and selected range label may have strange values because the time slider is supposed to accept date only, not number variable. But from the functionality viewpoint, this was good enough to test overall behaviors.

# Update
- data request to distribution
  - post processing the backend response for the time slider data format (0 and 1 in y-data and Visx format)
- adjusting input variable selector
  - set a condition to filter out variables at the variable tree: loose conditions for now as mentioned above
  - set a default variable so that in the beginning of the map instance, the time slider has a plot internally
- applying selected range to filters
  - after testing this, I decided to add submit button. Without that, it may lead to lots of data request when dragging range. Also, another reason is that without submit function, only one slider out of two (left and right) can be utilized: moving one slider results in new data request. The button is located at top right.
  - tried to change filters object when submitting the selected range of the ez time slider. By doing this, an advantage is that the selected range is shown as a filter chip at top left of the map. Since all data request is designed to respect filters, the plot of ez time slider will be back to its default plot when removing corresponding filter chip: so technically this is similar to a function of reset plot.
  - noticed that some variables are quite lengthy so variable tree (inputVariables) can take double lines. Thus, as compared to the previous commit, I slightly increase the height of the draggable panel so that Extend/Shrink button is not trimmed.
- hide ez time slider if no appropriate variable exists
- fixed a display issue when all y values are zeros


https://github.com/VEuPathDB/web-monorepo/assets/12802305/d7db094e-cda7-46db-8b34-70eaa58e509f

moontrip:
@dmfalke @bobular I leave some of my thoughts next to Dave's comment in the following (starting with DK:), hoping that these will help Bob continue this works on my behalf. Please be cautious that some of my comments may be incorrect thus deserve tests with implementations.

> Here are my thoughts on a minimal set of things to do for b65:


> * Minimized mode should be a small handle-looking thing, hanging off the header. It should display the selected time range and an expand icon.

DK: First of all, I am curious whether the selected time range you meant is pure text (like 2002-07-08 - 2022-12-20) or plot. Anyway, just for a clarification, the reason why the current display (like shown in the latest video capture) was made is that Steve wanted to have that style/format in our last mapveu ux meeting several weeks ago. 


> * Expanded mode should have a minimize/collapse icon


> * It should not make a "big" filter. Instead, it should be stored on the map appState in a dedicated property. The small filter should be appended to subset filters and passed to markers and plots.

DK: As far as I know, if we change/update filter(s) object, then filter chip catches the change and show it: this is why filter chip is shown when applying selected range at the time slider to filter(s). Note that this selected range is automatically reflected when dealing with Add filter, so I thought it was not that bad: in other words, if I set and applied Birth Year range from 2001-01-01 to 2010-01-01 via the time slider, and access to Birth year via Add filter button (top left), then the change was shown in there as well, which may make less confusion to user.


> * No submit button -- apply the small filter immediately

DK: If we should remove submit button, one way is to move updateFilter() at onSubmitTimeSlider() in the DraggableTimeFilter to onBrushChange() in the EzTimeFilter component: it may need some debounce() but I already implemented it though. A disadvantage without the submit button is that only one side of range can be used as a filtered range at once. Anyway, this work may be tricky and may cause side-effects.


> * Selection brush should always reflect state of filter (it currently resets, and the bars are filtered, which is not so good)

DK: I thought that the time slider plot (0 or 1 bar) should be changed when applying time slider's range/filter. But based on your  comments, you seem to prefer not to change the plot. IIRC, the data request to distribution backend (for time slider plot) currently has filters dependency, so this can probably be achieved by simply removing the dependency to avoid the plot change. But this also needs some tests.


> * Maybe have a "reset" or "select all" button in the ez filter

DK: it existed before but removed (using forwardRef). Previous commits may be helpful.


> * Only appears if the study has a time variable

DK: I think that this was already implemented.


> 
> Further UI improvements can be done for b66.

dmfalke:
Thanks, @moontrip. See my responses below.

> @dmfalke @bobular I leave some of my thoughts next to Dave's comment in the following (starting with DK:), hoping that these will help Bob continue this works on my behalf. Please be cautious that some of my comments may be incorrect thus deserve tests with implementations.
> 
> > Here are my thoughts on a minimal set of things to do for b65:
> 
> > * Minimized mode should be a small handle-looking thing, hanging off the header. It should display the selected time range and an expand icon.
> 
> DK: First of all, I am curious whether the selected time range you meant is pure text (like 2002-07-08 - 2022-12-20) or plot. Anyway, just for a clarification, the reason why the current display (like shown in the latest video capture) was made is that Steve wanted to have that style/format in our last mapveu ux meeting several weeks ago.

I talked with Steve, and he confirmed that we discussed what I described above. There may be a mock-up floating around. If I find it, I will link to it in another comment. I think the selected time range will be displayed as text, for example **Collection date: 2012-01-01 - 2018-12-31**.

> 
> > * Expanded mode should have a minimize/collapse icon
> 
> > * It should not make a "big" filter. Instead, it should be stored on the map appState in a dedicated property. The small filter should be appended to subset filters and passed to markers and plots.
> 
> DK: As far as I know, if we change/update filter(s) object, then filter chip catches the change and show it: this is why filter chip is shown when applying selected range at the time slider to filter(s). Note that this selected range is automatically reflected when dealing with Add filter, so I thought it was not that bad: in other words, if I set and applied Birth Year range from 2001-01-01 to 2010-01-01 via the time slider, and access to Birth year via Add filter button (top left), then the change was shown in there as well, which may make less confusion to user.

One of the requirements for this feature is that it does not impact subsetting. That could change in the future, but for now, we will just make a fake filter and concatenate it with subset filters, and then pass that to markers and plots.

> 
> > * No submit button -- apply the small filter immediately
> 
> DK: If we should remove submit button, one way is to move updateFilter() at onSubmitTimeSlider() in the DraggableTimeFilter to onBrushChange() in the EzTimeFilter component: it may need some debounce() but I already implemented it though. A disadvantage without the submit button is that only one side of range can be used as a filtered range at once. Anyway, this work may be tricky and may cause side-effects.

Thanks, this is useful!

> 
> > * Selection brush should always reflect state of filter (it currently resets, and the bars are filtered, which is not so good)
> 
> DK: I thought that the time slider plot (0 or 1 bar) should be changed when applying time slider's range/filter. But based on your comments, you seem to prefer not to change the plot. IIRC, the data request to distribution backend (for time slider plot) currently has filters dependency, so this can probably be achieved by simply removing the dependency to avoid the plot change. But this also needs some tests.

In the rest of the EDA application, it is standard that a filter UI is not sensitive to the filter it creates, other than to show the selection.

> 
> > * Maybe have a "reset" or "select all" button in the ez filter
> 
> DK: it existed before but removed (using forwardRef). Previous commits may be helpful.

Great, thanks!

> 
> > * Only appears if the study has a time variable
> 
> DK: I think that this was already implemented.

Great! I think we just need to modify the constraints for only temporal variables. We'll need to confirm the final criteria, maybe with @d-callan.

> 
> > Further UI improvements can be done for b66.

packages/libs/eda/src/lib/core/hooks/workspace.ts
dmfalke, line 1: I'd like to see the additions in the file moved to a file specific to the ez time slider

packages/libs/eda/src/lib/map/analysis/DraggableTimeFilter.tsx
dmfalke, line 1: I'm wondering if the business logic in the component should be factored out into a new file, `packages/libs/eda/src/lib/core/components/filter/EZTimeFilter.tsx`. I would remove the DraggablePanel part of it.

packages/libs/eda/src/lib/map/analysis/appState.ts
```diff
@@ -98,6 +99,12 @@ export const AppState = t.intersection([
       variableId: t.string,
     }),
     isSubsetPanelOpen: t.boolean,
+    timeSliderVariable: VariableDescriptor,
+    timeSliderSelectedRange: t.type({
+      start: t.string,
+      end: t.string,
+    }),
+    timeSliderActive: t.boolean,
```
dmfalke, line 107: I'd like to see this all combined into one object, something like:
```
timeSliderConfig: {
  variable: VariableDescriptor,
  selectedRange: t.type({
    start: t.string,
    end: t.string,
  }),
  isActive: t.boolean,
}
```

packages/libs/eda/src/lib/map/analysis/appState.ts
```diff
@@ -224,5 +249,8 @@ export function useAppState(uiStateKey: string, analysisState: AnalysisState) {
     setIsSubsetPanelOpen: useSetter('isSubsetPanelOpen'),
     setSubsetVariableAndEntity: useSetter('subsetVariableAndEntity'),
     setViewport: useSetter('viewport'),
+    setTimeSliderVariable: useSetter('timeSliderVariable'),
+    setTimeSliderSelectedRange: useSetter('timeSliderSelectedRange'),
+    setTimeSliderActive: useSetter('timeSliderActive'),
```
dmfalke, line 254: Similar to my other comment, this could be
```
setTimeSliderConfig: useSetter('timeSliderconfig')
```

dmfalke (commented):
Overall, I think this code is looking nice and healthy. I made a few suggestions, which I think will make things a little bit easier to find and work with. Let me know what you think, @bobular

dmfalke (commented):
I just made an observation: the date filters that get generated by the ez time filter are missing the time component, which causes a backend error, so that needs to be addressed somehow.

A related thought I just had is, perhaps appState can store a filter object? E.g.,

```
ezTimeFilterConfig: {
  isActive: boolean,
  filter?: Filter
}
```

packages/libs/eda/src/lib/core/hooks/workspace.ts
```diff
@@ -262,3 +269,62 @@ export function useGetDefaultVariableDescriptor() {
...
+          (entity) =>
+            entity.variables.filter((variable) => Variable.is(variable))
+              .length > 0
+        )[0];
+
+      // then take the first variable from it
+      const defaultTimeSliderVariable: Variable | undefined =
+        defaultTimeSliderEntity.variables.filter(
+          (variable): variable is Variable => Variable.is(variable)
+        )[0];
```
dmfalke, line 318: I ran into a runtime bug with this code, when I was playing with constraints. For some reason, typescript is allowing you to assume that `defaultTimeSliderEntity` is defined on line 316, even thought it might be null. The suggested change below will address this, and I think it's also a little more performant (`filter` will iterate the entire array, while `find` and `some` will terminate after the first element for which the callback returns `true`).
```suggestion
      // take the first suitable variable from the filtered variable tree

      // first find the first entity with some variables that passed the filter
      const defaultTimeSliderEntity = Array.from(
        preorder(temporalVariableTree, (node) => node.children ?? [])
      ).find((entity) => entity.variables.some(Variable.is));

      // then take the first variable from it
      const defaultTimeSliderVariable = defaultTimeSliderEntity?.variables.find(
        Variable.is
      );

      return defaultTimeSliderEntity != null &&
        defaultTimeSliderVariable != null
        ? {
            entityId: defaultTimeSliderEntity.id,
            variableId: defaultTimeSliderVariable.id,
          }
        : undefined;
```

dmfalke (changes requested):
I ran into a bug that will impact studies that don't have a temporal date variable.

packages/libs/eda/src/lib/core/hooks/workspace.ts
dmfalke: Furthermore, when this function returns `undefined`, the `MapAnalysisImpl` component enters and infinite loop.

packages/libs/eda/src/lib/core/hooks/workspace.ts
bobular: Good idea. Sorted in https://github.com/VEuPathDB/web-monorepo/pull/320/commits/8b4c75fa0092f8617863c11386e7c4e26433a6b9

packages/libs/eda/src/lib/map/analysis/DraggableTimeFilter.tsx
bobular: I've already renamed it EZTimeFilter and removed the draggable. I haven't done any further refactoring as yet.

bobular:
Hi @dmfalke - thanks again for your pre-review.

I've now [refactored](https://github.com/VEuPathDB/web-monorepo/pull/320/commits/8b4b84e798d72688bf4ff7a8a1cfd350881d5c89) the `timeSlider*` props into `timeSliderConfig` and simplified quite a lot.

It's working the same as before, even though I haven't wrapped the `setSelectedRange` handler passed to `EZTimeFilter` in a `useCallback`. Was that OK?

dmfalke (approved):
I think this is looking great. We should merge this asap so folks can see it.

PR #325, opened 2023-06-27
Feature 69 volcano truncation indicators

### LINKED ISSUE #69: Volcano: add truncation warnings
If a user sets the range so small that data points will be left out, we should show truncation indicators on the plot.

The question is... should we implement the same yellow boxes we don't like for consistency? Or move away from the boxes and show a pattern instead (easier with visx)? Has the time come to move to a new truncation warning indicator??

@d-callan @bobular

asizemore:
Resolves #69 

A bunch of thoughts:

1. One interesting question. So if we do patterned lines, by default there's no fill so we can see the points behind the pattern. I went to add a background color so that we'd hide those points, but then wondered could this be a potential feature not a bug? It clearly shows there are points out there but you, user, are hiding them. Something for us to think about...

Screen Shot 2023-07-18 at 10 57 19 AM

2. Is it better to move the TruncationBars component to its own file? Or wait until we have the second visx plot?
3. While implementing the truncation bars, i realized `clamp=true` in the XYChart axes was not doing what i thought. It was taking every point that fell outside the axis range and putting them at the extremes of the axis instead of chopping them off. Now the data accessors just ignore points that fall outside the axes. I considered having the visualization do this but it seems like the component should be sure it doesn't plot weird outlier points off the axes. 
4. As discussed in a data viz meeting, we have a prop for controlling if the bars should be the standard yellow or a fun pattern!
5. The indicators should automatically appear once the volcano viz accepts user-inputted axis ranges. (see #358 )


packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -166,12 +203,19 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
...
 
   // For the actual volcano plot data
+  // Only return data if the points fall within the specified range! Otherwise they'll show up on the plot.
   const dataAccessors = {
     xAccessor: (d: VolcanoPlotDataPoint) => {
-      return Number(d?.log2foldChange);
+      return Number(d?.log2foldChange) <= xAxisMax &&
+        Number(d?.log2foldChange) >= xAxisMin
+        ? Number(d?.log2foldChange)
+        : null;
```
dmfalke, line 212: Can you create a variable for the result of `Number(d?.log2foldChange)`, and use that in the comparisons/return?

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -166,12 +203,19 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
...
     },
     yAccessor: (d: VolcanoPlotDataPoint) => {
-      return -Math.log10(Number(d?.pValue));
+      return -Math.log10(Number(d?.pValue)) <= yAxisMax &&
+        -Math.log10(Number(d?.pValue)) >= yAxisMin
+        ? -Math.log10(Number(d?.pValue))
+        : null;
```
dmfalke, line 218: Can you create a variable for the result of `-Math.log10(Number(d?.pValue))`, and use that in the comparisons/return?

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -203,14 +258,19 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
         <XYChart
           xScale={{
             type: 'linear',
-            domain: [xAxisMin, xAxisMax],
-            clamp: true, // do not render points that fall outside of the scale domain (outside of the axis range)
+            domain: [
+              xAxisMin - +showXMinTruncationBar * xTruncationBarWidth,
+              xAxisMax + +showXMaxTruncationBar * xTruncationBarWidth,
```
dmfalke, line 263: What is the purpose of the `+` in `+showXMinTruncationBar` and `+showXMaxTruncationBar`?

dmfalke:
> Resolves #69
> 
> A bunch of thoughts:
> 
> 1. One interesting question. So if we do patterned lines, by default there's no fill so we can see the points behind the pattern. I went to add a background color so that we'd hide those points, but then wondered could this be a potential feature not a bug? It clearly shows there are points out there but you, user, are hiding them. Something for us to think about...
> 
> 
> Screen Shot 2023-07-18 at 10 57 19 AM

Seeing the points seems like a glitch. The purpose of the truncation bar is to indicate that there is data outside of the viewport. Are you able to remove points that are not between the truncation bars' inner edges?

> 2. Is it better to move the TruncationBars component to its own file? Or wait until we have the second visx plot?

I would wait. We might not need it, or we might have to approach it differently.


> 3. While implementing the truncation bars, i realized `clamp=true` in the XYChart axes was not doing what i thought. It was taking every point that fell outside the axis range and putting them at the extremes of the axis instead of chopping them off. Now the data accessors just ignore points that fall outside the axes. I considered having the visualization do this but it seems like the component should be sure it doesn't plot weird outlier points off the axes.

This makes sense to me. I'm wondering how this relates to my comment to item 1 above?

 
> 4. As discussed in a data viz meeting, we have a prop for controlling if the bars should be the standard yellow or a fun pattern!

:confetti_ball: 

 
> 5. The indicators should automatically appear once the volcano viz accepts user-inputted axis ranges. (see [Polishing differential abundance #358](https://github.com/VEuPathDB/web-monorepo/issues/358) )

:+1:

dmfalke (changes requested):
I have a few change requests. Otherwise, this is looking great!

packages/libs/components/src/plots/VolcanoPlot.tsx
asizemore: I was intending to succinctly extend the axis range only if the truncation bars were there, so i wanted the boolean `showXMinTruncationBar` to become 0 or 1 to help with that calculation. 

However, turns out that I can make `showXMinTruncationBar` and others numbers up earlier in the code and avoid all this `+` nonsense!

asizemore:
> The purpose of the truncation bar is to indicate that there is data outside of the viewport
Good point. Fixed! And you were right to question the connection of this with the filtering of data points through the accessor functions. Turns out that was doing all the work for us already and i just needed to add a fill to the pattern to put the last piece in place!

(There's another point that's being truncated but you can't see it)
Screen Shot 2023-07-21 at 1 55 52 PM

I also moved the comparison labels down based on the presence of y axis truncation bars. I considered moving them out horizontally, too, but i think they're just fine staying underneath the extremes of the data. 


Ready for another look!

packages/libs/components/src/plots/VolcanoPlot.tsx
dmfalke: I see! This is a little clever. Could you add a comment explaining that the value will be `0` or `1`, which will then determine if truncation width is added? Once that's done, I think it's good to merge!

dmfalke (approved):

PR #329, opened 2023-06-27
Stop map markers occasionally disappearing at minimum zoom

bobular:
Fixes https://github.com/VEuPathDB/web-components/issues/352 again!


bobular:
Hmm, not 100% sure about my logic here!

bobular:
It's very difficult to reproduce the bug.  But keep zooming out and it should happen (but not on this branch).

moontrip:
@bobular When I saw your first commit, I  wondered the same thing using the absolute value of the difference, which was applied in later commit 👍

moontrip (approved):

bobular:
Thanks DK!

PR #331, opened 2023-06-28
Fix problem with filtered date variables breaking visualization axis variables

### LINKED ISSUE #324: Filtered dates causing lineplot bug in visualisations in SAM and probably EDA
In feature.vectorbase.org (and current `main` using this back end) and **Maine Tick** data make a lineplot with collection date as x-axis, if you filter on collection date it explodes with a runtime error "invalid date".

I'm pretty sure EDA will also be affected.

I am on the case. It's presumably something to do with augmented study metadata related to filter-aware constraints and so on.

I'd even put money on it being related to trailing `"T00:00:00Z"`

bobular:
Fixes #324 

Turned out to be relatively simple.

In our "augmented study metadata" somehow some dates are appended `T00:00:00Z` and others `T00:00:00.000Z`. I'm not sure why. But I thought it was wise to handle both. I made the `Z` optional while I was at it.

### DIFF
--- a/packages/libs/eda/src/lib/core/hooks/workspace.ts
+++ b/packages/libs/eda/src/lib/core/hooks/workspace.ts
@@ -130,11 +130,11 @@ export function useStudyEntities(filters?: Filter[]) {
                             min:
                               filter.type === 'numberRange'
                                 ? filter.min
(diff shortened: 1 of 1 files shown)


moontrip (approved):
Looks good 👍

PR #332, opened 2023-06-28
Add SiteSearch Type-Ahead

Foxcapades:
Replaces the SiteSearch text input box with a type-ahead component that offers word completion suggestions based on the user's input.

### DIFF
--- a/packages/libs/web-common/src/components/SiteSearch/TypeAheadInput.tsx
+++ b/packages/libs/web-common/src/components/SiteSearch/TypeAheadInput.tsx
@@ -83,7 +83,7 @@ class TypeAheadAPI {
 
     if ((json as Array<any>).length == 0) return json as Array<string>;
 
-    for (const entry in json as Array<any>) {
+    for (const entry of json as Array<any>) {
       if (typeof entry != 'string') {
         throw new Error(
           'type-ahead API response contained a non-string value!'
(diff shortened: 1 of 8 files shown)


packages/libs/web-common/src/components/SiteSearch/TypeAheadInput.tsx
```diff
@@ -0,0 +1,504 @@
...
+// endregion Debouncer
+
+// region TypeAheadAPI
+
+const TYPEAHEAD_PATH: string = 'suggest';
+
+/**
+ * Wrapper for the SiteSearch Type-Ahead HTTP API.
+ */
+class TypeAheadAPI {
```
dmfalke, line 195: See https://github.com/VEuPathDB/web-monorepo/blob/main/packages/libs/http-utils/src/FetchClient.ts, for a base class you can use to implement this. Note that you can use it with https://github.com/VEuPathDB/web-monorepo/blob/main/packages/libs/http-utils/src/ioTransformer.ts to build run-time validation.

See an example of how it's used here: https://github.com/VEuPathDB/web-monorepo/blob/main/packages/libs/eda/src/lib/core/api/SubsettingClient/index.ts.

Also, note that API methods return a `Promise<T>`, rather than taking a callback, so you would also have to modify the caller of `typeahead`, if you go this route.

dmfalke (commented):
This is looking great. I am not opposed to everything residing in this module, as-is. One exception is `buildDebounder`. I suppose it can go in `/packages/libs/web-common/src/util`; or even a new package, `/packages/libs/utils`.

dmfalke (approved):
Changes look good!

PR #335, opened 2023-06-28
Enable UI for in_basket view filter

### LINKED ISSUE #317: Add toggle for turning "basket view filter" on/off
Gene results will soon include a view filter to show only rows that are in a user's basket. We will need to add a toggle to both the step result table, and to the attribute tabular reporter form.

The default should be "off". When turned "on", it should only apply to the step, and should only persist for the current session (probably using sessionStorage).

Placement of the toggle for the step results is TBD. I suggest putting it in the header for the basket column, but that will take a little bit of work:

image

### Decisions from UX Jun 28, 2023

- the basket view filter (checkbox) will say "Show only Genes in my Basket" (with a green icon on the left)
- the "Send to.."  dropdown will offer:
  - Basket (add to 345 genes)
  - Basket (replace)
- we offer it only in gene results
- we add the checkbox in the download form (there is already one for the transcript filter)
- we do not persist the checkbox (in contrast, the transcript view filter is stored as a global user preference)

dmfalke:
closes #317 

This PR enables the UI for the `in_basket_filter` view filter.

**Screenshots**

image

image

### DIFF
--- a/packages/libs/wdk-client/src/Views/ResultTableSummaryView/ResultTable.tsx
+++ b/packages/libs/wdk-client/src/Views/ResultTableSummaryView/ResultTable.tsx
@@ -11,7 +11,6 @@ import { pure, wrappable } from '../../Utils/ComponentUtils';
 import { Mesa, MesaState } from '../../Components/Mesa';
 import Link from '../../Components/Link';
 import BasketCell from '../../Views/ResultTableSummaryView/BasketCell';
-import BasketHeading from '../../Views/ResultTableSummaryView/BasketHeading';
 import PrimaryKeyCell from '../../Views/ResultTableSummaryView/PrimaryKeyCell';
 import AttributeCell from '../../Views/ResultTableSummaryView/AttributeCell';
 import AttributeHeading from '../../Views/ResultTableSummaryView/AttributeHeading';
@@ -23,7 +22,6 @@ import {
   RequestUpdateBasket,
   RequestAddStepToBasket,
   ViewPageNumber,
-  UpdateInBasketFilter,
   RequestPageSizeUpdate,
   ShowHideAddColumnsDialog,
   OpenAttributeAnalysis,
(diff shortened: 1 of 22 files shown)


aurreco-uga (approved):
as usual cant provide any useful feedback untl i see it in action :)

packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/records/ResultExportSelector.tsx
```diff
@@ -131,40 +126,36 @@ function useStyles<T extends string, S, E>(
...
           ? {
               color: '#3e3e3e',
             }
-          : {}
-      )
+          : {}),
+      }),
+      menu: (baseStyles) => ({
+        ...baseStyles,
+        width: '20em',
```
jernestmyers, line 151: I spy the change!

We really need to figure out a bulk formatting PR to get rid of these annoying diffs!

jernestmyers (approved):
I did what I could to comb through the code and play around with it on a local dev site server. LGTM!

packages/libs/wdk-client/src/Views/ResultTableSummaryView/ResultTable.tsx
```diff
@@ -76,6 +79,8 @@ function ResultTable(props: Props) {
     selectedIds,
     userIsGuest,
     showLoginWarning,
+    updateInBasketFilter,
+    inBasketFilterEnabled,
```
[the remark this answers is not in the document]
dmfalke: Oh, good catch. I changed some of the implementation and forgot to finish removing things from this part of the code. I'll do so before committing. Thanks!

packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/records/ResultExportSelector.tsx
dmfalke: Yes, we should!

PR #340, opened 2023-07-01
Remove redux middleware that is no longer needed

dmfalke:
This PR remove redux middleware used for data restrictions. Need to make sure we won't need this in the future, before merging.

This was requested by JB (on slack) when qa.vectorbase , eda-enabled, had eda-rbld missing some tables needed for the permissions endpoint  and searches were breaking.

### DIFF
--- a/packages/libs/eda/src/index.tsx
+++ b/packages/libs/eda/src/index.tsx
@@ -39,7 +39,6 @@ import { wrapWdkDependencies } from '@veupathdb/study-data-access/lib/shared/wra
 import {
   disableRestriction,
   enableRestriction,
-  reduxMiddleware,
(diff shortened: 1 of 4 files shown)


aurreco-uga (approved):
i understand this relates to checking on permissions on a few wdk user actions such as : run a search, access a record page, page a result, and download.. all within the wdk client. 
i dont know if this will ever become needed.. 
i think some kind of dataset access restrictions might be needed in the future but i would guess in a new form. 
so i would go ahead and remove the feature since it is not needed and we dont have a plan for it coming back ,

dmfalke:
During scrum, we agreed that this code can be removed.

PR #347, opened 2023-07-06
Combine map mouseModes and remove toggle

### LINKED ISSUE #196: Remove cursor modes
Discussed in [styling](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/270663681/Styling+Group+Meeting+May+10+2023) 05/10/23

Instead of having two modes, let's just have one cursor mode that combines them both. So upon mouseover, we get _both_ the large version of the marker and the gray data box. We could anchor the tooltip to the gray data box so that it's not hidden.

### LINKED ISSUE #230: Map - consolidate mouse modes into one default mode
- [x] tether magnified "popup" versions of the marker to the appropriate edge of the BoundsDriftMarker's bounding box (`bounds` and/or `boundingBox`) - so if the popup is popping down, it needs to be attached/tethered to the bottom edge of the grey bounding box rectangle. If it pops left, on the left edge, and so so on.
- [x] turn on magnify mode permanently in SAM and EDA MapViz
- [x] remove the mousemode component from SAM and EDA

jernestmyers:
Resolves #196
Resolves #230 

Some notes to reviewers:

1. Regarding this request from issue 230:
>  tether magnified "popup" versions of the marker to the appropriate edge of the BoundsDriftMarker's bounding box (bounds and/or boundingBox) - so if the popup is popping down, it needs to be attached/tethered to the bottom edge of the grey bounding box rectangle. If it pops left, on the left edge, and so so on.

I actually opted to tether the popup to whichever boundary extends furthest in the orientation desired. In other words, if the popup is popping down and the grey box's bottom edge _**does not**_ extend below the marker, we tether to the marker's bottom edge. (screenshots below)

2. The real work for this PR is in `BoundsDriftMarker.tsx`. I removed the mouse tools components in `MapVEuMap.tsx` and reduced opacity of the hovered marker in `map-styles.css`, but otherwise the diffs are simply removing the types, state, and props associated with the old, one-at-a-time mouse mode logic.

Marker is below the grey box, so tether popup to marker:
image

Grey box is below marker, so tether popup to grey box:
image


packages/libs/components/src/map/BoundsDriftMarker.tsx
```diff
@@ -30,12 +39,26 @@ export default function BoundsDriftMarker({
...
+  const boundsRectangle = L.rectangle(boundingBox, {
+    color: 'gray',
+    weight: 1,
+  });
+  useEffect(() => {
+    /**
+     * Prevents an edge case where the boundsRectangle persists if simultaneously a marker is hovered
+     * and a user changes the viewport
+     */
+    return () => {
```
bobular, line 57: While I normally like the `() => { }` syntax, it's nice for newbies if this is

`return function cleanup() { ... }`

bobular (approved):
This looks and works great. Nice job with all the complex positioning. The greyed out original marker works well.

One small issue that I found (thanks to one of your comments) was a "popupRef.current is null" runtime error when zooming in and out with popups active.

This seems easily solved by protecting any uses of ref values with `if (blahRef.current == null) return;` and similar. I'm going to trust you to do that and approve!

bobular:
Do we want another ticket and/or UX discussion to allow the user to keep the popup active my mousing over the grey box or the popup itself?  This may not be trivial - hence a new ticket suggestion.

jernestmyers:
> Do we want another ticket and/or UX discussion to allow the user to keep the popup active my mousing over the grey box or the popup itself? This may not be trivial - hence a new ticket suggestion.

A couple thoughts on this:
1. As far as I know, no additional information is gained by "interacting" with the popup/box. If I'm mistaken, then this proposal makes more sense to me.
2. A different take is that it's possible that a user _**may think**_ that the popup/box are interactive, in which case there's some value in letting the user dispel that notion by letting them hover over the popup/box.

bobular:
> > Do we want another ticket and/or UX discussion to allow the user to keep the popup active my mousing over the grey box or the popup itself? This may not be trivial - hence a new ticket suggestion.
> 
> A couple thoughts on this:
> 
>     1. As far as I know, no additional information is gained by "interacting" with the popup/box. If I'm mistaken, then this proposal makes more sense to me.
> 
>     2. A different take is that it's possible that a user _**may think**_ that the popup/box are interactive, in which case there's some value in letting the user dispel that notion by letting them hover over the popup/box.

Good points.

I guess I was thinking ahead to when the popups might be more interactive (potentially with shortcuts to open the marker config menu). It's probably safer to leave them as they are for now!

PR #352, opened 2023-07-10
plot resize with draggable panel

moontrip:
This is a test for resizing plotly's plot embedded in a draggable (and resizable) panel, like used in SAM. 
Tested several ways to make it happen but the only way I could do was to use draggable panel's onPanelResize() that returns current size of the panel. Then, a plot size in which a barplot is utilized for test purpose is determined by the current panel size. Please note that current format has an issue for the actual panel size, thus it is modified to avoid infinite loop of plot resize.

Here is a demo video:

https://github.com/VEuPathDB/web-monorepo/assets/12802305/3f0a971a-ae15-4371-87fe-fe2eba23a092

### DIFF
--- a/packages/libs/components/src/stories/plots/Barplot.stories.tsx
+++ b/packages/libs/components/src/stories/plots/Barplot.stories.tsx
@@ -6,6 +6,9 @@ import FacetedBarplot from '../../plots/facetedPlots/FacetedBarplot';
 import AxisRangeControl from '../../components/plotControls/AxisRangeControl';
 import { NumberRange, NumberOrDateRange } from '../../types/general';
 import { Toggle } from '@veupathdb/coreui';
+import DraggablePanel, {
+  HeightAndWidthInPixels,
+} from '@veupathdb/coreui/lib/components/containers/DraggablePanel';
 
 export default {
   title: 'Plots/Barplot',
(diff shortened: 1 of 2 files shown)


packages/libs/components/src/stories/plots/Barplot.stories.tsx
```diff
@@ -185,3 +188,47 @@ LogScale.args = {
...
+      defaultPosition={{ x: 200, y: 200 }}
+      confineToParentContainer
+      key={panelTitle}
+      isOpen
+      panelTitle={panelTitle}
+      showPanelTitle={true}
+      styleOverrides={{
+        resize: 'both',
+      }}
+      onPanelResize={(dimensions: HeightAndWidthInPixels) => {
```
bobular, line 214: I think we need to use `useCallback` on the function here and put the dependency back (see below)

packages/libs/coreui/src/components/containers/DraggablePanel/index.tsx
```diff
@@ -102,18 +102,31 @@ export default function DraggablePanel({
...
-        height: height,
+        // Note: since height from useResizeObserver includes dragHandle's height, 2rem,
+        // dragHandle's height should be substracted from height to compute main panel size
+        // without this, the use of useResizeObserver's height may lead to infinite loop
+        // when changing plot size inside the pannel
+        height: height - dragHandleHeight,
         width: width,
       });
     },
-    [height, width, onPanelResize]
```
bobular, line 116: I think this needs to be reinstated.

bobular (commented):
This looks great. Works very smoothly too.

We should discuss integration into SAM floaters next. Not trivial!

packages/libs/components/src/stories/plots/Barplot.stories.tsx
bobular: `useState` setters are guaranteed to be stable

bobular:
Do we need to add `mainPlotWidth` and `mainPlotHeight` props here?

```
export interface VisualizationProps<Options = undefined> {
  options?: Options;
  visualization: Visualization;
  dataElementConstraints?: Record<string, DataElementConstraint>[];
  dataElementDependencyOrder?: string[][];
  updateConfiguration: (configuration: unknown) => void;
  updateThumbnail?: (source: string) => void;
  computation: Computation;
  filters?: Filter[];
  starredVariables: VariableDescriptor[];
  toggleStarredVariable: (targetVariableId: VariableDescriptor) => void;
  totalCounts: PromiseHookState<EntityCounts>;
  filteredCounts: PromiseHookState<EntityCounts>;
  geoConfigs: GeoConfig[];
  otherVizOverviews: VisualizationOverview[];
  computeJobStatus?: JobStatus;
}
```

Do all the floaters have a "main plot"  - the contingency table might not. Can't test right now due to back end probems.

bobular:
I think maybe we only let the draggable size affect the width of the plot, and we set the height according to the aspect ratio of the default size?  We probably need to play with it a bit.

moontrip:
> I think maybe we only let the draggable size control the width of the plot, and we set the height according to the aspect ratio of the default size? We probably need to play with it a bit.

First of all thank you for your feedbacks above, @bobular 👍 As for this, actually that's what I wanted to ask/suggest when I alluded at Slack (😅) , because it is arguably the simplest solution without bothering the other components in a floating Viz such as input variable selector at top, plot controls and bottom, etc.

Issue #353, opened 2023-07-11
Map Time Series Visualization: Binning should be on by default


d-callan:
binning should be on by default, and possibly/ probably we shouldnt offer the ability to toggle it off. it doesnt really make sense to have an unbinned histogram.

PR #355, opened 2023-07-11
adjust behaviors of binning options for timeline Viz

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/353.
Screenshots are not attached as this is better to be tested directly. Added two reviewers but tests and approval from one of them is good enough, I suppose.


packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -1102,6 +1102,11 @@ function LineplotViz(props: VisualizationProps<Options>) {
     !showDependentAxisBanner &&
     data0?.binWidthSlider == null; // for ordinal string x-variables
 
+  // always enable useBinning for timeline Viz
+  const alwaysEnableUseBinning = useMemo(() => {
```
bobular, line 1106: Not sure a `useMemo` is needed here.  No heavy calculations!

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -1203,6 +1208,11 @@ function LineplotViz(props: VisualizationProps<Options>) {
     setTruncatedDependentAxisWarning,
   ]);
 
+  // change default vizConfig.useBinning to be true if timeline Viz
```
bobular, line 1211: I think there are two ways to avoid using the effect (which I presume could cause extra renders)

1. figure out how to pass the `options` to the `createDefaultConfig` function (I have no idea how!)
2. instead of updating the `vizConfig.useBinning` state, just use a regular "variable".

```
  const useBinning = alwaysEnableUseBinning || vizConfig.useBinning;
``` 

Then replace all `vizConfig.useBinning` with `useBinning`

bobular (changes requested):
I think it would be good to do it without the effect - see my suggestions

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
bobular: But maybe it doesn't work if we are passing `vizConfig` whole to other functions...

bobular: Like `getRequestParams` - bummer 😞

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: @bobular Thank you for your reviews 👍 I was debating whether it uses useMemo or not but can remove it :)

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: @bobular For your #1, the problem is that createDefaultConfig is defined out of Lineplot Viz component, thus options cannot be used as it is passed to Lineplot Viz component. As for #2, yes that was the reason why I didn't use a local variable like const useBinning, as it could have caused other side effects.

bobular: Can't we do something like this (completely untested):

```
  const customDefaultConfig = () => {
    const config = createDefaultConfig();
    if (options?.showMarginalHistogram) {
      return { ...config, useBinning: false };
    } else {
      return config;
    }
  };
  const [vizConfig, updateVizConfig] = useVizConfig(
    visualization.descriptor.configuration,
    LineplotConfig,
    customDefaultConfig,
    updateConfiguration
  );
```

You're correct that `createDefaultConfig` is exposed to the "outside" but I think it's only called in one place: (VisualizationsContainer.tsx:L377) and I don't think it's needed there (I will do a quick test).

moontrip: @bobular yes actually I have been doing similar thing to test as well 👍

bobular: Yes it works fine with `configuration: undefined` at L377.

The visualization component fills in the default config if the stored configuration isn't "correct".

bobular: Cool. But with `useBinning: true` of course!

bobular:
I made a comment on the commit but it hasn't shown up in the "conversation" here.

Here's a link to it https://github.com/VEuPathDB/web-monorepo/commit/b504619697896f447472de97641521b60559c88f#r122022133

bobular:
Answering [this](https://github.com/VEuPathDB/web-monorepo/commit/b504619697896f447472de97641521b60559c88f#r122094273) outside of the black hole... 

I still prefer this approach: https://github.com/VEuPathDB/web-monorepo/pull/355#discussion_r1265611857

moontrip:
> Answering [this](https://github.com/VEuPathDB/web-monorepo/commit/b504619697896f447472de97641521b60559c88f#r122094273) outside of the black hole...
> 
> I still prefer this approach: [#355 (comment)](https://github.com/VEuPathDB/web-monorepo/pull/355#discussion_r1265611857)

Thank you for your suggestion, @bobular 👍 So, you don't like using useEffect() 😄

moontrip:
@bobular I have added some conditions to avoid the issue. As I mentioned, your customDefaultConfig() is not a must item as this new commit (https://github.com/VEuPathDB/web-monorepo/pull/355/commits/d8284097ab55a8a7fe0da41c0d25f07580103dae) still needs. I am hesitant to introduce customDefaultConfig() as I found that the createDefaultConfig() is used in other place (StartPage) too and no specific need to change it to resolve the ticket.

It would be great if you can test this and let me know if there is a glitch.

bobular:
The `StartPage` usage was calling `createDefaultConfiguration` on a `ComputationPlugin` not a visualization (it's all very confusing) so I don't think that would have been a problem.

bobular (approved):

bobular:
Having said that I like the new logic!

Unfortunately I'm unable to test right now due to dfalke-b back end problems.

moontrip:
> Having said that I like the new logic!
> 
> Unfortunately I'm unable to test right now due to dfalke-b back end problems.

Thank you for your reviews, @bobular Yes I couldn't access to dfalke-b as you already saw my empty map at our meeting haha.

PR #356, opened 2023-07-12
proof of concept bipartite network

asizemore:
Work towards a correlation app for mbio. Gathering information.

Current screenshots:

small size
Screen Shot 2023-07-12 at 7 18 19 AM

lots and lots of points (scrolls vertically)
Screen Shot 2023-07-12 at 7 18 51 AM

Open Questions:
1. Let's say we let the user filter nodes and edges out of the network. Is that a part of the viz request? Or does the client do that? I could see either right now
2. I think this will be a separate component from the unipartite network (hairball), but i'd like the data structure to be the same. This seems acheivable.

Current idea of data coming in from the backend:
```
[{
nodeA: "node A name or id"
nodeB: "node B name or id"
edgeWeight?: number
edgeOpacity?: number
}]
```

I'm thinking about using visx `Graph`, but i haven't totally ruled out the `XYChart` approach for consistency with other vizs. 

More thoughts in the confluence doc [Networks](https://veupathdb.atlassian.net/wiki/spaces/FG/pages/198082575/Networks)

### DIFF
--- a/packages/libs/components/src/stories/plots/BipartiteNetwork.stories.tsx
+++ b/packages/libs/components/src/stories/plots/BipartiteNetwork.stories.tsx
@@ -1,0 +1,118 @@
+import VolcanoPlot, { VolcanoPlotProps } from '../../plots/VolcanoPlot';
+import { Story, Meta } from '@storybook/react/types-6-0';
+import { range } from 'lodash';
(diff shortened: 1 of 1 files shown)


asizemore:
This POC served its purpose. Looking forward to the real implementation soon!

PR #362, opened 2023-07-17
Fix bugs related to 'All other values' attribute

### LINKED ISSUE #361: Map: `CategoricalMarkerConfigurationTable` mishandling deselect/select all behavior
<h3>To reproduce</h3>

- Use the tick study
- Select organism identification datum as a variable (the variable must have exactly 7 values)
- Use the top-most checkbox to deselect all values in the table
- Then use the same top-most checkbox to select all values
- the checkboxes will be marked as checked but nothing happens on the screen 

<h3>Issue</h3>
The “deselect all” functionality sets the overlay config to be `UNSELECTED_TOKEN`, which aggregates to the `All other values` attribute

- this was done because setting overlay config to an empty array, null, or undefined results in undesirable behaviors (empty array) or errors (null/undefined)

The “deselect all” functionality results in `UNSELECTED_TOKEN` being included in overlay configs for vars that have less than the maximum allowable values, meaning these vars first render **_without_** the `All other values` attribute because `All other values` would be 0. Currently, the `All other values` attribute will remain as users select more values in the table. Again, this arguably makes sense because the `All other values` attribute has a value up until we’ve selected all available values (remember, this is possible because the number of available values is less than the max). Currently, this logic is where the bugs/misbehaviors sit.

<h3>Solutions</h3>

1. I didn’t think about how “deselect all” and then “select all” (or selecting one-by-one up to all values) would include `UNSELECTED_TOKEN` for vars with less than the max selections. My conditional only sets a new config if we’ve selected less than the maximum allowable number (based on the color palette, which includes a space for `UNSELECTED_TOKEN`). By checking if we are less than **_or equal to_** the max, we can fix this bug.
2. I didn’t account for this “rogue” `UNSELECTED_TOKEN` when thinking about the “single select” handler. We should add a check that removes the token once `All other values` would be 0, meaning if we’ve selected all available values and we’re below the max threshold, we need to remove `UNSELECTED_TOKEN`.

jernestmyers:
Resolves #361


bobular (approved):
Looks great! I did a fair amount of testing while doing the marker-center summation thing and then the response synchronisation thing.

PR #370, opened 2023-07-19
Suggestion to synchronise vocabulary and rawMarkersData dependencies

bobular:
Leveraging the usePromise to deliver the new vocabulary only when the data request has completed.

The main thing to know is that whenever the callback function inside the usePromise returns undefined, the payload in, for example `rawMarkersPayload.value` remains as the previous successful return value.

The implementation is ugly - the naming is nasty in particular (I didn't like `rawMarkersDataAndVocabulary` either!). It's hopefully possible to do something like this with fewer ugly changes (e.g. lots of `rawMarkersPayload.value?.rawMarkersData`)

A bonus advantage of this approach is that the markers and legend don't change until the data has come back from the back end.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
@@ -110,6 +110,7 @@ export function useStandaloneMapMarkers(
   // when switching between pie and bar markers when using the same variable
   const selectedOverlayVariable = useDeepValue(sov);
   const overlayConfig = useDeepValue(oc);
+  const overlayType = overlayConfig?.overlayType;
 
   const dataClient: DataClient = useDataClient();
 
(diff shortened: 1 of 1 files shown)


packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
```diff
@@ -154,15 +155,13 @@ export function useStandaloneMapMarkers(
...
-  const overlayType = overlayConfig?.overlayType;
-  const vocabulary =
-    overlayType === 'categorical' // switch statement style guide time!!
-      ? overlayConfig?.overlayValues
-      : overlayType === 'continuous'
-      ? overlayConfig?.overlayValues.map((ov) => ov.binLabel)
-      : undefined;
-
-  const rawMarkersData = usePromise<StandaloneMapMarkersResponse | undefined>(
+  const rawMarkersPayload = usePromise<
```
bobular, line 158: This is the main bit - we're returning the data and the vocabulary.

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
```diff
@@ -339,7 +347,9 @@ export function useStandaloneMapMarkers(
 
         const count =
           vocabulary != null // if there's an overlay (all expected use cases)
-            ? overlayValues.reduce((sum, { count }) => (sum = sum + count), 0)
+            ? overlayValues
+                .filter(({ binLabel }) => vocabulary.includes(binLabel))
```
bobular, line 351: Here's the filter I suggested in Slack that makes sure the counts are correct for low-cardinality vars.

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
```diff
@@ -154,15 +155,13 @@ export function useStandaloneMapMarkers(
     [boundsZoomLevel?.zoomLevel, geoConfig]
```
bobular, line 155: I also noticed that I had done a missing dependency... tut tut.

I think it would be fixed by putting `boundZoomLevel?.zoomLevel != null` in place of `boundsZoomLevel` at line 146/147 above.

jernestmyers (approved):
Unable to test right now with feature and qa down, but this looks good to me nonetheless. I also don't have any issue with the naming, but we may want to add some documentation that explains what and why we're lumping the vocab into the "payload" promise.

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
bobular: Fixed it.

PR #371, opened 2023-07-20
Omit required input logic and clear button for non-nullable inputs

### LINKED ISSUE #313: Remove the asterisk from the variable selector when the input is "non-nullable"
In the SAM marker configuration (for example, there are probably others) we have a variable selector that is mandatory/required and we provide a default variable for it _and we don't provide a (x) button to void the selection_.  In this case we don't need to render the asterisk to say it's mandatory.

jernestmyers:
Resolves #313

Note that I moved `showClearSelectionButton` from the `InputVariables` props to the `InputSpec` object with the idea that we may want to have more granularity in the future.


dmfalke (approved):
LGTM!

bobular:
Had a look, and it looks good!

PR #373, opened 2023-07-21
Open link attributes for external sites in a new tab

dmfalke:
closes r31965

This PR adds HTML attributes to open WDK linkAttributes for external links in a new tab.


aurreco-uga (approved):
would this work for jbrowse, apollo, mapveu  opening in a new tab?

PR #374, opened 2023-07-24
Relax contraints and set outputEntity appropriately to handle 1-to-1 entity relationships

### LINKED ISSUE #344: Visualizations: Relax constraints between entities when 1-to-1
This hasn't quite been signed off in a large UX meeting - so please hold back unless you are keen to get a draft version working.

This related to the "output entity" agenda item [here](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/311590913/2023-06-29+Map+UX).

The study metadata contains an entity property `isManyToOneWithParent` - if this is false, somehow we need to treat it as at the same level as the parent entity.

Probably needs to be implemented in or around `disabledVariablesForInput` in data-element-constraints.ts

The functions `ancestorEntitiesForEntityId` and `descendantEntitiesForEntityId` currently return same-or-ancestor and same-or-descendant entities.  They should probably be wrapped in a function that expands a list of entities to include any 1-to-1 entities related to the original list. Should be recursive somehow, so that if A-B-C-D are all 1-to-1, expanding a list of `[ B ]` would produce `[ A, B, C, D ]` (I don't think the order matters).

bobular:
Resolves #344 

Mainly tested on qa.vectorbase.org back end with the Maine tick dataset - this back end provides the appropriate relationship information. Check before testing that the Maine ED looks like this (one one-to-many and the rest one-to-one):

image

### Backwards compatibility

TL;DR: we're going to have to be careful about deployment of this vs. back end versions

Testing with qa.clinepidb.org, which provides a blanket `false` value for `hasManyToOneWithParent` - this means that the new client code will behave extremely permissively (it thinks everything is one-to-one) and this can allow illegal variable selections, e.g. below in SCORE Seasonal Transmission S. haematobium Cluster Randomized Trial

image


packages/libs/eda/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -408,7 +413,14 @@ function BarplotViz(props: VisualizationProps<Options>) {
...
           vizConfig: dataRequestConfig,
-        }) ?? getRequestParams(studyId, filters ?? [], dataRequestConfig);
+          outputEntityId: outputEntity.id,
+        }) ??
+        getRequestParams(
+          studyId,
+          filters ?? [],
+          dataRequestConfig,
+          outputEntity
+        );
```
bobular, line 423: Just one new arg and reformatting!

packages/libs/eda/src/lib/core/utils/data-element-constraints.ts
```diff
@@ -420,10 +430,16 @@ export function mergeMaxNumValues(
  */
 export function ancestorEntitiesForEntityId(
   entityId: string,
-  entities: StudyEntity[]
+  entities: StudyEntity[],
+  childFilter: (entity: StudyEntity) => boolean = () => true
 ): StudyEntity[] {
   const ancestors = entities.reduceRight((ancestors, entity) => {
```
dmfalke, line 436: It looks like this relies on `entities` being in a particular order. That makes me nervous about potential future changes. I wonder if it would be better for this function to take a root entity, and perform a tree traversal to find ancestor entities; but this might require upstream changes, so might be better to add a `TODO` comment for now?

dmfalke (commented):
Code looks good to me. I made a comment below.

packages/libs/eda/src/lib/core/utils/data-element-constraints.ts
bobular: Indeed - you've commented almost exactly this before. I think I added some comments that time.
Is it time to dive in and fix properly?

bobular:
> > Testing with qa.clinepidb.org, which provides a blanket `false` value for `hasManyToOneWithParent`
> 
> If it's easy to change to `true`, would that cause other issues? If not, maybe we can explore that?

Do you remember where it happens?  I'll ask on dd-ui-infra and CC Ryan.

bobular:
Liaised with Ryan who kindly switched all the default zeroes to 1s in eda-inc. (It turned out that they were not NULLs after all.)

This can now be merged.  If eda-inc still drives sites upon b65 release we will need to introduce an app configuration flag to enforce single connectors on all entity diagrams.

dmfalke (approved):

PR #375, opened 2023-07-24
Fix lost selection state in value picker

jernestmyers:
Fixes a bug with the `ValuePicker`s for Group A and Group B values.

<h3>The bug</h3>
Changing the comparator var resets the group A and group B values and the component will re-render a handful of times. As a result, the `selectedValues` will be assigned a new empty array literal during <strong><em>each render</em></strong> due to this line: <br /><br />
<code>selectedValues={configuration?.comparator?.groupA ?? []}</code>
<br /><br />
The `ValuePicker` component sees this as a new state due to object equality/reference reasons and thus it will re-render and reset its state back to an empty array.

<h3>Steps to reproduce the bug</h3>
<ol>
  <li>Select a comparator var</li>
  <li>Quickly-ish open a group selection list and make a selection</li>
  <li>Notice  selection state disappears</li>
</ol>

<h3>Solution</h3>
By removing the nullish coalescing operator (`??`), we can <a href="https://github.com/VEuPathDB/web-monorepo/blob/d0e102ebc29d465050177e0f76878d7b88572364/packages/libs/eda/src/lib/core/components/visualizations/implementations/ValuePicker.tsx#L16">let `ValuePicker` assign an undefined value the default `EMPTY_ARRAY` array reference</a>, which will be stable.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
@@ -258,7 +258,7 @@ export function DifferentialAbundanceConfiguration(
         <div style={{ justifySelf: 'end', fontWeight: 500 }}>Group A</div>
         <ValuePicker
           allowedValues={selectedComparatorVariable?.variable.vocabulary}
-          selectedValues={configuration?.comparator?.groupA ?? []}
+          selectedValues={configuration?.comparator?.groupA}
           onSelectedValuesChange={(newValues) =>
             changeConfigHandler('comparator', {
               variable: configuration?.comparator?.variable ?? undefined,
@@ -270,7 +270,7 @@ export function DifferentialAbundanceConfiguration(
         <div style={{ justifySelf: 'end', fontWeight: 500 }}>Group B</div>
         <ValuePicker
           allowedValues={selectedComparatorVariable?.variable.vocabulary}
-          selectedValues={configuration?.comparator?.groupB ?? []}
+          selectedValues={configuration?.comparator?.groupB}
           onSelectedValuesChange={(newValues) =>
             changeConfigHandler('comparator', {
               variable: configuration?.comparator?.variable ?? undefined,


bobular (approved):
Just want to say that logically this makes perfect sense. Array literals are often a problem like this. 


I haven't tested it though!

asizemore:
So ValuePicker sees `[]` as a new state because it is not the same as either some array with values or `EMPTY_ARRAY`, yes? Good find and thank you for the explanation!

bobular:
Each time an array is created "literally" in your code (e.g. `const foo = [ 1, 2, 3 ]`) it is created anew in a different place in memory and as such they are all "referentially distinct/unequal".

It's easier to demonstrate in Perl, because when you "print" a reference to something it prints its memory address:

```
print [];
# output: ARRAY(0x6484868)
print [];
# output: ARRAY(0x64844d8)
```

and so

```
print [] == [];
# output: false
```

but "saving" it in a variable (like `EMPTY_ARRAY` in our case)
```
$empty = [];
print $empty == $empty;
# output: true
```

bobular:
You're welcome.

Our solution was to use a "package global" (not sure if that's the right terminology for JavaScript) in ValuePicker.tsx.  It's not very pretty but it's readable and concise.

Another solution would be to memoize the array inside the functional component.  Below is not tested, but it would look something like this. Not as concise!

```
export function ValuePicker({
  allowedValues,
  selectedValues,
  onSelectedValuesChange,
}: ValuePickerProps) {
  const EMPTY_ARRAY = useMemo(() => [], []);

  # then use `allowedValues ?? EMPTY_ARRAY` where you would normally use `allowedValues`
```

or perhaps this is cleaner, but still not as clear as what we've done IMO

```
export function ValuePicker(props: ValuePickerProps) {
  const EMPTY_ARRAY = useMemo(() => [], []);
  const allowedValues = useMemo(() => props.allowedValues ?? EMPTY_ARRAY, [ props.allowedValues, EMPTY_ARRAY ]);
```
Maybe there's a more concise way, but you get my drift...

jernestmyers:
@dmfalke helped me debug this, so much of the credit goes to him!

Issue #369, opened 2023-07-18
Visualizations: truncation bars not showing


asizemore:
Not sure what i did wrong here, but here's the autozoom
<img width="793" alt="Screen Shot 2023-07-18 at 10 52 47 AM" src="https://github.com/VEuPathDB/EdaNewIssues/assets/11710234/bf9e17d3-d1c2-4e12-ba13-79a9eabdbcec">

And here's me trying to get a truncation bar to show up. Sometimes it'll appear for a moment and then it will disappear. There should be one on the left side of the plot
<img width="847" alt="Screen Shot 2023-07-18 at 10 51 55 AM" src="https://github.com/VEuPathDB/EdaNewIssues/assets/11710234/6284846e-87af-4968-a034-d556750e6648">

moontrip:
> Oh, this was a known limitation with a non-trivial, possibly back end IIRC, solution. @asizemore @moontrip

Thanks @bobular 👍 I missed your comment while commenting at PR 😅

bobular:
No @moontrip after posting the above I saw on your PR that you had looked into the history of this issue. My apologies! I'm reviewing now.

moontrip:
> Is this addressed by #376?

@dmfalke Sorry, I forgot to close this ticket. Yes, it was resolved by #376. Also, I think that the #376 also addressed long-lasting ticket, https://github.com/VEuPathDB/web-eda/issues/1295: can you please confirm this, @bobular?

bobular:
> > Is this addressed by #376?
> 
> @dmfalke Sorry, I forgot to close this ticket. Yes, it was resolved by #376. Also, I think that the #376 also addressed long-lasting ticket, [VEuPathDB/web-eda#1295](https://github.com/VEuPathDB/web-eda/issues/1295): can you please confirm this, @bobular?

Yes, please also close 1295. It also mentions lineplot, but lineplot does not have "semantic zooming" so I don't the same problem applies.

PR #376, opened 2023-07-24
independent axis truncation for Histogram Viz

moontrip:
This will probably address https://github.com/VEuPathDB/web-monorepo/issues/369. 

I tested all Viz with axis control and found that only independent axis of Histogram Viz has the issue, even at ClinEpiDB. If I remember correctly, this issue has not been addressed for a long time, e.g., https://github.com/VEuPathDB/web-eda/issues/1295 even if there have been some efforts to resolve it at other PRs, e.g., https://github.com/VEuPathDB/web-eda/pull/1464 & https://github.com/VEuPathDB/web-eda/pull/1623.

For Histogram Viz, independent axis control like custom range triggers data request via viewport change. Thus, default independent axis range for Histogram Viz should remain the same regardless of setting custom range to judge if it is truncated or not. To resolve this, an additional prop is passed to useDefaultAxisRange hook so that the default range can be kept.


moontrip:
@asizemore Somehow I cannot run microbiome's Visualization so I could not test your case, however, this would work. I am not quite sure if a configuration (.env.local) was changed in there though.

moontrip:
Here is a demo: GEMS1 with the main variable of Age. As you can see, truncation bar is shown while the Bin Width varies depending on custom range. 

https://github.com/VEuPathDB/web-monorepo/assets/12802305/b7edcb71-851c-46e3-b0b1-d6405c79aef0

asizemore:
Thanks @moontrip !

So far i've just really tested the branch a bit. I noticed that the truncation bars appear now when the x axis range is shortened, even if no data has been cut off. For example, below i've changed the x axis to start at 1 instead of 0. I can verify there's no data between 0 and 1 but the truncation bar shows up. I think we need to check if the left side of the very first bin is less than the user-inputted x axis range min, and if the right side of the final bin is greater than the user-inputted x axis range max. 

Screen Shot 2023-07-25 at 7 06 56 AM

Another question i have:
> default independent axis range for Histogram Viz should remain the same regardless of setting custom range

Shouldn't this be true for all plots, not just histogram? Stop me if im expanding scope but id imagine that all plots should have the same default axis ranges regardless of the custom range. Or am i misunderstanding?

moontrip:
Thanks @asizemore for your tests! 👍 I left my response next to your question in the following.

> So far i've just really tested the branch a bit. I noticed that the truncation bars appear now when the x axis range is shortened, even if no data has been cut off. For example, below i've changed the x axis to start at 1 instead of 0. I can verify there's no data between 0 and 1 but the truncation bar shows up. I think we need to check if the left side of the very first bin is less than the user-inputted x axis range min, and if the right side of the final bin is greater than the user-inputted x axis range max.

**DK:** So, in your test case of the lower side, I know what the problem is: it is an edge case. The problem is that by request, we have set the following condition: if metadata.displayMin does not exist, then 0 is used to determine the default min range. Or, though this is rare case, the metadata.rangeMin information is incorrect and contains 0. The test variable most likely falls into the former, thus default min range becomes 0 in your test case, not 1, which causes the issue you observed: we have overlooked this edge case for truncation. If we should resolve this, then a simple solution is probably to compute ranges with and without such a constrained condition and use proper range for judging truncation, but I will need to check carefully whether this approach causes side effects.

> Another question i have:
> 
> > default independent axis range for Histogram Viz should remain the same regardless of setting custom range
> 
> Shouldn't this be true for all plots, not just histogram? Stop me if im expanding scope but id imagine that all plots should have the same default axis ranges regardless of the custom range. Or am i misunderstanding?

**DK:** I am sorry if it makes you confusion. Yes, default independent axis range for all Viz should remain the same as you said. What I wanted to say that under current format (before this PR), default independent axis range for Histogram Viz was changed whenever changing independent axis range because the Viz triggers data request to retrieve best bin width from the backend.

bobular:
Hi @moontrip - I found the same problem as Ann (also GEMS1)
image

It's not an edge case, it's a problem that the for every visualisation axis _except the histogram x-axis_, we are able to show the yellow warning ONLY when actual data has been cut off (because none of the other visualizations use the viewport to achieve "semantic zooming" - the min/max range of the data can be determined from the back end response). The histogram x-axis is a known bug as mentioned in my comment below.
https://github.com/VEuPathDB/web-eda/issues/1295#issuecomment-1320050297

The only solutions involve making extra client requests (maybe just a request to `distribution` would be sufficient), or performing the check in the back end and returning truncation flags from the back end.

bobular:
Although I see that we still use the language "may have been truncated" - so maybe your solution (with some simplifications I can suggest) is the best for users right now.

bobular (changes requested):

packages/libs/eda/src/lib/core/utils/default-axis-range.ts
```diff
@@ -12,14 +12,16 @@ export function numberDateDefaultAxisRange(
   observedMax: number | string | undefined,
   /** are we using a log scale */
   logScale?: boolean,
-  axisRangeSpec = 'Full'
+  axisRangeSpec = 'Full',
+  histogramViz: boolean = false
```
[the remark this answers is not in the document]
moontrip: Thank you for your comments, @bobular 👍 Actually, I tried similar thing to change the second and fourth argument to include Custom case, but somehow I didn't try your suggestion. It would probably work: will test

moontrip:
> Hi @moontrip - I found the same problem as Ann (also GEMS1) image
> 
> It's not an edge case, it's a problem that the for every visualisation axis _except the histogram x-axis_, we are able to show the yellow warning ONLY when actual data has been cut off (because none of the other visualizations use the viewport to achieve "semantic zooming" - the min/max range of the data can be determined from the back end response). The histogram x-axis is a known bug as mentioned in my comment below. [VEuPathDB/web-eda#1295 (comment)](https://github.com/VEuPathDB/web-eda/issues/1295#issuecomment-1320050297)
> 
> The only solutions involve making extra client requests (maybe just a request to `distribution` would be sufficient), or performing the check in the back end and returning truncation flags from the back end.

@bobular I understand what you have in mind but just wonder (or want to make sure) if we need to wait for backend work or solve it at front end level. For Ann's case, it happened due to the difference between actual data (or even metadata, starting from 1 at min) and display range (starting from 0 due to our constrained condition). Defining actual default range (e.g., starting from 1, not 0) to judge the display of the truncation may solve Ann's issue.

bobular:
Taking a deep dive into this.  **All tests done with the current `main` branch.**

Let's use GEMS1 -> participant repeated measure -> **height** EUPATH_0010075 as our example variable
 
distributionDefaults | value
-- | --
displayRangeMin | 0
displayRangeMax | 121
rangeMin | 17.4
rangeMax | 117.4
binWidth | 0.88
units | "cm"

The default visualization (and subsetting looks the same/similar) looks like this:
image

Not shown is the default "custom range" which goes from 17.4 to 119.4.

Given that the minimum value is 17.4, we should be able to do a custom range starting at 10 and not get truncation warnings:

And indeed we can:

image

If we started the range at 20, we would hope to see a truncation warning, but no we don't. Well, we half do. We get one temporarily (calculated using the data from the previous back end call) and the blue warning text shows up and persists:

image

Note also that if we filter on a different variable (using a different variable so we don't get confused with "filter-aware" behaviours), e.g. "**Height, 2nd**" then the default "custom range" for the "Height" histogram is based on the filtered data from the back end, not the annotated ranges:

image

If we change the custom range start to 20 in the above, we get no warnings.

So from this we learn that  the truncation flags are determined from the data returned from the back end, not the range annotations.

We can also see that in the code (HistogramVisualization.tsx). If we follow backwards from the `truncationConfig` call, we see that the "reference" ranges (`independentAxisRange` in particular) are passed in as `defaultUIState` - which contains `independentAxisRange: defaultIndependentRange` and `defaultIndependentRange` comes from `useDefaultAxisRange` which, _in custom and auto modes only_, gets its min and max from `independentAxisMinMax` which comes from `histogramDefaultIndependentAxisMinMax` which uses `findIndependentAxisMinMax` to get the min binStart and max binEnd from the response data.

> an aside: it looks like the comment ` // using annotated range, NOT the actual data` in HistorgramVisualization.tsx is wrong!

OK, so where do we go from here? 

Here are some options:
1. use a new, separate back end request to the `distributions` endpoint to get the values for `independentAxisMinMax` - no other code needs to change (relative to `main` at the time of writing)
2. do not even attempt to determine the min and max from the data; remove `independentAxisMinMax` entirely, and call `useDefaultAxisRange(xAxisVariable, undefined, undefined, undefined, undefined, 'Full')` forcing 'Full' mode; this would then use the annotated ranges in all cases -  hmm but then auto-zoom is broken.
3. use the `useDefaultAxisRange` final argument hack I suggested here: https://github.com/VEuPathDB/web-monorepo/pull/376#discussion_r1273592361 - testing this on top of `main` I see that
  - auto-zoom works
  - custom zoom doesn't start out with the auto-zoom limits like every other viz does
  - truncation warnings are shown if the range is anywhere inside the annotated range - but "Data **may** have been truncated" is not a lie!

Option 2 is a no go. Option 3 fixes the missing truncation warnings but we end up with more warnings than strictly necessary and the initial custom zoom behaviour is non standard (it uses the annotated range, not the data range) - though that might be fixable.

**My preference is for Option 1**. It shouldn't be too tricky. See `subsettingClient.getDistribution` in HistogramFilter. The response even contains `statistics.subsetMin` and `statistics.subsetMax` - and I bet you can send an empty or missing `binSpec` and the back end will do something sensible (yes you can, `binSpec` is optional in the request). Just remember to pass the current `filters`.

bobular:
I'd also recommend calling the distribution service to obtain `distributionPromise` in a separate `usePromise` "before" the main `data = usePromise`.

Add `distributionPromise` as a dependency to the main `data = usePromise` and return `undefined` if `distributionPromise.value` is nullish (and maybe also if `distributionPromise.pending`  - see what we already do with `filteredCounts`).

Then we have to return the `distributionPromise.value.statistics.subsetMin` and `Max` as part of the main `data = usePromise` payload, so need to add two more props (`subsetMin` and `subsetMax`) to `HistogramDataWithCoverageStatistics`. 

Then pull `independentAxisMinMax` out of `data.value.subsetMin` and `data.value.subsetMax` in the useMemo instead of calling `histogramDefaultIndependentAxisMinMax`.

This "bundling" of the data from two sources into one usePromise payload will create a smooth experience for the user - no multiple re-renders when the various back end calls come back. I did [something similar here](https://github.com/VEuPathDB/web-monorepo/pull/370/files).

moontrip:
Thank you for your detailed comments and suggestions @bobular 👍 Although I didn't read all very carefully, from my understanding I tend to agree with your comments. Will do my homework next week!

moontrip:
@bobular I have reflected your changes, though they may not be exactly the same. I found it later that some implementations were not necessary, e.g., binSpec.displayRangeMin/Max values etc., but I just left it as was. 

Also, noticed that using stats were not sufficient when multiple filters are used: e.g., GEMS1 main variable: birth date at Histogram Viz, while filtering/subsetting birth date and age. From my observation, it should consider min/max of both distribution data and stats to accurately compute min/max: I made a note concerning this.

One thing I want to say that I used subsettingClient.getDistribution as I was not quite sure how to request distribution data without using subsettingClient.getDistribution. It seems to me that independent axis truncation is now working correctly, but it would be great if you can test/review them.

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -442,6 +451,85 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
+        },
+        (filters) => {
+          return subsettingClient.getDistribution(
+            studyMetadata.id,
+            vizConfig.xAxisVariable?.entityId ?? '',
+            vizConfig.xAxisVariable?.variableId ?? '',
+            {
+              valueSpec: 'count',
+              filters,
+              binSpec: {
```
bobular, line 494: Isn't it easier to leave out the `binSpec`?  

It's optional
https://veupathdb.github.io/EdaSubsettingService/api.html#resources:/studies/{study-id}/entities/{entity-id}/variables/{variable-id}/distribution:post

let the back end figure it all out...

moontrip: Thank you for your comment, @bobular  👍 IIRC, I just tested with empty sub-parameters such as binSpec.displayRangeMin etc. that complained, but honestly didn't try to strip out whole binSpec parameter. Will check if it works: yes this will definitely simplify things if doable.

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -442,6 +451,85 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
+            NumberVariable.is(xAxisVariable) ? 'number' : 'date'
+          ),
+        ],
+      };
+
+      return series;
+    }
+
+    return undefined;
+  }, [filters, xAxisVariable, vizConfig.xAxisVariable]);
```
bobular, line 527: \+ `subsettingClient` ?

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -442,6 +451,85 @@ function HistogramViz(props: VisualizationProps<Options>) {
     [options, providedOverlayVariable, providedOverlayVariableDescriptor]
   );
 
+  // get distribution data
+  const subsettingClient = useSubsettingClient();
+
+  const getData = useCallback(async () => {
```
bobular, line 457: I would call this `getDistributionData`

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -442,6 +451,85 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
+        ],
+      };
+
+      return series;
+    }
+
+    return undefined;
+  }, [filters, xAxisVariable, vizConfig.xAxisVariable]);
+
+  const getDistributionData = usePromise(
```
bobular, line 529: And I would call this `distributionDataPromise`

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -442,6 +451,85 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
+    useCallback(() => getData(), [getData])
```
bobular, line 530: `getData` is already wrapped in a `useCallback` - I don't think it needs another one.

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
bobular: I've also just removed the entire `binSpec` object from the request data and things seem to be working the same (and much more readable).

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -442,6 +451,85 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
+                .displayRangeMin ??
+                (xAxisVariable as DateVariable).distributionDefaults.rangeMin,
+              (xAxisVariable as DateVariable).distributionDefaults
+                .displayRangeMax ??
+                (xAxisVariable as DateVariable).distributionDefaults.rangeMax,
+              (xAxisVariable as DateVariable).distributionDefaults.binWidth,
+              (xAxisVariable as DateVariable).distributionDefaults.binUnits,
+            ];
+
+      const distribution = await getDistribution<DistributionResponse>(
```
bobular, line 480: I think it should be possible to call `subsettingClient.getDistribution()` directly, rather than as a callback to (same-named!) `getDistribution()`.  It will only return the "foreground" distribution, which is all we need.

I haven't tried this myself, but I assume it should be possible!

bobular (commented):
This is looking great and it works great except in the case you mentioned.
When the x-axis variable has been filtered on, auto-zoom doesn't work properly.
I will look into why in the next day or so.

I've left some stylistic and other code-related comments.

bobular:
> When the x-axis variable has been filtered on, auto-zoom doesn't work properly.

Well that's very odd. It's working fine now, and there are no new commits. Sorry about the fake news!

bobular:
Hi @moontrip - I've just looked through all the code and it's also looking good. I have two change requests still, however:

1. remove the `binSpec` from the request (trivial)
2. avoid the two `distribution` requests (no need for the "background" one with no filters)

let me know if you have any issues with that

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: will test but I think it can be added

moontrip:
> Hi @moontrip - I've just looked through all the code and it's also looking good. I have two change requests still, however:
> 
> 1. remove the `binSpec` from the request (trivial)
> 2. avoid the two `distribution` requests (no need for the "background" one with no filters)
> 
> let me know if you have any issues with that

Hi @bobular 👋 Thank you for your detailed observations and suggestions. 👍 Yes X-axis range control has been working fine from my test and I am glad that it was not an exception you might have found.

Will try to address your suggestions and get back to you.

moontrip:
@bobular I wonder if you could get a success by removing the binSpec from the request? By doing so, it threw errors and could not make it done from my tests.

bobular:
Sorry - the auto-zoom is still broken for e.g. GEMS1 "Height, 2nd" as x-axis on histogram when "Height, 2nd" is also filtered on.

I was, ahem, on the wrong branch earlier. 🤦

moontrip:
@bobular No problem and thank you for sharing the example. I quickly tested it myself but it looks just fine for me. Below are screeshots using Height, 2nd filter from the range of 60 (min) to 90 (max). Can you please share your numbers used for subsetting/filter, just in case?

a) full range

histogramVIz-truncation01


b) auto-zoom

histogramVIz-truncation02

bobular:
The problem with the broken auto-zoom when a variable is directly filtered is simply due to the use of `getDistribution` from `filter/utils.ts` 

This function strips out the current variable from the filters. We don't want it to do that.  If we call `subsettingClient.getDistribution` directly with the filters, it should be fine.

PS. dropping the `binSpec` works fine for me. I'll slack you the diff

PR #381, opened 2023-07-25
Add tooltip w/ visx customization to volcano plots

### LINKED ISSUE #66: Volcano: add tooltips
See details [here](https://github.com/VEuPathDB/web-monorepo/pull/19)

Tooltips should show the point label (using `pointID` to find display name), the log2 fold change, the p value, and the adjusted p value.

See [this example](https://airbnb.io/visx/docs/xychart) for getting started.

jernestmyers:
Resolves #66 

This PR imports our changes to the `@visx/xychart` in place of `airbnb/visx`'s latest version. I did so as follows:

- I forked the `airbnb/visx` repo, made the changes we needed to achieve the desired tooltip behavior, then [opened a PR](https://github.com/airbnb/visx/pull/1734) in `airbnb/visx`
- I branched off my `visx` fork's `xychart-issue-979-tooltip-bug` branch to create [the `deploy-visx-xychart` branch](https://github.com/jernestmyers/visx/tree/deploy-visx-xychart) where I introduced [the `gh-pages` dev dependency](https://www.npmjs.com/package/gh-pages) and a [simple `deploy` script](https://github.com/airbnb/visx/commit/38f036624d8f5557bf580920b98c7f4a6f5c3e67#diff-f70a6f56869bd42f8fdb3a2f687c977bec5b2519245424d3159826afa6fea46e)
- I built the `visx` code locally, referencing [their monorepo's root `package.json` scripts](https://github.com/jernestmyers/visx/blob/38f036624d8f5557bf580920b98c7f4a6f5c3e67/package.json#L38C34-L38C34)
- I ran the `deploy` script on the `deploy-visx-xychart` branch (from the `visx/xy-chart` directory, not the root of the monorepo), which [created a new branch with only the `visx/xychart` package](https://github.com/jernestmyers/visx/tree/visx-xychart).
- In this PR, I amended the `components` `package.json` to import the `visx/xychart` package from my fork, like:
```json
  "@visx/xychart": "https://github.com/jernestmyers/visx.git#visx-xychart",
```

<h3>Testing</h3>
This work can be tested in our normal workflow, by:

1. Checking out this branch (`66-volcano-plot-tooltips`)
2. Running `yarn` at the root of the monorepo
3. Spinning up a local `mbio-site` dev server (and skipping the cache just to be safe) via:
```sh
  yarn nx start @veupathdb/mbio-site --skip-nx-cache
```


packages/libs/components/src/plots/VolcanoPlot.css
```diff
@@ -0,0 +1,14 @@
+.VolcanoPlotTooltip > ul {
+  margin: 0;
+  padding: 0.25em;
+  list-style: none;
+  line-height: 1.5em;
+}
+
+.VolcanoPlotTooltip > ul > li {
+  font-weight: normal;
+}
```
jernestmyers, line 10: NOTE: I'm just glancing back over this and can't see why this would be necessary. Perhaps I'm misremembering something. If not, I'll remove when I address feedback!

packages/libs/components/src/types/plots/volcanoplot.ts
```diff
@@ -8,7 +8,7 @@ export type VolcanoPlotDataPoint = {
   // Used for thresholding and tooltip
   adjustedPValue?: string;
   // Used for tooltip
-  pointID?: string;
+  pointID?: string[];
```
dmfalke, line 11: **Nit** can this be renamed `pointIDs`?

jernestmyers:
Here's a screencast demonstrating the styles and the logic that combines pointIDs into a single tooltip when the coordinates are shared.

Note how the grey background uses black text whereas the other background colors use white text. This is to achieve an acceptable color ratio. We may want to render the background colors at the default opacity, which may be more reflective of the marker color as I'm unsure how many users will adjust the marker opacity anyway. Doing so may also necessitate changes to the text colors.

https://github.com/VEuPathDB/web-monorepo/assets/69446567/d96b4d97-fd10-4d3d-b97d-4d7f6492a714

dmfalke:
Regarding building our code with your changes to visx, here is an idea:

In your fork, set up a workflow that builds the code and commits it to a branch named something like `dist`. In our repo, we can specify your fork and the branch for the visx package. See https://yarnpkg.com/cli/add.

I'm not sure if we will have to do this for all visx packages.

I'll try to find an example workflow that does what I suggested.

jernestmyers:
**NOTE**: the directions for testing have been edited. We no longer need to do anything special to test this now that I've deployed the `visx/xychart` code into a branch on my fork and pointed our `components` `package.json` to that branch.

dmfalke (approved):
I think this looks great!

Issue #358, opened 2023-07-14
Polishing differential abundance


asizemore:
Parent ticket for many (hopefully small!) polishing tasks for the differential abundance app:

- [x] Check to ensure no values overlap in groupA and groupB. So if "peanuts" chosen in group A, disable it in group B and add tooltip that says "Values cannot overlap between groups" or something  like that (credit @jernestmyers ) https://github.com/VEuPathDB/web-monorepo/pull/388
- [x] Add swap button to group A group B https://github.com/VEuPathDB/web-monorepo/pull/396
- [x] Add empty vizx plot. Otherwise the panel looks really weird before the viz data is ready. https://github.com/VEuPathDB/web-monorepo/pull/390
- [x] Add legend (driven by VolcanoPlot, always the same labels, no need to have it be clickable) https://github.com/VEuPathDB/web-monorepo/pull/392
- [x] Add axis controls to x and y axis. Component is ready. Just need axis range control, no log transform or anything! https://github.com/VEuPathDB/web-monorepo/pull/404
- [x] Add opacity slider for markers. https://github.com/VEuPathDB/web-monorepo/pull/404
- [x] Add axis truncation indicators (remember use props+data to derive this instead of state!) IP here https://github.com/VEuPathDB/web-monorepo/pull/325
- [x] update numberInputRange to take increment prop (right now changes by 1, but we need it to change by 0.001 or so) https://github.com/VEuPathDB/web-monorepo/pull/384
- [x] Layout inputs and such nicer (may require update to LabelledGroup) https://github.com/VEuPathDB/web-monorepo/pull/396

jernestmyers:
Closing now that all sub-tasks have been merged.

PR #384, opened 2023-07-26
Add step prop to NumberInput component and use in VolcanoPlot

jernestmyers:
Relates to second-to-last item in https://github.com/VEuPathDB/web-monorepo/issues/358

Note: I passed `step={0.001}` to the `P-Value` input but the `Fold Change` step remains undefined, thus remains the same default behavior (step of 1).

### DIFF
--- a/packages/libs/components/src/components/widgets/NumberAndDateInputs.tsx
+++ b/packages/libs/components/src/components/widgets/NumberAndDateInputs.tsx
@@ -34,7 +34,7 @@ type BaseProps<M extends NumberOrDate> = {
   disabled?: boolean;
 };
 
-export type NumberInputProps = BaseProps<number>;
+export type NumberInputProps = BaseProps<number> & { step?: number };
 
 export function NumberInput(props: NumberInputProps) {
   return <BaseInput {...props} valueType="number" />;
@@ -82,6 +82,7 @@ function BaseInput({
   containerStyles,
   displayRangeViolationWarnings = true,
   disabled = false,
+  ...props
 }: BaseInputProps) {
   if (validator && (required || minValue != null || maxValue != null))
     console.log(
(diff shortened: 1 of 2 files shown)


packages/libs/components/src/components/widgets/NumberAndDateInputs.tsx
```diff
@@ -202,6 +206,7 @@ function BaseInput({
       <div style={{ display: 'flex', flexDirection: 'row' }}>
         <TextField
           InputProps={{ classes }}
+          inputProps={{ step }}
```
asizemore, line 209: This is magic to me. How does passing a number in here affect the step? I would have expected to have to pass an object with "step: {step}" or something. 
Also MUI making an `InputProps` and an `inputProps` is not cool lol.

asizemore (approved):
Looks great! I like the way the p-value adjuster works and that the fold change one is left as is. 
Thanks @jernestmyers !

packages/libs/components/src/components/widgets/NumberAndDateInputs.tsx
jernestmyers: >  Also MUI making an InputProps and an inputProps is not cool lol.

For real! Had to re-read the docs a couple times to figure this one out!

> This is magic to me. How does passing a number in here affect the step? I would have expected to have to pass an object with "step: {step}" or something.

This is using [object shorthand](https://attacomsian.com/blog/javascript-object-property-shorthand). Since `step` is both the name of the variable and the HTML attribute we're setting in the input, I went this route. In re-thinking this, I'm wondering if it's problematic that when `step` is undefined, we're passing in `{ step: undefined }`... 🤔

asizemore: ahhhhhhhhh object shorthand. Very cool! Oh JS, you surprise me in new ways every day!

PR #389, opened 2023-07-26
only qa.restricted #51836

aurreco-uga:
i know it needs css, and will have to add others..  but since it goes only to restricted i figured it could be merged


dmfalke (commented):
The menu is very hard to interact with. Should we try to fix that before merging this?

dmfalke (approved):

PR #390, opened 2023-07-27
Add empty placeholder viz for volcano plots

jernestmyers:
Resolves item 3 in https://github.com/VEuPathDB/web-monorepo/issues/358

Screenshot of empty placeholder viz:
image


asizemore (approved):
Looks great! Why have the one point in the empty data? Just so the whole XYChart will render?

asizemore:
Nope, i didn't have anything in mind at all! I think it's pretty clever!

PR #392, opened 2023-07-27
Add legend to volcano plots

jernestmyers:
Resolves item 4 in https://github.com/VEuPathDB/web-monorepo/issues/358


packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -138,87 +145,26 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
...
-  // of the log transform and the little bit of padding, or because axis ranges
-  // are supplied.
-  let xAxisMin: number;
-  let xAxisMax: number;
-  let yAxisMin: number;
-  let yAxisMax: number;
-  const AXIS_PADDING_FACTOR = 0.05; // The padding ensures we don't clip off part of the glyphs that represent
-  // the most extreme points. We could have also used d3.scale.nice but then we dont have precise control of where
-  // the extremes are, which is important for user-defined ranges and truncation bars.
+  // Set maxes and mins of the data itself from rawDataMinMaxValues prop
```
asizemore, line 148: since `rawDataMinMaxValues` is optional, why not determine xmins and maxes from the data if `rawDataMinMaxValues` isn't supplied?
One reason is that it would duplicate a lot of logic. Are there other reasons?

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -148,6 +159,146 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
...
+          max: dataXMax + (dataXMax - dataXMin) * AXIS_PADDING_FACTOR,
+        };
+      } else {
+        return { min: 0, max: 0 };
+      }
+    }
+  }, [data.value, xAxisRange, rawDataMinMaxValues]);
+
+  // NOTE: this state may be unnecessary depending on how we implement user-controlled axis ranges
+  const [yAxisRange, setYAxisRange] =
```
asizemore, line 212: why might this state be unnecessary?

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -148,6 +159,146 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
...
+    data.value,
+    independentAxisRange,
+    dependentAxisRange,
+    significanceThreshold,
+    log2FoldChangeThreshold,
+    significanceColors,
+  ]);
+
+  // TODO: Given that I cannot figure out how to type this correctly, I suspect there's a simpler, cleaner
+  // way to get the counts.
```
asizemore, line 282: the other way i could think of is `significanceColors.map(color, finalData.filter(blah = color).length())` but i think your way is more performant?

asizemore (commented):
I still need to test it but i read through it and it looks great!

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
jernestmyers: Unclear to me if we could just update and use the `vizConfig` data. I think that will become more clear in the follow up tasks to add axis controls and an opacity slider. 🤷

packages/libs/components/src/plots/VolcanoPlot.tsx
jernestmyers: That was really the only reason. I was hoping to determine the raw max/min values once. I was inclined to make them required, which broke all the `VolcanoPlot` stories. Since I wasn't convinced they should be required, I reverted and opted for optional. If you notice in lines 172ish of the `VolcanoPlotViz`, we default to all 0s, so `rawDataMinMaxValues` will always be passed in. Perhaps it's better to just update the stories and go with a required prop?

asizemore: yeah i think just make it required. A default of 0 won't give us anything interesting on the plot, so we might as well just require that value to be passed in.

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
jernestmyers: @dmfalke Do you have thoughts about this? I've found myself reaching for `reduce` a lot lately and I struggle to type it every time, which leads me to believe it's a code smell.

dmfalke: I'm looking at this now...

jernestmyers: Thanks for the feedback. Yours is far more digestible, so I'll go with that approach. I've been having too much fun with `reduce` lately! 😆

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -152,6 +160,130 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
...
+   * The standard x axis is the log2 fold change. The standard
+   * y axis is -log10 raw p value.
+   */
+
+  // Find maxes and mins of the data itself
+  const rawDataMinMaxValues: RawDataMinMaxValues = useMemo(() => {
+    if (!data.value)
+      return {
+        x: { min: 0, max: 0 },
+        y: { min: 1, max: 1 },
```
jernestmyers, line 174: NOTE: I changed the fallback y-values to be 1 so that the empty plot doesn't get truncated. Previously, the values were 0 and `Math.log10(0)` returns `-Infinity`, resulting in the yMax truncation bar rendering on the empty graph placeholder.

asizemore:
Even cleaner than before! It looks really nice thanks @jernestmyers !

asizemore (approved):

PR #396, opened 2023-07-28
Adjust diff abundance configuration layout

jernestmyers:
Resolves 2nd and last items from https://github.com/VEuPathDB/web-monorepo/issues/358

In the process of making these adjustments, I took the opportunity to pay off some tech debt by creating a new SASS file that replaces the inline styling used in our plugin configuration menus and in each plugin's description component.

1. New viz with no selections. Group A, swap button, and Group B disabled. Notice I'm using the native `title` tooltip but would we prefer a tooltip component?
image

2. Group comparison variable now chosen so Group A and Group B are enabled. Notice the swap button remains disabled. Also note that we are unable to render a tooltip explaining why it's disabled due to the way MUI handles disabling their components, but I suspect users will figure it out easily enough.
image

3. Swap button enabled by any group value selection, now that there's something to actually swap.
image

Hit me with your best feedback! 👊

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
@@ -99,10 +99,10 @@ function AbundanceConfigDescriptionComponent({
     )
   );
   return (
-    <>
-      <h4 style={{ padding: '15px 0 0 0', marginLeft: 20 }}>
+    <div className="ConfigDescriptionContainer">
+      <h4>
         Data:{' '}
-        <span style={{ fontWeight: 300 }}>
+        <span>
           {updatedCollectionVariable ? (
             `${updatedCollectionVariable?.entityDisplayName} > ${updatedCollectionVariable?.displayName}`
           ) : (
@@ -110,9 +110,9 @@ function AbundanceConfigDescriptionComponent({
           )}
         </span>
       </h4>
-      <h4 style={{ padding: 0, marginLeft: 20 }}>
+      <h4>
         Method:{' '}
-        <span style={{ fontWeight: 300 }}>
+        <span>
           {rankingMethod ? (
             rankingMethod[0].toUpperCase() + rankingMethod.slice(1)
           ) : (
@@ -120,7 +120,7 @@ function AbundanceConfigDescriptionComponent({
           )}
         </span>
       </h4>
-    </>
+    </div>
   );
 }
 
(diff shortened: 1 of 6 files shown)


asizemore:
Looking great! It's all very natural to use!


1. What are the pros/cons to using a true tooltip?
2. Agree. I think users can figure it out.
3. Lovely
4. Can the title popup more quickly? I was hovering for a bit and thought my browser was blocking the popup but it just turns out im not so patient 😄 
5. I noticed that the swap button comes on as soon as one group is chosen. I like it!
Screen Shot 2023-07-31 at 10 53 35 AM

I've just played around with it so far and haven't been through the code yet, but i should get a moment to this afternoon!

jernestmyers:
> - What are the pros/cons to using a true tooltip?
> - Can the title popup more quickly? I was hovering for a bit and thought my browser was blocking the popup but it just turns out im not so patient 😄

Well, you've already discovered the biggest con of using title: I'm not aware of any way to control _**how quickly**_ the tooltip renders! I was assuming we'd want the tooltip- and I'm expecting that it's a trivial insert- but I thought I'd present without just for a discussion, though @dmfalke is more likely to have a stronger opinion about tooltip vs title because of performance issues (and that native HTML has the added benefit of more out-of-the-box accessibility) we have run into in the past. But this is just one tooltip, so it's not necessarily a performance issue.

packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -11,14 +11,10 @@ import { Computation } from '../../../types/visualization';
...
+import { makeClassNameHelper } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
 
-export const sharedConfigCssStyles = {
-  display: 'flex',
-  gap: '0 2em',
-  padding: '1em 0',
-  alignItems: 'center',
-  marginLeft: '3em',
-};
+const cx = makeClassNameHelper('StepOneConfigurationContainer');
```
asizemore, line 17: Would this class name be confusing if we have a step 1 elsewhere that doesn't have to do with configuring a compute? Do you think this is general enough to be just `StepConfigurationContainer`? Or maybe specific to computes like `AppStepConfigurationContainer`?

asizemore (approved):
Quite lovely! Excited about that scss cleanup!

packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
jernestmyers: Hmm, these are good questions. I didn't suspect it would apply beyond app/plugin configurations, which is why I named the file `Plugins.scss`. While not convinced of my naming (ever, really), I think the fact that these styles live in `Plugins.scss` may be enough to avoid confusion? That said, I think `AppStepConfigurationContainer` may be a better name anyway.

PR #401, opened 2023-08-02
Some additional clean-up

dmfalke:
Move active viz to realm of map type config, and clean up some labels and ids.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -596,7 +596,6 @@ function MapAnalysisImpl(props: ImplProps) {
           children: [
             {
               type: 'item',
-              // concatenating the parent and subMenu labels creates a unique ID
               id: 'single-variable-pie',
               labelText: MarkerTypeLabels.pie,
               rightIcon: <DonutMarker style={{ height: '1.25em' }} />,
@@ -697,7 +696,6 @@ function MapAnalysisImpl(props: ImplProps) {
             },
             {
               type: 'item',
-              // concatenating the parent and subMenu labels creates a unique ID
               id: 'single-variable-bar',
               labelText: MarkerTypeLabels.barplot,
               leftIcon:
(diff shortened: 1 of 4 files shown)


packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -580,12 +592,12 @@ function MapAnalysisImpl(props: ImplProps) {
...
         {
           type: 'subheading',
-          labelText: 'Single Variable Maps',
+          labelText: MapSideNavItemLabels.SingleVariableMaps,
           children: [
             {
               type: 'item',
               // concatenating the parent and subMenu labels creates a unique ID
-              id: MapSideNavItemLabels.ConfigureMap + MarkerTypeLabels.pie,
+              id: 'single-variable-pie',
```
jernestmyers, line 600: nit: we can remove the preceding comment re: concatenating labels

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -684,7 +698,7 @@ function MapAnalysisImpl(props: ImplProps) {
             {
               type: 'item',
               // concatenating the parent and subMenu labels creates a unique ID
-              id: MapSideNavItemLabels.ConfigureMap + MarkerTypeLabels.barplot,
+              id: 'single-variable-bar',
```
jernestmyers, line 701: nit: same as above; comment can be removed

jernestmyers (approved):
LGTM 💪

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
dmfalke: Oh my, of course!

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
dmfalke: Yes!

PR #402, opened 2023-08-03
Filter out unknown attributes from user pref and step details

dmfalke:
This PR addresses an issue where step results will not load if user preference and/or step details includes an unknown attribute.

### Question: Should we "fix" the source of the problem?

This would be done by removing the unknown attributes from user prefs and step details.

If we don't fix it, we will get repeated error logs. On the other hand, if the attribute was mistakenly removed from the model, then the user will see the column in their results again, when it is added back.

If we do fix it, we will only get one error log per step, but the column will not reappear if it is added back to the model.

Thoughts?


aurreco-uga (approved):

dmfalke:
I spoke with @aurreco-uga and we decided to wait on "fixing" the source of the problem. If we start getting lots of error emails for this, we can revisit.

PR #404, opened 2023-08-04
Add plot controls to volcano plots

jernestmyers:
Resolves items 5 and 6 from https://github.com/VEuPathDB/web-monorepo/issues/358

~~NOTE: this work is branched off of https://github.com/VEuPathDB/web-monorepo/pull/392~~ (merged)

image

Notes:
- The calculations for the default axis ranges get very messy, so we decided (for now, at least) to use `Math.floor()` on the min values and `Math.ceil()` on the max values.
- Notice I'm not labeling these `X-axis controls` with `X-axis range beneath it and likewise for the y-axis. It felt weird to have the Header/Subheader relationship when the only controls are for the ranges.
- I didn't add the radio buttons for "Full", "Auto-zoom" and "Custom" as I wasn't sure if they were desired.

Happy to make any changes to the decisions/assumptions I made.


packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -166,6 +166,24 @@ const modalPlotContainerStyles = {
   margin: 'auto',
 };
 
+// implement gradient color for slider opacity
+export const colorSpecProps: SliderWidgetProps['colorSpec'] = {
```
asizemore, line 170: since we're going to use this elsewhere, shall we give it a more descriptive name? Maybe `sliderOpacityColorSpecProps`? Or something with "gradient color" in it?

asizemore: I wonder do we have anywhere more general for this to live?

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -75,15 +81,19 @@ function createDefaultConfig(): VolcanoPlotConfig {
...
     significanceThreshold: DEFAULT_SIG_THRESHOLD,
     markerBodyOpacity: 0.5,
+    independentAxisRange: undefined,
+    dependentAxisRange: undefined,
   };
 }
 
 export type VolcanoPlotConfig = t.TypeOf<typeof VolcanoPlotConfig>;
-
+// eslint-disable-next-line @typescript-eslint/no-redeclare
```
asizemore, line 90: was the second declaration giving you issues? Or is this to clean up warnings?

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -343,7 +347,123 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
   const plotNode = <VolcanoPlot {...volcanoPlotProps} ref={plotRef} />;
 
   // TODO
```
asizemore, line 349: can remove this comment woot!

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -343,7 +347,123 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
...
-  const controlsNode = <> </>;
+  const controlsNode = (
+    <div style={{ margin: '2em 1em' }}>
+      <SliderWidget
+        minimum={0}
+        maximum={1}
+        step={0.1}
+        value={vizConfig.markerBodyOpacity ?? 0.5}
```
asizemore, line 356: Got some feedback that the points are a little tough to see. Let's make the default 0.7 please!

asizemore (commented):
Looks great! Works perfectly! Can we put the marker opacity slider under Plot Controls please?

And i agree with all the notes you left. We don't need the auto-zoom, full, and custom because we don't have a difference between auto-zoom and full, since there are no display ranges here!

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
jernestmyers: I remember thinking the same thing, so I'll see what I can come up with!

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
jernestmyers: Just cleaning up ESLint warnings, now that VSCode screams at me! 😂

jernestmyers:
> Can we put the marker opacity slider under Plot Controls please?

Do we have any concerns that it lines up beneath the x-axis controls?
image

asizemore:
sorry my comment was too vague - i mean can the opacity live in its own section called Plot Controls? Like we do on the scatterplot where it says Plot Mode (marker opacity shouldn't really be in. the Plot Mode section, instead that section should be called Plot Controls or something... anyways... lol). My hope is to prevent the friendly opacity slider from floating around without a section header to guide it 😄 

Screen Shot 2023-08-08 at 6 45 40 AM

jernestmyers:
Updated screenshot:

image

asizemore (approved):
Looks great! Thanks @jernestmyers !

PR #412, opened 2023-08-08
Add basic network components

asizemore:
This PR adds two components that will help us towards our goal of creating a bipartite network (see figma mockups for correlation app)

I went round and round trying to figure out which components we needed. I've settled (so far!) on the following: `NodeWithLabel`, `Link`, and then `BipartiteNetwork` (for later work). I think if we have our own node and edge components then for a bit we don't need our own `Network` component. Eventually it will be helpful for setting a bunch of defaults, but i dont think we need to add it this moment. We'll see i guess when we add the real bipartite network! 

So anyways my goal was to keep these two components simple but flexible. Sort of like a core ui balance. I'm only allowing a particular amount of styling that i think will be actually used, though in the future i could see some style overrides or something of the sort being added. All that is to say my hope is that these properties for the node and link (edge) will serve us well for a while!

One reason why i broke out `NodeWithLabel` into its own component was that we would like the user to be able to click on the node or its label and have some things happen. Someday we may have something similar for the edges, but nothing is planned for right now in the bipartite network. 

Check out the [figma mockups](https://www.figma.com/file/5A893Wsvr61jRBNkZbJrAu/Mbio-Correlation-app?type=design&node-id=1-2&mode=design&t=p4Y3Zsnj2JwhSQvD-0) for a better idea of what we're hoping for in the end. Also happy to discuss further if helpful!


packages/libs/components/src/stories/plots/Network.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+ManyPoints.args = {
+  data: manyPointsData,
+};
+
+// Gerenate a network with a given number of nodes and random edges
+function genNetwork(nNodes: number, addNodeLabel: boolean) {
+  // Create nodes with random positioning, an id, and optionally a label
+  const nodes: NodeData[] = [...Array(nNodes).keys()].map((i) => {
+    return {
+      x: Math.floor(Math.random() * 500),
```
jernestmyers, line 61: I assume we're multiplying by 500 because the `<svg>` container has a height and width of 500? I think it would be easier to maintain this story if we used a variable for all the 500s instead.

packages/libs/components/src/stories/plots/Network.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+      y: Math.floor(Math.random() * 500),
+      id: String(i),
+      label: addNodeLabel ? 'Node ' + String(i) : undefined,
+    };
+  });
+
+  // Create {nNodes} links. Just basic links no weighting or colors for now.
+  const links: LinkData[] = [...Array(nodes.length).keys()].map(() => {
```
jernestmyers, line 69: Super-duper nit: I wonder if we should just use `nNodes` here instead of `nodes.length`? We use `nNodes` in the lines below as well, so it feels a bit out of place.

jernestmyers (approved):
I think this looks great.

I imagine it's on your radar, but something we'll have to think about is how to prevent nodes and/or their labels from getting cut off (see screenshot below). You may have some ideas in mind or perhaps our implementations of these components will obviate this concern. It seems non-trivial, but hopefully I'm wrong! 😄 

image

dmfalke (commented):
This looks cool! I'm wondering if it would make sense to have a `Network` component that does something similar to the `Network` story template?

Also, see my comment below.

dmfalke:
> This looks cool! I'm wondering if it would make sense to have a `Network` component that does something similar to the `Network` story template?

You can ignore this question. I didn't read the description very carefully!

dmfalke (approved):

asizemore:
> something we'll have to think about is how to prevent nodes and/or their labels from getting cut off

You're totally right @jernestmyers ! And indeed i don't think it'll be trivial at all :) but i also plan to engineer the visualization and user actions so that we don't have to run into that problem! For example i dont think we should have all labels showing at one time unless the network is in a very specified layout. Not only will the label get clipped possibly but they overlap and they're just so hard to read 😄

PR #414, opened 2023-08-10
add time series viz icon

### LINKED ISSUE #323: timeline plot viz icon
make the icon

asizemore:
Resolves #323 

Adds time series viz icon. Used in the map.

We've had lots of discussion around this icon in slack so i picked a version folks were okay with and went for it :) 
I'm hoping that once it's really on the site we can better decide how we may want to modify it or that it just needs to be distinct from other icons. 

Screen Shot 2023-08-10 at 6 25 48 AM


chowington (approved):
I like that you tagged me as the most skeptical one, relatively :) Looks good!

PR #418, opened 2023-08-14
Add row counts to record page tables

jernestmyers:
Resolves [redmine#51970](https://redmine.apidb.org/issues/51970).

**NOTE**: I opted against using [Mesa's `RowCounter` component](https://github.com/VEuPathDB/web-monorepo/blob/main/packages/libs/wdk-client/src/Components/Mesa/Ui/RowCounter.jsx). I felt the props in `RowCounter` would have required more paging state to be held in `DataTables` than we really need.

Screenshots of this work:
image
image


dmfalke (approved):
Looks great!

Issue #416, opened 2023-08-10
bubble markers: select all numerator values by default when switching variable


bobular:
I think that's self-explanatory?
Decided here (follow up items): https://veupathdb.atlassian.net/wiki/spaces/FG/pages/345735201/2023-08-09+Front+End+Dev+Meeting+Notes

PR #419, opened 2023-08-14
Selected all numerator values by default in bubble config

chowington:
Closes https://github.com/VEuPathDB/web-monorepo/issues/416


bobular (approved):
Tricky one!

Issue #403, opened 2023-08-03
On first map entry, default marker mode config panel should be fully open; plot visible in all menu "modes"; improve general persistence


bobular:
- [x] entry behaviour as in title
- [x] floating plots should be visible in any menu mode (just remove the conditional rendering)
- Above done, but (I don't think) by me (Connor)
- [ ] improve general persistence (optional - pls make another ticket if not doing)

As decided here
https://veupathdb.atlassian.net/wiki/spaces/WG/pages/337117185/2023-07-27+Map+UX#default-panel

Best to branch from/target to this branch https://github.com/VEuPathDB/web-monorepo/tree/map-collection-variable-map-type if it's still unmerged (see https://github.com/VEuPathDB/web-monorepo/pull/399)

I'm currently assuming that the default marker mode is the first marker mode (hardcoded in client nav panel, IIRC).

When the user first enters the map, the full marker panel should be open (in "Markers" tab).

Note that all UI state should be preserved in the user database. At the moment it's a bit hit and miss.

For example if I open a supporting plot, and minimise the menu (leaving the plot on screen), wait for network requests to stop, then reload the page, the page re-opens with the menu open (it should be minimised).

The goal is for the map to look identical upon reloading (or opening a share link) if in the same-sized browser window, and only slightly different if in a different size/shape window (e.g. legend and plot positions).

chowington:
For the second task, I assume that we don't want the vizs for a different map type to show for the currently-selected map type. Eg, if I have a viz open in the donut map type, when I click over to the bubble map type, the viz should disappear

chowington:
Assigning to @dmfalke as merging his PR (https://github.com/VEuPathDB/web-monorepo/pull/399) will fix this

PR #421, opened 2023-08-14
Side panel UX improvements

chowington:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/403

I'm marking this as ready to merge while we're still considering the idea of general UI persistence. Further work can be in another ticket.


bobular (approved):
Looks great. The functionality works for the marker mode and menu/drawer open/hidden state is preserved on page refresh.

I made a comment for the follow-on work https://github.com/VEuPathDB/web-monorepo/issues/444#issuecomment-1693225049

PR #429, opened 2023-08-17
Change rounding algorithm in bubble size legend

bobular:
This issue was that the maxSizeValue from the back end was 196,503. The `toPrecision(1)` rounded this up.  If it had been 149,999, it would have been rounded down and the max bubble size would have been 200,000 as desired.  I've changed the algorithm a bit to hopefully do the right thing.

Before:
image

After:
image


chowington (approved):
Good catch, @bobular!

PR #430, opened 2023-08-17
Reset paging if number of filtered rows obviates currentPage

jernestmyers:
Resolves https://github.com/VEuPathDB/web-eda/issues/1430

I added some comments to explain the solution, but I'll recap here as well. Whenever we search in a membership table, we will:
- check if the number of filtered rows is enough to render rows on the `currentPage`
  - yes? carry on as normal!
  - no? set `currentPage` to 1

This seemed a simple, low-cost solution to this problem. It has the drawback where a user may "lose" their paging after a search, but trying to maintain a user's pre-filtering paging would require additional state and complexity that I didn't think made sense in what's already a bit of an edge case.


dmfalke (approved):
This is a great solution.

PR #434, opened 2023-08-18
use variable display name in tooltip

### LINKED ISSUE #433: Volcano: tooltip use variable display names
The gorgeous volcano tooltip is currently showing the variable id, but it would be more helpful for the user if it showed the display name of that variable. 

There's a function `fixVarIdLabel` called in `ScatterplotVisualization` that takes a label and the collection entity information and returns the nice display label. We should be able to use that same machinery in the tooltip i'd guess.

asizemore:
Resolves #433 

This PR adds a `displayLabel` to the `VolcanoDataPoint` so that if defined, the tooltip can show the user-friendly name instead of the pointID we get from the backend. The pointID is formatted {entityID}.{variableID}, so we can use this to find a display label.

Not too much else to say about this one i think!

Screen Shot 2023-08-18 at 7 15 17 AM

### DIFF
--- a/packages/libs/components/src/plots/VolcanoPlot.tsx
+++ b/packages/libs/components/src/plots/VolcanoPlot.tsx
@@ -363,8 +363,8 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
                   }}
                 >
                   <ul>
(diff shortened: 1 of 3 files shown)


jernestmyers (approved):
LGTM!

PR #438, opened 2023-08-22
Resolve marker preview bugs w/ bar plot map type 

### LINKED ISSUE #326: Map - issues with barplot marker previews
Please check this is still a live issue after #190 is completed (and remove the `question` label).

### Issue 1
The logscale mode for the preview bar marker doesn't seem to be right.

Here are no-logs markers - the preview looks similar to the three map markers:
image

Here are logged markers - now they don't look at all similar:
image

_Originally posted by @bobular in https://github.com/VEuPathDB/web-monorepo/issues/285#issuecomment-1604791616_

### Issue 2
The proportion mode is not working in the bar marker preview since the data for the marker preview is coming from the subsetting service, which does not support `valueSpec: 'proportion'`

jernestmyers:
Resolves #326 

This PR addresses two bugs in marker previews:

### log scale bug
- the map markers receive a `dependentAxisRange` that the marker previews were not receiving
- providing the same `dependentAxisRange` to the marker preview appears to resolve this issue

### proportion mode w/ continuous vars
- in proportion mode, the data was erroneously adding together (in a reduce function) _**proportion**_ values that corresponded to the same overlay label in the same manner that we do for counts values; this resulted in an erroneous number displayed as `totalCount` in the marker preview
- now, we add a `count` property to the data that is used to determine `totalCount` irrespective of the plot mode

### DIFF
--- a/packages/libs/components/src/map/ChartMarker.tsx
+++ b/packages/libs/components/src/map/ChartMarker.tsx
@@ -323,7 +323,12 @@ export function getChartMarkerDependentAxisRange(
   isLogScale: boolean
 ) {
   return {
-    min: isLogScale ? 0.1 : 0, // matches what we pass into map markers
+    min: isLogScale
+      ? Math.min(
+          0.1,
+          ...data.filter(({ value }) => value > 0).map(({ value }) => value)
+        )
+      : 0,
     max: Math.max(...data.map((d) => d.value)),
   };
 }
--- a/packages/libs/components/src/map/ChartMarker.tsx
+++ b/packages/libs/components/src/map/ChartMarker.tsx
@@ -315,3 +315,13 @@ function chartMarkerSVGIcon(props: ChartMarkerStandaloneProps): {
     sumValuesString,
   };
 }
+
+export function getChartMarkerDependentAxisRange(
+  data: ChartMarkerProps['data'],
+  isLogScale: boolean
+) {
+  return {
+    min: isLogScale ? 0.1 : 0, // matches what we pass into map markers
+    max: Math.max(...data.map((d) => d.value)),
+  };
+}
--- a/packages/libs/components/src/map/ChartMarker.tsx
+++ b/packages/libs/components/src/map/ChartMarker.tsx
@@ -23,6 +23,7 @@ export interface ChartMarkerProps
   borderWidth?: number;
   data: {
     value: number;
+    count?: number;
     label: string;
     color?: string;
   }[];
--- a/packages/libs/components/src/map/ChartMarker.tsx
+++ b/packages/libs/components/src/map/ChartMarker.tsx
@@ -15,18 +15,19 @@ import {
   MarkerScaleDefault,
 } from '../types/plots';
 
+export type BaseMarkerData = {
+  value: number;
+  label: string;
+  color?: string;
+};
+
 export interface ChartMarkerProps
   extends BoundsDriftMarkerProps,
     MarkerScaleAddon,
     DependentAxisLogScaleAddon {
   borderColor?: string;
   borderWidth?: number;
-  data: {
-    value: number;
-    count?: number;
-    label: string;
-    color?: string;
-  }[];
+  data: BaseMarkerData[];
   isAtomic?: boolean; // add a special thumbtack icon if this is true (it's a marker that won't disaggregate if zoomed in further)
   // changed to dependentAxisRange
   dependentAxisRange?: NumberRange | null; // y-axis range for setting global max
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -557,17 +557,19 @@ function MapAnalysisImpl(props: ImplProps) {
       ...(data.color ? { color: data.color } : {}),
     }));
     /**
-     * In the chart marker's proportion mode, the values are pre-calculated proportion values so summing them for totalCount is inaccurate.
-     * Instead, we'll reduce the counts and use them later when determining totalCount and when determining the data for
-     * ChartMarkerStandalone.
+     * In the chart marker's proportion mode, the values are pre-calculated proportion values. Using these pre-calculated proportion values results
+     * in an erroneous totalCount summation and some off visualizations in the marker previews. Since no axes/numbers are displayed in the marker
+     * previews, let's just overwrite the value property with the count property.
      *
-     * NOTE: the donut preview doesn't care about the counts, but it's receiving them anyway.
+     * NOTE: the donut preview doesn't have proportion mode and was working just fine, but now it's going to receive count data that it neither
+     * needs nor consumes.
      */
-    const dataWithCounts = typedData.reduce(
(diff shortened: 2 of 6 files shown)


packages/libs/components/src/map/ChartMarker.tsx
```diff
@@ -23,6 +23,7 @@ export interface ChartMarkerProps
   borderWidth?: number;
   data: {
     value: number;
+    count?: number;
```
bobular, line 26: It seems a shame to introduce this as an optional prop when our EDA usage depends on it (and requires the `count` prop in the `StandaloneMapMarkersResponse` io-ts type. In MapAnalysis we have to fall back to zero so it compiles:
`          count: (data.count ?? 0) + prevData[index].count,`


Of course adding as optional is the right thing to do for backwards compatibility. But we are only using it in storybook stories and the EDA, so let's see how much would need to be changed if we made it compulsory...  A lot of the storybook code would need `{value: 100, count: 100}` duplication (and the counts would be unused), and then also some changes would be needed in src/lib/core/hooks/mapMarkers.tsx and src/lib/map/analysis/MarkerConfiguration/CategoricalMarkerPreview.tsx

I guess the problem here is that we've converted `StandaloneMapMarkersResponse` data into `xxxMarkerProps` data well before it is used to create markers and _while we still need to do calculations on it_.

It seems like a fudge to add an optional prop to a low level component (in `components`) so that operations in higher level components (`eda`) can work, and that the lower-level component will never use.

So I think the solution should lie purely in the EDA code - but maybe check with @dmfalke because big changes are being made to the marker data code. Merge conflict alert!

If Dave's branch didn't exist, I'd suggest creating the following types - probably in `standaloneMapMarkers.tsx` 

```
interface DonutMarkerPropsWithCounts extends DonutMarkerProps {
  data: DonutMarkerProps['data'] & {
    count: number;
  }[];
}
interface ChartMarkerPropsWithCounts extends ChartMarkerProps {
  data: ChartMarkerProps['data'] & {
    count: number;
  }[];
}
```

And `useStandaloneMapMarkers` would return these types instead of the vanilla versions.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -503,25 +513,46 @@ function MapAnalysisImpl(props: ImplProps) {
...
+      const dependentAxisLogScale =
+        activeMarkerConfiguration &&
+        'dependentAxisLogScale' in activeMarkerConfiguration
+          ? activeMarkerConfiguration.dependentAxisLogScale
+          : false;
+      const isPlotModeProportion =
+        activeMarkerConfiguration &&
+        'selectedPlotMode' in activeMarkerConfiguration &&
+        activeMarkerConfiguration.selectedPlotMode === 'proportion';
+      const finalChartMarkerData = isPlotModeProportion
```
bobular, line 538: There's no real need for this ternary switch statement.  The preview marker will look exactly the same if we divide all the counts by `totalCount` or not. The preview marker has no mouse-overs or numbers/axes displayed so it's really just a matter of coloured pixels.

We could keep this in if it feels better though!

bobular (commented):
Works great, but I didn't realise the `count` prop was going to be so problematic.

I've made some comments about how I think the data typing should work. Let me know if you need further clarification. Probably best to discuss with Dave and me in the context of his branch.

packages/libs/components/src/map/ChartMarker.tsx
dmfalke: What about combining the marker props in `useStandaloneMarkers`? This could be controlled by an option, maybe `summarizeMarkers: boolean`.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
jernestmyers: I went ahead and removed this logic.

packages/libs/components/src/map/ChartMarker.tsx
jernestmyers: I opted for @bobular's approach as it felt more straightforward in my hazy brain!

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -484,17 +485,26 @@ function MapAnalysisImpl(props: ImplProps) {
     const initialDataObject = previewMarkerData[0].data.map((data) => ({
       label: data.label,
       value: 0,
+      count: 0,
       ...(data.color ? { color: data.color } : {}),
     }));
     const typedData =
       markerType === 'pie'
         ? ([...previewMarkerData] as DonutMarkerProps[])
```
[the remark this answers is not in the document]
jernestmyers: Good catch!

packages/libs/components/src/map/ChartMarker.tsx
```diff
@@ -315,3 +317,13 @@ function chartMarkerSVGIcon(props: ChartMarkerStandaloneProps): {
     sumValuesString,
   };
 }
+
+export function getChartMarkerDependentAxisRange(
+  data: ChartMarkerProps['data'],
+  isLogScale: boolean
+) {
+  return {
+    min: isLogScale ? 0.1 : 0, // matches what we pass into map markers
```
bobular, line 326: 0.1 would be fine as a minimum for count-based dependent axes, but I see that in at least one place (`chartMarkerData` for `continuousMarkerPreview` in MapAnalysis.txt) proportion data is being passed.

So something like this would be safer, in case there are positive values smaller than 0.1.

```
min: isLogScale ? Math.min(0.1, ...data.filter(({value}) => value > 0).map(({value}) => value)) : 0,
```

bobular: In fact, it seems to be an issue with continuous vars...

before:
image

after:
image

Hmm, it's better, but something's still not right. It doesn't look roughly like all the map markers:

image

bobular: OK, it seems like the answer is NOT to do `value: d.count / totalCount,` in MapAnalysis.tsx:L601 

It's fine to pass the regular `dataWithCounts` to `ChartMarkerStandalone` and `getChartMarkerDependentAxisRange`

The maths for the proportions doesn't quite work out properly, but I think it's OK because they look kind of right, and a bit different from the count-based log preview markers, which is desirable, to give some UI feedback when the user changes from counts to proportions (in log mode or not).

Let's look at this again when you've had a chance to have another look.

bobular (commented):
Everything looks great - let's just have another look at the minimum thing I pointed out.

packages/libs/components/src/map/ChartMarker.tsx
bobular: To be really strict - the "proportion to counts to sum to proportions" is probably necessary but I'm not sure we need to. Let's see how it looks/behaves.

bobular (approved):
Tested fully and it's looking great.

I properly reviewed the final commit and that looks great too.  I think I had looked at earlier work already so let's call this a wrap!

PR #445, opened 2023-08-24
Tweak mbio layouts

### LINKED ISSUE #357: Mbio: misaligned viz pickers
Screen Shot 2023-07-12 at 5 13 23 AM

This happened when i changed the name of the beta div scatter to "Dimensional reduction scatter plot". It's a looong name! I just realized it creates a misalignment with the other viz pickers. 

I think this is our first viz title that is wider than the picker icon. Should we align the picker icons with the titles then possibly stretching outside of the nice line created with the picker alignment, or align the leftmost part of each picker + title combo ( present)? I'd vote for aligning the picker icons since i think those are better at making the visual line when all aligned, but curious what others think! @jernestmyers  @chowington ?

### LINKED ISSUE #443: diff abund layout: make threshold line inputs side by side
Can we please change the vertically stacked threshold line inputs to a horizontal layout? So that then the two inputs are side by side? For the regular vizs we stacked the x and y variables because of space concerns iirc. We don't have stratification vars in the volcano viz so we don't have the same space concerns. Would it be possible to override the vertical stacking in this case?

Screen Shot 2023-08-23 at 10 45 58 AM

jernestmyers:
Resolves #443 and #357

Regarding issue 357 (viz picker alignment), I opted to set the width of the container to `13em`, which enforces a similar wrapping of our longest description string on the live site while fixing the icon misalignment. Personally, it doesn't bother me that the text extends beyond the icon, which is why I opted not to set the container's width to `10em`, equal to the button's width.

image

And here's a screenshot that reflects the changes for issue 447:
image

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
@@ -339,6 +339,7 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
     dependentAxisRange,
     significanceThreshold,
     log2FoldChangeThreshold,
+    entities,
   ]);
 
   // For the legend, we need the counts of the data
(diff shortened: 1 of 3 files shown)


packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -339,6 +339,7 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
     dependentAxisRange,
     significanceThreshold,
     log2FoldChangeThreshold,
+    entities,
```
asizemore, line 342: nice find!!

asizemore (approved):
Looks great! I like 13em and i agree it's totally fine to have the text extend.

Thank you for picking up these tickets!

PR #467, opened 2023-09-01
Feature 360 diff abund cont vars

### LINKED ISSUE #360: Differential abundance: add continuous variables
Currently diff abund only works for categorical variables. Add the ability to use continuous variables.

The continuous variable selection options (the checkbox options in the Group A, Group B dropdowns) should be bins like how we do continuous variables in the map. I believe we decided on using quartiles? 

Depends on https://github.com/VEuPathDB/microbiomeComputations/issues/41

asizemore:
Resolves #360 

The main goal was to add support for continuous variables to the differential abundance app. In doing so on the backend, we ended up changing the api a bit, so be warned!

Major changes:
1. The groups in the `Comparator` are now of type `LabeledRange` instead of `string[]`. This is the main api change that allows the continuous vars to work. Continuous vars will always have defined min, max, and label, but the categorical vars will only have a label. 
2. New request to the backend to get the continuous variable bins (thank you, map, for your endpoints!)
3. Added a loading spinner to the `SelectList`. When using a local backend, the requests for the continuous var bins were taking a really long time so i wanted to add the spinner. Now that all the backend is merged and we can use the qa site as the backend, the requests are very snappy and we almost never see the spinner! So if we're happy with how i've implemented the spinner we can keep it, or since it's not as big of a value add as i had thought, if this implementation needs work we should just pull it out of the PR. I was a bit unsure about importing a components component into coreui... it just felt wrong lol.

Note for reviewers: _please use the qa mbio backend. The volcano plot reqeust is failing for some backend reason that i haven't figured out yet. I _have_ successfully made a volcano plot with this frontend code using a local backend, so it's definitely just changing some backend things and no frontend issue that is causing the 422._ Update 09/11/23: qa backend is working! Testing is now easier!

To Dos:

- [x] Use the distribution endpoint to get bins
- [x] Update the diff abund request to send the new `Comparator` object
- [x] Check response :) 
- [x] Update display labels
- [x] Handle the selector waiting time
- [x] Dont request job status until all vars are selected


asizemore:
@jernestmyers might be interested in how this all shapes up!

packages/libs/coreui/src/components/inputs/SelectList.tsx
```diff
@@ -1,12 +1,15 @@
 import { ReactNode, useEffect, useState } from 'react';
 import PopoverButton from '../buttons/PopoverButton/PopoverButton';
 import CheckboxList, { CheckboxListProps } from './checkboxes/CheckboxList';
+import Spinner from '@veupathdb/components/lib/components/Spinner';
```
dmfalke, line 4: This breaks the build, because it causes a circular dependency. You might need to move the `Spinner` component to this repo. Or, just use text, like **Loading...**?

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -207,10 +222,54 @@ export function DifferentialAbundanceConfiguration(
...
+        ...configuration.comparator?.variable,
+        filters: filters ?? [],
+        dataClient,
+        binningMethod: 'quantile',
+      };
+      const bins = await getBinRanges(binRangeProps);
+      return bins;
+    }, [
+      dataClient,
+      configuration,
```
dmfalke, line 246: I would change this to
```suggestion
      configuration.comparator?.variable,
```

As it is, a request is made when the user chooses bins in the dropdown.

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -207,10 +222,54 @@ export function DifferentialAbundanceConfiguration(
...
+          min: bin.binStart,
+          max: bin.binEnd,
+          label: bin.binLabel,
+        } as LabeledRange;
+      })
+    : selectedComparatorVariable?.variable.vocabulary?.map((value) => {
+        return {
+          label: value,
+        } as LabeledRange;
+      });
```
dmfalke, line 271: Instead of using a type cast, adding a return type to the callback function would be better. Type casts are more aggressive and can sometimes allow incorrect types to pass.
```suggestion
  const groupValueOptions = continuousVariableBins.value
    ? continuousVariableBins.value.map((bin): LabeledRange => {
        return {
          min: bin.binStart,
          max: bin.binEnd,
          label: bin.binLabel,
        };
      })
    : selectedComparatorVariable?.variable.vocabulary?.map((value): LabeledRange => {
        return {
          label: value,
        };
      });
```

dmfalke (changes requested):
This is looking good! I made some suggestions. Let me know if you have any questions!

packages/libs/coreui/src/components/inputs/SelectList.tsx
asizemore: "Loading..." is way better 👍

asizemore:
@dmfalke ready for your review again!

dmfalke (approved):
:rocket:

Issue #472, opened 2023-09-06
Remove temporary hack: frontend app constraints


asizemore:
Once the new backend app constraints are up, remove the frontend constraints. See #471

d-callan:
going to reference [this ](https://github.com/VEuPathDB/web-monorepo/pull/526)here, as its related/ i modified the existing constraints for diff abund

asizemore:
going to start thinking about this as we add correlation

PR #488, opened 2023-09-11
Fix 471 appconstraints

### LINKED ISSUE #471: Temporary hack: frontend app constraints
*** To be removed in b66! ***

Differential abundance is nearly ready to go out for 65 but a final big part that is missing is constraining the app so that only the proper collections are available to the user. Only a few collections (absolute abundance collections) work in the differential abundance app, so let's just show the user those instead of all the collections.

The proper fix includes some infrastructure work, which is not going to make it for b65 but 66 looks good. The frontend hack we implement here can make use of the new collections annotations to remove from each app the improper collections.

asizemore:
Resolves #471 and must have for b65

This PR adds a temporary frontend filtering mechanism for the collection variables. In our new database we have both Relative Abundance data and Absolute Abundance data (along wiht lots of other data types). Each app can only handle _either_ the relative abundance or absolute abundance data. The collection vars now include some added properties that we can use to do this filtering on the frontend. 

However, the frontend is _not_ where this filtering should go long term! The infra team is hoping to work on a mechanism for the backend to declare which collections are appropriate for which app, in a way that's similar to how the vizs say they can take categorical but not continuous vars on the x axis, for example. So, #472 was created to ensure we delete these frontend filtering steps when the real solution is ready :) 

To check to ensure my logic is correct, each app should have only one set of Kingdom, Phylum, Class, etc. in the data dropdown. The differential abundance app should _only_ have these options, and no others (no pathway abundance or coverage or anything else). The other three apps can have options that include pathway abundance and coverage and such.

Also note that I can't (or not sure how to?) test these constraints for DIY datasets until it's merged and on qa. The DIY datasets shouldn't have any of these nice annotations, so i've just said if the annotation does not exist go ahead and pass the collection along.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
@@ -150,14 +150,18 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
   );
 
   const collectionVarItems = useMemo(() => {
-    return collections.map((collectionVar) => ({
-      value: {
-        variableId: collectionVar.id,
-        entityId: collectionVar.entityId,
-      },
-      display:
-        collectionVar.entityDisplayName + ' > ' + collectionVar.displayName,
-    }));
+    return collections
+      .filter((collectionVar) => {
+        return collectionVar.normalizationMethod !== 'NULL';
+      })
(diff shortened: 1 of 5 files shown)


dmfalke (approved):
This looks great!

Issue #426, opened 2023-08-16
Map - slim down Donut marker config panel to a reasonable width


bobular:
Thanks to #403 the user will see the map and the marker config on first entry. It would be good if we can minimise how much of the map is obscured by the panel.

In the screenshot below we can see that the "Binning method" radio buttons are very wide and do not even need to be there for categoricals.

Tasks and things to consider

- [ ] remove Binning method radio buttons for variable types that don't need it (presumably categorical)
- [ ] narrow the radio buttons for when they are needed - the default featured variable could be continuous
- [ ] Remove the "Donut marker controls" heading? It doesn't save any width but it does save height and is not telling the user anything useful
- [ ] Think about long variable names in the variable picker (max width/ellipsis if not already implemented?)
- [ ] Think about long value names (in the categorical config table) (ellipsis or wrap?)

![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/13faddea-6487-450e-9306-0904ad59798b)

PR #491, opened 2023-09-11
slim down donut and bar plots panels

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/426. Detailed descriptions are listed in the following:

- requirements (taken from the ticket)
- [x] remove Binning method radio buttons for variable types that don't need it (presumably categorical)
: addressed

- [x] narrow the radio buttons for when they are needed - the default featured variable could be continuous
: I think that one of the best approach is to use acronym for certain radio button label: there are not much room for reducing the spacing in the radio buttons, I suppose. I chose deviation as dev. as it is commonly used as a short word: or we may use Std as the acronym of the standard deviation. Just in case, full word is explained at button title, but perhaps a text below the radio buttons can be utilized as well.

- [x] Remove the "Donut marker controls" heading? It doesn't save any width but it does save height and is not telling the user anything useful
: addressed

- [x] Think about long variable names in the variable picker (max width/ellipsis if not already implemented?)
- [x] Think about long value names (in the categorical config table) (ellipsis or wrap?)
: for both cases, there were no implementation on the ellipsis for both variable picker and data table. After seeking out a solution, perhaps the easiest solution is to set max width, then lengthy strings are naturally wrapped

In addition, barplot width is also adjusted so that it can be fitted to the max width. And some margins are changed so that it can be better displayed overall.

Screenshots are attached in the following for demonstrating above changes:

- original layout
sam donut layout01

- after changing: categorical variable with lengthy table values
sam donut layout01-2

- after changing: continuous variable (with lengthy variable at the variable picker). As mentioned above, (Dev. = Deviation) may be located in the bottom of the radio buttons
SAM donut layout02

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
@@ -214,10 +214,10 @@ export function PieMarkerConfigurationMenu({
               // marginTop: 20,
             }
           }
-          label="Binning method (Dev. = Deviation)"
+          label="Binning method"
           selectedOption={configuration.binningMethod ?? 'equalInterval'}
           options={['equalInterval', 'quantile', 'standardDeviation']}
-          optionLabels={['Equal interval', 'Quantile (10)', 'Standard dev.']}
+          optionLabels={['Equal interval', 'Quantile (10)', 'Std. dev.']}
           buttonColor={'primary'}
           // margins={['0em', '0', '0', '1em']}
           onOptionSelected={handleBinningMethodSelection}
--- a/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
@@ -193,9 +193,9 @@ export function PieMarkerConfigurationMenu({
         />
       </div>
       <div style={{ margin: '5px 0 0 0' }}>
-        <span style={{ fontWeight: 'bold' }}>
+        <div style={{ fontWeight: 'bold', marginBottom: '0.5em' }}>
           Summary marker (all filtered data)
-        </span>
+        </div>
         {overlayConfiguration?.overlayType === 'categorical' ? (
           <CategoricalMarkerPreview
             overlayConfiguration={overlayConfiguration}
--- a/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
@@ -17,7 +17,6 @@ import { CategoricalMarkerPreview } from './CategoricalMarkerPreview';
 import Barplot from '@veupathdb/components/lib/plots/Barplot';
 import { SubsettingClient } from '../../../core/api';
 import RadioButtonGroup from '@veupathdb/components/lib/components/widgets/RadioButtonGroup';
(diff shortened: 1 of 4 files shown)


packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/PieMarkerConfigurationMenu.tsx
```diff
@@ -203,21 +207,17 @@ export function PieMarkerConfigurationMenu({
...
-      <LabelledGroup label="Donut marker controls">
+      {overlayConfiguration?.overlayType === 'continuous' && (
         <RadioButtonGroup
           containerStyles={
             {
               // marginTop: 20,
             }
           }
-          label="Binning method"
+          label="Binning method (Dev. = Deviation)"
```
bobular, line 217: Not too keen on the `(Dev. = Deviation)` in the header

I think `Std. dev.` would be perfectly understandable by anyone who already understands "Quantile".

Ideally mouse-overs would clarify but it's not worth implementing just for this. Maybe leave a comment for the future that we should add this if/when we implement mouseovers in RadioButtonGroup?

bobular (changes requested):
Looks great - thanks very much.

I have one trivial change to request and a comment about the table width for the Sample.species variable in the megastudy (qa.vectorbase.org back end at the moment)

I'm getting a horizontal scroll bar (see below). 

image

Not the end of the world. One of the issues is that the heading "Distribution" is quite long. I can get a reasonable compromise by adding this CSS rule manually in dev tools. 

```
.MesaComponent td {
	word-break: break-word;
}
```
This allows the long species name words (e.g. "quadrimaculatus") to be split over lines, but unfortunately without a hyphen. 

image

Oh, wait, there's `hyphens: auto`

```
.MesaComponent td {
	word-break: break-word;
	hyphens: auto;
}
```
image

Hmm, but that's breaking too aggressively now.  "al-bopictus" and "com-plex" are not necessary.

It looks a bit better without the `word-break: break-word` but the ideal solution would be to break when only absolutely necessary (e.g. like with `word-break: break-word`) and to use hyphens. Is that possible?

If not, I'd go with `word-break: break-word; hyphens: auto;` if that's easy to pass to Mesa as a custom TD style?

bobular:
> If not, I'd go with `word-break: break-word; hyphens: auto;` if that's easy to pass to Mesa as a custom TD style?

And if that's not possible, maybe change the final column heading to "Distrib."?  Or do some fancy CSS to make it shorten to Distrib. when necessary (or ellipses or whatever).

moontrip:
@bobular Thanks again for your detailed review/observations/suggestions. In view of release date tomorrow, the simplest approach is: a) use Std. Dev.; b) use Distr. instead of Distribution which is quite simple as the name/label is defined in the frontend.

That being said, one of my concern to merge this PR is that this work is only for Donut config panel, not thinking of the other panel, Bar plots (chart marker) panel which has quite wide width: Bubble one is okay. Perhaps just moving Marker Y-axis control to the next line of Marker Y-axis control would work as a simple solution?

What do you think?

bobular:
> What do you think?

Hi. "Distrib." should be OK (not "Distr.") - and "Std. dev." note the lowercase "d" consistent with casing of "Equal interval" will be good.

I think your suggestions for bar markers (good catch!) would be good - wrap the (presumably flex) marker axis controls to the next line and do the same changes to the radio group (and DRY up if time permits).

bobular:
Or maybe "Chart". 

image

It's probably not trivial to make the count heading span both columns?  That would also work.

bobular (approved):

moontrip:
For record, I scaled down the preview marker from 3 to 2.5 which is better I suppose. Also added some spacing between the title and the preview marker. Thanks @bobular for your reviews and great suggestions 👍 Will merge this PR. 

- scale 3 (previous)
SAM-Chart-panel022

- scale 2.5 (changed)
SAM-Chart-panel021

Issue #469, opened 2023-09-05
Volcano: plot points with pvalue=0


asizemore:
in the diff abund app sometimes we get points with a pval=0. Since the y axis is -log(pvalue) this value should get plotted at infinity. Because our screens aren't that large yet, we need another solution.

One option is to plot these at the top of the plot always. So they just live along the top axis.

asizemore:
also note i dont think these inf points are being included in the legend counts

PR #497, opened 2023-09-12
Feature 469 volcano infinity

asizemore:
Begins to address #469 

Scoping work for how we want to handle points with pvalue = 0. Who knew there were so many??

(latest screenshot)

Screen Shot 2023-09-12 at 1 53 43 PM


jernestmyers (approved):
Didn't test anything on a dev server (down for release), but the code looks good to me.

PR #504, opened 2023-09-14
Add condition to deleting marker floaters logic

### LINKED ISSUE #427: Map - deleting a floating plot from the supporting plots list has unwanted side effects
If we have floaters A and B and B is currently active. If I click the trash can icon for A, I would expect B to stay on the screen, but it does not. I think somehow A is getting selected by the click, just before being deleted.

jernestmyers:
Resolves #427 

Previously, deleting a floater always invoked `setActiveVisualizationId(undefined)`. Now, we'll only do so if the viz being deleted is the active viz.

**NOTE**
The deletion will update `analysisState`, causing the loader to spin on the active viz. This may be another thing to consider when we have our deep think about our data requests. As a user, I'd be a bit confused why deleting a non-active viz results in a loading state. This issue feels a bit out of scope for this ticket, but perhaps it's not? Thoughts on if I should investigate further in the context of this ticket, or should it wait until our proper deep think?

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapVizManagement.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapVizManagement.tsx
@@ -243,7 +243,12 @@ function VisualizationsList({
                               analysisState.deleteVisualization(
                                 viz.visualizationId
                               );
-                              setActiveVisualizationId(undefined);
+                              if (
+                                activeVisualization?.visualizationId ===
+                                viz.visualizationId
+                              ) {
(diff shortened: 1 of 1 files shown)


dmfalke (approved):
This is good. Regarding the spinner, I didn't see it in my test. I had a timeseries plot open when I deleted a line plot. Maybe it depends on the type of the open viz?

bobular:
I agree we should review some of the data call dependency logic.

I'll see if I can reproduce the spinner thing later/next week.

PR #505, opened 2023-09-15
Proposal for graphics-centric minimized time slider

### LINKED ISSUE #514: Map - slim down time slider and address some usability issues.
**QA guidance**: please check general appearance and behaviour of the new time slider, ideally in qa.vectorbase.org map. Don't forget to apply "big" filters on the date variable and check behaviour (with and without an existing pink selection on the timeslider - this pink selection can be outside the "big" filter's range).

- [x] make smaller, pin to header, add some transparency
- [x] make show/hide explicit
- [x] make data request sensitive to subset
- [x] fix z-index issues (https://epvb.slack.com/archives/CMQ8K9XRS/p1695229428653309?thread_ts=1695224812.112409&cid=CMQ8K9XRS)

bobular:
Fixes #514 

This is a draft until we get UX approval.

I went further than anticipated and did the following

1. changed the background to match the header - though I see there's a hardcoded `=== 'VectorBase'` that I copied - do we not have a VectorBase theme? @asizemore was asking about this the other day
2. changed the layout 
    - variable name picker is the axis label in expanded mode
    - no more text to show the selected range, removed "choose a different date variable:"
    - timeline is slimmer when minimized
3. hiding behaviour is now fully manual
4. z-index changes
    - header (including EZ slider) is now only at level 2 - this allows draggable panels to be placed over the header (and EZ-slider). Before, it was possible to lose the handle of the draggable under the header or EZ slider.
    - left hand menu is at 10 - it is now not possible to cover the menu expand handle with a draggable.

Still do to
- [x] z-axis issues w.r.t. floaters (may wait until after refactor merge)
- [x] when switched off, the x-axis label is not greyed out.
- [x] rename components
- [x] remove subset-sensitivity of the timeslider x-axis (the black bars will still be subset-sensitive of course)
- [x] comment out fake controlled behaviour and investigate any "drift" that might occur from saved analyses (hopefully not)
- [x] add ticket for figuring out extending "little filters" throughout the app, e.g. the little filter from the marker variable config to the timeslider and floating viz.


bobular:
Video on Slack https://epvb.slack.com/archives/CMQ8K9XRS/p1694787354770699

If you look carefully, there's a bit of a clash of drop shadows (in the join between header and the slider) - not sure how to fix that.

bobular:
Ready for review. Open to styling/layout suggestions as always!

jernestmyers (commented):
I think this is looking good, but I have some questions/comments that come to mind that may serve as a start to a future UX discussion.

1. Should we have an info icon w/ tooltip as part of the on/off toggle that articulates what the actual component is doing? Or does this icon w/ tooltip need to be in a position that would render in the "unexpanded" state?
2. Because I'm clearly seeing something in the "unexpanded" state that first renders, my brain sort of overlooks the toggle as a way to expand the filter. Instead, my brain sees that toggle as a way to collapse the filter in spite of the arrow down (which should suggest to me that it expands). Perhaps the cognitive dissonance is specific to my feeble brain, though!
3. Nothing comes to mind as a fix for the drop shadow clashing, though it's not that severe either.

jernestmyers (approved):

bobular:
I merged the filter-sensitivity into this PR.

Then I found an edge case where if you make a selection in the time slider and then change the filters so that the old selection is outside the time range of the current filtered data, that selection lives "off screen" and is still applied as a little filter, giving no results and a 500 (see https://github.com/VEuPathDB/EdaDataService/issues/312) 

I thought it would be better to remove the selection/little filter if it has gone off-screen, so added this in the recent commit.

It uses a `useEffect` however, so would appreciate @dmfalke giving it the once-over.

dmfalke:
> I merged the filter-sensitivity into this PR.
> 
> Then I found an edge case where if you make a selection in the time slider and then change the filters so that the old selection is outside the time range of the current filtered data, that selection lives "off screen" and is still applied as a little filter, giving no results and a 500 (see [VEuPathDB/EdaDataService#312](https://github.com/VEuPathDB/EdaDataService/issues/312))
> 
> I thought it would be better to remove the selection/little filter if it has gone off-screen, so added this in the recent [commit](https://github.com/VEuPathDB/web-monorepo/pull/505/commits/9b7e137c3397e1366f5d17d367d7dac937dcde09).
> 
> It uses a `useEffect` however, so would appreciate @dmfalke giving it the once-over.

I'm not I like the UX here, since it might not be obvious that the time filter was removed. Instead, could we extend the range of the filter to include the selection? Or maybe we should always show the full range (at least for this phase)?

bobular:
Thanks @dmfalke - I have been working on this today and will have something available for testing soon.

I got into a bit of a rabbit hole with semantic zooming and making sure the histogram is robustly showing the correct bin start/ends etc.  Nearly done!

bobular:
Hi all - @jernestmyers @dmfalke @moontrip @chowington 

On Friday I did some major readjustments:

* The "key hack" to make the visx `Brush` behave like a controlled component wasn't actually working. When I fixed it, it introduced a secondary bug where the brush selection drifted wider with each handle adjustment. I made a very hacky fix (see the +2/-2 discussion). I don't think reverting to uncontrolled behaviour is a great idea (I'm pretty sure the drifting would still happen upon every fresh render of the object (e.g. when opening a saved analysis)).
* The timeslider visualisation only deals in binStart coordinates. It renders the final bin with a width of zero. Selecting/brushing the entire timeslider was NOT applying a little filter for the whole time range because the right-most date was the binStart of the final bin, not the bin end. Therefore in [this line](https://github.com/VEuPathDB/web-monorepo/pull/505/files#diff-1ed8dd704e444d668120d181f4f1d6c6fd7face32e7cfc4469a203c4f775c5b7R127) an extra "bin" is added as input to the slider component.
* The filter-aware (big filters directly on the date variable) behaviour needed to also [set the bin width resolution](https://github.com/VEuPathDB/web-monorepo/pull/505/files#diff-c97f01b67eb5e0321353d3d5fb9d4d8183b2e584239f2dbefe6d8ed27f8827fdR199) (year/month/week/day). 
* The black bars drawn on the timeslider needed to be different widths in February vs August (for example) - fixed
* The "auto zoom" (filter and subset sensitive at the moment) behaviour also takes into account the user's pink selection. This means that the selection doesn't go off-screen while still active.
* Structured, and hopefully improved, the tooltip text

bobular (commented):
Thanks to @dmfalke and @jernestmyers

bobular:
I removed the "controlled" nature of the Brush.

If you duplicate a browser tab with a pink date range, it will shift to the left a bit, but the underlying little filter does not change. I think this OK.

bobular:
This is ready for final review. Feedback from our meeting has been taken on board and sorted, thanks! :-) @dmfalke @jernestmyers

packages/libs/components/src/components/plotControls/TimeSlider.tsx
```diff
@@ -97,13 +100,17 @@ function TimeSlider(props: TimeSliderProps) {
...
+          // don't let range go outside the xAxisRange, if provided
+          start: xAxisRange
+            ? startDate < xAxisRange.start
+              ? xAxisRange.start
+              : startDate
+            : startDate,
+          end: xAxisRange
+            ? endDate > xAxisRange.end
+              ? xAxisRange.end
+              : endDate
```
bobular, line 112: Just spotted I need to fix up the dependencies below. Will do so.

bobular: Done :heavy_check_mark:

bobular:
With react-query's client-side caching we are a step closer to DIY timeline animations.
To demo this, expand the tab, make a time selection and toggle it on and off.

If we add the :arrow_left: and :arrow_right: buttons that move the selection left/right, that will eventually all get cached.

However, such controls will need re-enabling the fake controlled component behaviour.

jernestmyers:
If I have no filters and select the entire time slider, should my `View` count not also match the `All` and `Filtered` counts?
image

packages/libs/eda/src/lib/core/hooks/workspace.ts
```diff
@@ -160,7 +164,23 @@ export function useStudyEntities(filters?: Filter[]) {
...
+                      const binWidth = 1;
+                      const binUnits = (['year', 'month', 'week', 'day'].find(
+                        (unit) => {
+                          if (filterRange) {
+                            const diff = DateMath.diff(
+                              new Date(filterRange.min as string),
+                              new Date(filterRange.max as string),
+                              unit as DateMath.Unit
+                            );
+                            return diff >= 12;
```
dmfalke, line 179: How did you arrive at 12?

dmfalke (approved):
This seems good to me

dmfalke:
> If I have no filters and select the entire time slider, should my `View` count not also match the `All` and `Filtered` counts? image

Could this be explained by only a subset of markers having data for the time filter variable?

packages/libs/eda/src/lib/core/hooks/workspace.ts
bobular: I left home at about 11:30 !

bobular:
> Could this be explained by only a subset of markers having data for the time filter variable?

Yes, and/or lat/long.

packages/libs/eda/src/lib/core/hooks/workspace.ts
bobular: But it was more "if there are more than 12 days, use weeks. If there are more than 12 weeks, use months, and so on."

Maybe it's too small. If there are 13 months, do we really want year bins?  Perhaps 30 would be better.

bobular: Yeah, I'll go with 30. And leave a comment.

PR #508, opened 2023-09-18
turn on diff abund

asizemore:
Turns on the differential abundance app, which had been turned off for the release.


dmfalke (approved):
:rocket:

Issue #512, opened 2023-09-19
Correlation: Get visualization on page and show the bipartite network


asizemore:
See [here](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/12976129/Correlation+app) for more details.

Tasks

- [x] Create the formal bipartite network component #517 
- [x] add empty correlation endpoint to data service https://github.com/VEuPathDB/EdaDataService/pull/322
- [x] add the Correlation compute plugin #522 
- [x] add the bipartite network viz #522 
- [x] use fake data to generate a bipartite network when the viz is visited #522 
- [x] make sure thumbnail works #522

asizemore:
All done! Closing

asizemore:
Nothing to QA

PR #522, opened 2023-09-25
added correlation plugin

asizemore:
Resolves tasks in #512 

This PR:

- [x]  adds the Correlation compute plugin
- [x]  adds the bipartite network viz
- [x]  uses fake data to generate a bipartite network when the viz is visited
- [x]  adds a working thumbnail

Using the `add-correlation` of the data service and the rest of the services on their QA images, we can get the following: 

https://github.com/VEuPathDB/web-monorepo/assets/11710234/9a31e7b7-de01-49f1-ac27-122864e6dd9e

Turns out that though it's not so hard to generate fake data, it is a little more involved to sub in the requests and compute job status and such so that we can make the app appear and think that the Step 3 should be enabled. So I did a bit on the backend to help us at least just see the app (see `add-correlation` branch in the data service) but stopped there since we can still see the network enough to verify that it works!

If you do test with the local backend, do not try actually selecting the data in Step 1. The app will try to make a request to the backend but that part of the backend isn't set up yet. In the video, i demo that one _can_ use the drop down, but i don't actually select anything.

### DIFF
--- a/packages/libs/components/src/plots/BipartiteNetwork.tsx
+++ b/packages/libs/components/src/plots/BipartiteNetwork.tsx
@@ -7,10 +7,18 @@ import { partition } from 'lodash';
 import { LabelPosition, Link, NodeWithLabel } from './Network';
 import { Graph } from '@visx/network';
 import { Text } from '@visx/text';
-import { CSSProperties } from 'react';
+import {
+  CSSProperties,
+  Ref,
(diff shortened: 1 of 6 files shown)


jernestmyers (approved):
LGTM. I only played with the updated stories, but between the code and the screencast, things are looking good! 💪

PR #523, opened 2023-09-25
Tweak UI for new SRT

jernestmyers:
Addresses the following points [from this issue](https://github.com/VEuPathDB/EbrcWebsiteCommon/issues/193):
> 1. Mockup updated after UX meeting Jan 18 2023 https://docs.google.com/presentation/d/1m43IHO6BnkXHpusYR2cnJgWdgazm8jCy1euqBiNyKRA/edit#slide=id.p
> 2. Add a checkbox as in the tabular reporter to select one transcript per gene to all report forms where this is now supported. See
https://github.com/VEuPathDB/ApiCommonWebsite/issues/117

Below is a screenshot of the UI changes [modeled off of this mockup](https://docs.google.com/presentation/d/1m43IHO6BnkXHpusYR2cnJgWdgazm8jCy1euqBiNyKRA/edit#slide=id.p).

Note that there's a comment in the code that wonders if the flattening of the defline options, which requires adding `gene_id` to the `deflineFields` array, would cause trouble with the backend. It does not appear to break anything. I tested my branch with a local dev server pointed at @ryanrdoherty's dev site and the same reports succeed and fail when using this branch versus `main`.

image

### DIFF
--- a/packages/libs/web-common/src/components/reporters/BedAndSequenceGenomicSequenceReporterForms.jsx
+++ b/packages/libs/web-common/src/components/reporters/BedAndSequenceGenomicSequenceReporterForms.jsx
@@ -41,7 +41,7 @@ let SequenceRegionRange = (props) => {
         end={10000}
         step={1}
         onChange={getUpdateHandler(offset)}
-        size="6"
+        size={6}
       />
       nucleotides
     </React.Fragment>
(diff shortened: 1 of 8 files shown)


dmfalke (approved):
This looks great!

aurreco-uga:
@jernestmyers could you remove the 1 transcript option in record types other than genes?
Screen Shot 2023-11-15 at 9 04 29 AM

PR #526, opened 2023-09-29
Maaslin method for diff abund

d-callan:
this turned into more than i thought it would be... the idea was to add a new input for the diff abund compute, to support choosing a method explicitly. backend that meant what i thought was a minor api change, which was something less minor here than i had assumed. so ive gone and basically added more explicit support for variable collections as well.

Still need to make the new method input prettier and have the available method differ by the collection chosen. im considering leaving the available methods unconstrained for the time being, until some backend work is done to better support that.

### DIFF
--- a/packages/libs/components/src/plots/VolcanoPlot.tsx
+++ b/packages/libs/components/src/plots/VolcanoPlot.tsx
@@ -8,6 +8,7 @@ import {
 import {
   VolcanoPlotData,
   VolcanoPlotDataPoint,
+  VolcanoPlotStats,
 } from '../types/plots/volcanoplot';
 import { NumberRange } from '../types/general';
 import { SignificanceColors, significanceColors } from '../types/plots';
@@ -46,13 +47,13 @@ export interface RawDataMinMaxValues {
 }
 
 export interface VolcanoPlotProps {
-  /** Data for the plot. An array of VolcanoPlotDataPoints */
+  /** Data for the plot. An effectSizeLabel and an array of VolcanoPlotDataPoints */
   data: VolcanoPlotData | undefined;
   /**
    * Used to set the fold change thresholds. Will
    * set two thresholds at +/- this number. Affects point colors
    */
-  log2FoldChangeThreshold: number;
+  effectSizeThreshold: number;
   /** Set the threshold for significance. Affects point colors */
   significanceThreshold: number;
   /** x-axis range  */
@@ -83,9 +84,14 @@ export interface VolcanoPlotProps {
   minPValueCap?: number;
 }
 
-const EmptyVolcanoPlotData: VolcanoPlotData = [
-  { log2foldChange: '0', pValue: '1' },
+const EmptyVolcanoPlotStats: VolcanoPlotStats = [
+  { effectSize: '0', pValue: '1' },
 ];
+
+const EmptyVolcanoPlotData: VolcanoPlotData = {
+  effectSizeLabel: 'log2(FoldChange)',
+  statistics: EmptyVolcanoPlotStats,
+};
 
 const MARGIN_DEFAULT = 50;
 
@@ -130,7 +136,7 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
     independentAxisRange,
     dependentAxisRange,
     significanceThreshold,
-    log2FoldChangeThreshold,
+    effectSizeThreshold,
     markerBodyOpacity,
     containerClass = 'web-components-plot',
     containerStyles = { width: '100%', height: DEFAULT_CONTAINER_HEIGHT },
@@ -154,6 +160,8 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
     }),
     []
   );
+
+  const effectSizeLabel = data?.effectSizeLabel;
 
   // Set maxes and mins of the data itself from rawDataMinMaxValues prop
   const { min: dataXMin, max: dataXMax } = rawDataMinMaxValues.x;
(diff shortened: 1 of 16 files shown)


d-callan:
oh. i also need to figure yet what to do about saved analyses and this api change breaking them.

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -192,27 +209,28 @@ export function DifferentialAbundanceConfiguration(
   const collectionVarItems = useMemo(() => {
     return collections
       .filter((collectionVar) => {
-        return collectionVar.normalizationMethod
-          ? !collectionVar.isProportion &&
-              collectionVar.normalizationMethod === 'NULL' &&
-              !collectionVar.displayName?.includes('pathway')
+        return collectionVar.normalizationMethod // i guess diy stuff doesnt have this prop?
```
asizemore, line 212: It looks like the newly installed ones do, but just in case someone tries from an old analysis or something

asizemore:
If the collection work is in separate commits from the volcano work, it'd be helpful to break them into separate PRs. One i can't think about two things at once lol and two Dave is off this week and will be able to provide better feedback on the collection changes than i can. 

For the maaslin changes, i didn't test them just read through. In general i think they look good, but we should have the visualization split the data from the effectSizeLabel. That way the volcano component is getting just general volcano data in the `data` prop, and the visualization handles the specifics of our backend. Specifically, i think `VolcanoPlot` could gain an `effectSizeLabel` prop, and then `VolcanoPlotData` can just be the array of points again. In the visualization, we can snatch the effectSizeLabel off of the response and not worry about it during the data cleaning or anything.

d-callan:
when i first started looking at this things spiraled pretty quickly, and im not sure the collections stuff actually is in separate commits or trivial to separate at this point. thats on me :( but the collections stuff is also required to test the volcano changes, and i dont really want to wait a week for that if im real honest.. maybe @jernestmyers could take a look if hes around?

i have mixed feelings about splitting the label for the stats from the data, if im honest. i get where youre coming from............... i do. but idk, i dont even like that the adjusted p values dont have explicit labels attached to them.. theres more than one way to adjust p values. it doesnt feel like a coincidence or a technical detail of how the backend does its work to me.

d-callan:
so i dont think id be ok w a volcano plot that wasnt explicit about what the stats actually represented. if i saw 'effect size' as an axis label, id not trust that without knowing more. though i recognize im maybe an outlier. as for where the label comes from and whether it depends on context.. idk to me its the choice of deseq that determines the label, not a genomics vs mbio context. i can appreciate that mbio folks might like 'Fold Change' and genomics might like 'FC' or something, but the statistic from deseq will always be fold change regardless and im uncomfortable logically separating that label from the values. its too easy to write 'Fold Change' into some viz, then add some method where thats not accurate or appropriate, and miss it. its the statistics + its label that constitutes data.

packages/libs/components/src/plots/VolcanoPlot.tsx
```diff
@@ -155,6 +161,8 @@ function VolcanoPlot(props: VolcanoPlotProps, ref: Ref<HTMLDivElement>) {
     []
   );
 
+  const effectSizeLabel = data?.effectSizeLabel;
```
jernestmyers, line 164: I think we can remove the question mark here since we're passing in a default value to the `data` prop that defines the `effectSizeLabel` property.

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -599,11 +606,11 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
       <LabelledGroup label="Threshold lines" alignChildrenHorizontally={true}>
         <NumberInput
           onValueChange={(newValue?: NumberOrDate) =>
-            updateVizConfig({ log2FoldChangeThreshold: Number(newValue) })
+            updateVizConfig({ effectSizeThreshold: Number(newValue) })
           }
           label="log2(Fold Change)"
```
jernestmyers, line 611: Should this label be based on the data's `effectSizeLabel` instead?

packages/libs/eda/src/lib/core/types/variable.ts
```diff
@@ -16,3 +16,31 @@ export const StringVariableValue = t.intersection([
...
+  entityId: t.string,
+  collectionId: t.string,
+});
+
+export type VariableCollectionDescriptor = t.TypeOf<
+  typeof VariableCollectionDescriptor
+>;
+export const VariableCollectionDescriptor = _VariableCollectionBase;
+
+export function isVariableDescriptor(
```
jernestmyers, line 30: Would this function and `isVariableCollectionDescriptor` be better suited in`utils/study-metadata`?

jernestmyers (commented):
Just a few nits/comments/questions from me for now. I don't have much to offer in terms of the discussion you two are already having regarding the data and labels. Once that discussion and the UI is settled, I'm happy to take another look (and/or handle the UI tweaks) as needed/desired.

asizemore:
>  if i saw 'effect size' as an axis label, id not trust that without knowing more

I agree! But the visualization plugin is the context that gives us the information. If the plot said "effect size" as the label but the whole page was talking about the log fold change as the effect size, then it'd be clear, no? 

> i can appreciate that mbio folks might like 'Fold Change' and genomics might like 'FC' or something, but the 
statistic from deseq will always be fold change regardless

Right exactly i agree! But we shouldn't put all the genomics and mbio logic into `VolcanoPlot`, that all is supposed to live in `VolcanoPlotVisualization`. So then we could have the effectSize as part of `data` like in the PR and have an optional prop that gives a label for the effect size if we don't want to use whatever comes with the data. Another option is to have the data be just an array of data points and _require_ an effect size label prop. That way anyone using the component has to think about it at least. 

Both solve the problem, but i'm in favor of the latter because it makes the component easier to use when the inputs are more basic objects, and also easier to refactor if we wanted to change out the volcano component in the future (we have this problem with scatter where there's a ton of processing of the data to make it perfect for plotly but that causes a headache). @d-callan would this approach quell your concerns?

d-callan:
im not willing to assume that these plots will always be used in the context of some eda viz plugin and believe that what the plot components provide should be able to stand on their own. im not too worried about differences between mbio and genomics, bc i think theyll be minor things that we could resolve as they come. my point there was more to do w the fact that mbio and genomics actually CANT choose completely different labels, bc the label is inherently tied to the statistic/ analytical method providing the statistic. i know its a slightly involved shape for the data now i guess, but its not bc of some specific lib were using, its bc of the data themselves. if we made a required additional prop for effectSizeLabel, anybody could pass anything as a label, easily and even accidentally..

d-callan:
id like to merge this, as well as the backend changes to compute, rserve and data, asap. i dont like having that many threads lingering. this is now functional in the sense it does what i set out to make it do. we still need:

asizemore:
Sounds good to me. Also happy to help if needed.

PR #534, opened 2023-10-04
testing multiple markers selection with SAM

moontrip:
This PR intends to test multiple markers selection with SAM, thus branched out from https://github.com/VEuPathDB/web-monorepo/tree/431-multi-select-marker. It seems to work just fine.

I made this as a draft PR as I was not sure when this would be required and possible changes in the parent PR, https://github.com/VEuPathDB/web-monorepo/pull/513, that may affect this work. You can test it if you are interested, @bobular 😄


moontrip:
Please check the following comment for new update: https://github.com/VEuPathDB/web-monorepo/pull/581#issue-1971625909.

moontrip:
Since there are lots of changes made in https://github.com/VEuPathDB/web-monorepo/pull/588, which was branched out from this PR, we will use that PR instead this, thus closing this PR.

Issue #99, opened 2023-04-21
Allow user to collapse floater controls


steve-fischer-200:
The floater controls are big.  User wants to see the floaters over the map, without additional stuff.

Allow them to collapse the controls

adnauseum:
Mockups to prompt discussion: https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?node-id=991-14514&t=jQrnhVfZHQn6qlAk-4

adnauseum:
From my cursory glance, it looks like all the visualization components (BarplotVisualization.tsx, BoxplotVisualization.tsx, HistogramVisualization.tsx, LineplotVisualization.tsx, etc) could be implemented with some `<VisualizationControls>` component that takes a prop to toggle hiding.

d-callan:
id like to know how this relates, if at all, to us adding axes controls for things like binning histograms and setting axes ranges

chowington:
Fixed by https://github.com/VEuPathDB/web-monorepo/pull/535

PR #535, opened 2023-10-04
Collapse button for floating viz controls

chowington:
Closes https://github.com/VEuPathDB/web-monorepo/issues/99

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/DraggableVisualization.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/DraggableVisualization.tsx
@@ -86,18 +86,14 @@ export default function DraggableVisualization({
     >
       <div
         style={{
-          // Initial height & width.
-          height: 547,
-          width: 779,
-          // This prevents the panel from collapsing aburdly.
-          minWidth: 400,
-          minHeight: 200,
+          position: 'relative',
+          width: '100%',
+          height: '100%',
         }}
       >
         <div
           style={{
             position: 'absolute',
-            top: 0,
             right: 0,
             zIndex: 100,
             padding: '0.5rem',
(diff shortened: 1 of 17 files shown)


packages/libs/eda/src/lib/map/analysis/hooks/standaloneVizPlugins.ts
```diff
@@ -31,16 +31,18 @@ import { scatterplotRequest } from './plugins/scatterplot';
 import TimeSeriesSVG from '../../../core/components/visualizations/implementations/selectorIcons/TimeSeriesSVG';
 import _ from 'lodash';
 
+type StandaloneVizOptions = LayoutOptions & OverlayOptions;
+
 interface Props {
   selectedOverlayConfig?: OverlayConfig | BubbleOverlayConfig;
   overlayHelp?: ReactNode;
+  vizOptions?: StandaloneVizOptions;
```
bobular, line 39: Was this prop planned and never used?

packages/libs/eda/src/lib/map/analysis/DraggableVisualization.tsx
```diff
@@ -71,7 +76,11 @@ export default function DraggableVisualization({
       confineToParentContainer
       showPanelTitle
       isOpen
-      styleOverrides={{ zIndex: zIndexForStackingContext, resize: 'both' }}
+      styleOverrides={{
+        zIndex: zIndexForStackingContext,
+        resize: 'both',
+        overflow: 'hidden',
```
bobular, line 82: tbh I don't fully understand the `overflow` property usage here.
It may be a browser difference (though I've tried Chrome and Firefox).
I am seeing scroll bars appear on mouse-over. Perhaps you could explain to me later? Sorry!

bobular (changes requested):
Looks great. A few small comments in the code.

If it's trivial, without massively reengineering coreui, would you be able to make the new button a bit smaller? Ideally not larger than the `Add filters` button in the header, and maybe not all caps also. (this is the only firm "requested change" - others are for discussion)

There is also a more general problem of our plots (EDA and SAM) all wasting space at the top where a header can go (IIRC, we add a header when faceting in regular SAM).

We should talk about a follow-on ticket that considers the window sizing.

I've added a Map UX topic ([item 3 here](https://veupathdb.atlassian.net/wiki/spaces/WG/pages/391970817/2023-10-09+Map+UX+Monday+-+provisional)) to decide what's best for state management. Window size and location could perhaps be part of that discussion.


image

packages/libs/eda/src/lib/map/analysis/DraggableVisualization.tsx
chowington: We want it to scroll, but we don't want `DraggablePanel` to handle the scrolling; we want to handle it ourselves in the divs we pass as children of `DraggablePanel`. Thus you'll see `overflow: 'auto'` on one of the child divs below.

packages/libs/eda/src/lib/map/analysis/hooks/standaloneVizPlugins.ts
chowington: Yep! Good catch 👍

bobular (approved):
Excellent - thanks. I'll draft the next ticket and/or make mockups for the next phase regarding the draggable's sizing.

PR #542, opened 2023-10-09
Timeslider help icon and further slimming

bobular:
Note merge into parent branch

image

### DIFF
--- a/packages/libs/components/src/components/plotControls/EzTimeFilter.tsx
+++ b/packages/libs/components/src/components/plotControls/EzTimeFilter.tsx
@@ -115,8 +115,8 @@ function EzTimeFilter(props: EzTimeFilterProps) {
 
   // bounds
   const xBrushMax = Math.max(width - margin.left - margin.right, 0);
-  // take 80 % of given height considering axis tick/tick labels at the bottom
-  const yBrushMax = Math.max(0.8 * height - margin.top - margin.bottom, 0);
+  // take 70 % of given height considering axis tick/tick labels at the bottom
+  const yBrushMax = Math.max(0.7 * height - margin.top - margin.bottom, 0);
 
   // scaling
   const xBrushScale = useMemo(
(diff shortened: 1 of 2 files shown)


jernestmyers (commented):
Yikes, this is a tough UX/UI to work out.

I think I'd prefer the help icon to be with the on/off toggle, but then that dramatically affects the layout. I like its size in this iteration, though I wonder if folks are okay with the time filter not showing the selected variable?

bobular:
> Yikes, this is a tough UX/UI to work out.
> 
> I think I'd prefer the help icon to be with the on/off toggle, but then that dramatically affects the layout. I like its size in this iteration, though I wonder if folks are okay with the time filter not showing the selected variable?

A textless on/off toggle could go in the minimized version perhaps?

jernestmyers:
> > Yikes, this is a tough UX/UI to work out.
> > I think I'd prefer the help icon to be with the on/off toggle, but then that dramatically affects the layout. I like its size in this iteration, though I wonder if folks are okay with the time filter not showing the selected variable?
> 
> A textless on/off toggle could go in the minimized version perhaps?

If so, we'd maybe want the toggle's status at the top of the tooltip. I wonder how much wider the component gets with the toggle and tooltip next to the slider. 🤔

bobular:
Why not "textless"/no status label?

Otherwise it gets very wide

bobular:
I'm thinking that the on/off toggle is not core functionality - it's for the specific case where you want to toggle between time-sliced and not time-sliced. I think it's OK to live only in the expanded version?

bobular:
Why can I merge this without an approval?

jernestmyers:
> Why can I merge this without an approval?

I'm guessing because you're requesting to merge this into a branch that isn't `main`.

bobular:
> > Why can I merge this without an approval?
> 
> I'm guessing because you're requesting to merge this into a branch that isn't `main`.

Oh duh...
Nice work, Columbo!

PR #543, opened 2023-10-09
Use Maaslin only and report method in plot

### LINKED ISSUE #541: volcano: remove Method and instead report method used in plot
Discussed in 10/04/23 mbio meeting

Dan B suggested that users would get confused selecting their method, so instead he'd like us to select the method for them. Then we'd report the method used in the plot below. Ranked abundance shows some similar information in the plot subtitle, so we could probably just use that strategy here as well.

For later (not this ticket!):
There are other cases where we've discussed giving users some feedback about the computation (for example, ranked abundance). So i think it would be worth making some mockups of where this type of information might go. Probably up with the Generate Results button... but that's for later debate :)

asizemore:
Resolves #541 

Discussed in the mbio meeting, @dpbisme raised concerns about the method selection becoming too complicated for users. Instead he suggested choosing the method based on the data, so that the user doesn't have to choose. 
UPDATE: New decision is to only use the 'Maaslin' method for this first roll out. Code has been updated so as to only use Maaslin as the differential abundance method.

This PR 
1. Removes the Method input selector
2. Sets the `differentialAbundanceMethod` to always be `Maaslin`
3. Reports the method used in the volcano plot subtitle.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
@@ -206,11 +206,13 @@ export function DifferentialAbundanceConfiguration(
     );
 
   const collectionVarItems = useMemo(() => {
+    // Show all collections except for absolute abundance. Eventually this will be performed by
+    // the backend, similar to how we do visualization input var constraints.
     return collections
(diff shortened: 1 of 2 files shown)


jernestmyers (approved):
LGTM!

PR #545, opened 2023-10-10
replace variableId with collectionId

asizemore:
When the api changed last week, it replaced `variableId` with `collectionId` for collection variables. We missed one spot in the description component for each app where we should have also made the swap. This eventually made the component think there was no collection variable selected.

Before (qa):
image

After (this PR):
Screen Shot 2023-10-10 at 2 15 16 PM


dmfalke (approved):
Looks good!

PR #549, opened 2023-10-12
Wire up VDI client

jernestmyers:
This has gotten to be a rather large chunk of work, so perhaps it's best that I pause here for now and get a review.


packages/libs/user-datasets/src/lib/Service/api.ts
```diff
@@ -0,0 +1,230 @@
...
+import { submitAsForm } from '@veupathdb/wdk-client/lib/Utils/FormSubmitter';
+
+export const VDI_SERVICE_BASE_URL = 'http://localhost:8080';
+const VDI_SERVICE = '/vdi-datasets';
+
+const userIdsByEmailDecoder = record({
+  results: arrayOf(objectOf(number)),
+});
+
+export class UserDatasetApi extends FetchClientWithCredentials {
```
dmfalke, line 37: I see that you are using class field functions, rather than class methods. What's the reason for this? Maybe add a comment explaining why.

packages/libs/user-datasets/src/lib/Service/index.ts
```diff
@@ -1,29 +1,61 @@
...
-          ...userDatasetsServiceWrappers,
-          ...makeUserDatasetUploadServiceWrappers(serviceConfig),
-        };
+  const vdiService = new UserDatasetApi(
+    { baseUrl: VDI_SERVICE_BASE_URL },
+    wdkService
+  );
 
   return {
+    [vdiCompatibilityFlag]: null,
```
dmfalke, line 26: How about setting this to `vdiCompatibilityFlag` or `true`, just so it isn't a "falsey" value?

packages/libs/user-datasets/src/lib/StoreModules/UserDatasetListStoreModule.ts
```diff
@@ -84,6 +84,7 @@ export function reduce(state: State = initialState, action: Action): State {
           };
 
     case DETAIL_UPDATE_SUCCESS:
+      // @ts-ignore
```
dmfalke, line 87: I think the code compiles without this.

packages/libs/user-datasets/src/lib/StoreModules/UserDatasetListStoreModule.ts
```diff
@@ -98,6 +99,7 @@ export function reduce(state: State = initialState, action: Action): State {
         : state;
 
     case DETAIL_REMOVE_SUCCESS:
+      // @ts-ignore
```
dmfalke, line 102: I tried to remove this one, and get this error:
```
Type '{ userDatasets: (string | number)[]; userDatasetsById: { [x: string]: { isLoading: false; resource: UserDataset; } | undefined; }; status: "complete"; filterByProject: boolean; }' is not assignable to type 'State'.
  Types of property 'userDatasetsById' are incompatible.
    Type '{ [x: string]: { isLoading: false; resource: UserDataset; } | undefined; }' is not assignable to type 'Record<string, { isLoading: false; resource: UserDataset; }>'.
      'string' index signatures are incompatible.
        Type '{ isLoading: false; resource: UserDataset; } | undefined' is not assignable to type '{ isLoading: false; resource: UserDataset; }'.
          Type 'undefined' is not assignable to type '{ isLoading: false; resource: UserDataset; }'.ts(2322)
```

I think it is because we are setting the value of the removed dataset to `undefined`.

I think this would work, using `omit` from `lodash`:
```ts
      return state.status === 'complete'
        ? {
            ...state,
            userDatasets: difference(state.userDatasets, [
              action.payload.userDataset.id,
            ]),
            userDatasetsById: omit(state.userDatasetsById, action.payload.userDataset.id)
          }
        : state;
```

dmfalke (changes requested):
This is looking great. I've made a handful of suggestions. Let me know if you want to discuss anything in detail.

packages/libs/user-datasets/src/lib/Components/List/UserDatasetList.tsx
```diff
@@ -49,12 +49,12 @@ interface Props {
...
   filterByProject: boolean;
   shareUserDatasets: (
-    userDatasetIds: number[],
+    userDatasetIds: string[],
     recipientUserIds: number[]
   ) => any;
   unshareUserDatasets: (
-    userDatasetIds: number[],
-    recipientUserIds: number[]
+    userDatasetIds: string,
```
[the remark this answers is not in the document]
jernestmyers: For posterity since we already discussed this in another venue, our UI currently doesn't allow for us to batch together multiple unshare requests. Thus, this parameter should be renamed to `userDatasetId` to better reflect this limitation. We could also rename the method itself to be `unshareUserDataset`, but I suspect we'll be able to batch together multiple unshare requests in a future iteration so I propose we only change the parameter names in the interim.

packages/libs/user-datasets/src/lib/Components/Sharing/UserDatasetSharingReducer.ts
```diff
@@ -21,8 +21,8 @@ type State = Record<
 
 const initialState: State = {};
 
-const handleAdd = handleOperation('add');
-const handleDelete = handleOperation('delete');
+const handleAdd = handleOperation('grant');
+const handleDelete = handleOperation('revoke');
```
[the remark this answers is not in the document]
jernestmyers: In terms of sharing, the only actions we define in the current VDI docs are `grant` and `revoke`. I took this to mean they replace `add` and `delete`. And the initial phase automagically accepts the share for the recipient, as opposed to needing a new UI for users to `accept` or `reject` (the VDI actions) the share requests.

packages/libs/user-datasets/src/lib/Controllers/UserDatasetNewUploadController.tsx
```diff
@@ -14,12 +14,12 @@ import {
 
 import UploadForm, { FormSubmission } from '../Components/UploadForm';
 
-import { assertIsUserDatasetUploadCompatibleWdkService } from '../Service/UserDatasetUploadWrappers';
-
 import { StateSlice } from '../StoreModules/types';
 
 import { DatasetUploadTypeConfigEntry } from '../Utils/types';
 
+const SUPPORTED_FILE_UPLOAD_TYPES = ['csv', 'gz', 'tgz', 'tsv', 'txt', 'zip'];
```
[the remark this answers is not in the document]
jernestmyers: That's a question I've wondered myself and haven't yet asked. I'll add that to my list of questions.

packages/libs/user-datasets/src/lib/Service/api.ts
jernestmyers: Hmm, it wasn't a deliberate choice- more of a reflex. I don't suspect we'll have many instances of this API floating around, so it's not a huge issue for the functions to not be on the prototype. But it's probably worth switching them to class methods since I can't think of a need for them to be field functions. Am I missing something?

dmfalke (commented):
I looked at your latest changes, and they look good. I see some `var`s, which could be replaced with `const`s.

packages/libs/user-datasets/src/lib/Components/Detail/UserDatasetDetail.jsx
```diff
@@ -552,11 +501,81 @@ class UserDatasetDetail extends React.Component {
...
+     * indicates a successful install.
+     *
+     * We know a dataset is incompatible when the site-specific's install status
+     * indicates `missing-dependency`
+     */
+    const installStatusForCurrentProject = status.install?.find(
+      (d) => d.projectId === projectId
+    );
+
+    const isInstallable = projects.includes(projectId);
```
dmfalke, line 526: Maybe we can change this variable name to something like `isTargettingCurrentSite`? `isInstallable` could be interpreted to mean that there are errors preventing installation.

aurreco-uga (approved):

dmfalke (approved):

PR #560, opened 2023-10-20
Feature 556 connect correlation backend

### LINKED ISSUE #556: Bipartite network: wire up backend
Final part of Step 1. When the backend for the correlation app is complete, remove the fake data from the bipartite network and use the real data!

Requires correlation branch of all backend services.

In progress. Latest screenshot:
image

asizemore:
Resolves #556 

This PR adds the `correlationAssayMetadata` plugin and the `BipartiteNetworkVisualization`, as well as necessary upgrades to get that all working. 

Some things to note:
- I had a bit of trouble managing the sizing of the plot. In some ways it should adapt to the container but in others it shouldn't.... check out what i've done and please let me know if it's nonsensical
- I intentionally haven't constrained any collections at this point. There's still a discussion about how to handle the different types of data so for now it's just open. This is also why there's no input for method selection at this point either. It all just defaults to "spearman".
- Handling link colors took a bit of thinking, so let me know if what i've written isn't clear.
- I made up a way to have a "blank" plot to show while there is no data. A blank plot for a network is a little funk since we have no axes, so i just put a few gray nodes around to suggest something might be there soon.

Left for later
- truncating long labels. That should be a bpnet component feature, and doesn't deal with the plugin at all i wouldn't think. Use `truncateWithEllipsis` also add the full label on hover.
- handle when an empty network is returned.
- Adding an input for changing the correlation coefficient threhsold (but the structure is there)
- A real viz picker icon for the bipartite network

Currently working with the qa backend!

https://github.com/VEuPathDB/web-monorepo/assets/11710234/56bdaa6b-12eb-4fdd-b693-be8ad3b48f32


packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
```diff
@@ -64,32 +76,110 @@ interface Options
...
+    useCallback(async (): Promise<BipartiteNetworkResponse | undefined> => {
+      // Only need to check compute job status and filter status, since there are no
+      // viz input variables.
+      console.log(filteredCounts);
+      if (computeJobStatus !== 'complete') return undefined;
+      if (filteredCounts.pending || filteredCounts.value == null)
+        return undefined;
+
+      // There are _no_ viz request params for the volcano plot (config: {}).
+      // The data service streams the volcano data directly from the compute service.
```
d-callan, line 107: another reference to volcano here. this time though i think my larger point is that for the network i think we said itd end up being the data service job to do the pruning of edges by pvalue and correlation coefficient thresholds. (i dont actually remember the compute or data service prs well enough to remember if thats what we actually ended up doing. but think we should. it was a good plan.)

d-callan: anyway, just trying to say there will be viz plugin params, eventually, if not yet.

dmfalke:
I just viewed the screen cast. Here are some initial thoughts (I will review the code next):

- The empty plot looks good in the viz, but I'm concerned about it being used as a thumbnail in the listing page. I don't have any suggestions for that, at the moment.
- The scrolling is okay, but it would be really cool if you could scroll each column, independently. I have no idea if that's possible.

dmfalke (commented):
I've reviewed the code, and it looks great. I have one small suggestion, which you can find below.

When I was testing the new viz, I noticed something odd. A thumbnail is being generated constantly, even without interacting with the plot.

image

It's not clear what is causing this. Let me know if you want help figuring it out.

I'm also finding the text in the plot a bit hard to read. Maybe it's my browser (Firefox)?

image

Finally, I didn't see any tooltips. Is that intentional?

I tested using the DailyBaby study.

Great work!

packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
```diff
@@ -58,47 +65,163 @@ export const BipartiteNetworkConfig = t.partial({
...
+      links: data.value.bipartitenetwork.data.links.map((link) => {
+        return {
+          source: link.source,
+          target: link.target,
+          strokeWidth: Number(link.strokeWidth),
+          color: link.color ? linkColorScale(link.color.toString()) : '#000000',
+        };
+      }),
+    };
+  }, [data, entities, linkColorScale]);
```
jernestmyers, line 191: I think `linkColorScale` is causing the infinite loop with the thumbnail and patch request. All of the logic for `linkColorScale` is outside of the `useMemo` and new renders aren't guaranteed to be referentially equal. We can either move this logic inside of this `useMemo` or upset ESLint by removing `linkColorScale` from the dependency array.

Also, I suggest we specify `data.value` instead of just `data` in this array while we're at it.

asizemore:
Thank you both!! I'm sorry i wasn't able to get to this today after you both were so fast with your feedback, but i'll focus on the updates first thing tomorrow morning :)

asizemore:
@jernestmyers you were right! The colorscale was the issue! Thanks for the find!

@dmfalke responses to your questions below!

> The empty plot looks good in the viz, but I'm concerned about it being used as a thumbnail in the listing page. I don't have any suggestions for that, at the moment.

This is a great point, i hadn't thought about that. A user will see this only if the computation has finished but they haven't opened the viz yet (the same scenario for diff abund is shown above just for comparison). The little gray network isn't totally ridiculous looking, which was my big fear when you mentioned it, but it also doesn't really communicate "hey click on me i'm an empty plot". I also don't have a better way to solve that yet, other than just "Empty network" text? It might need some mockups and discussion.


Screen Shot 2023-11-17 at 5 58 23 AM


> Finally, I didn't see any tooltips. Is that intentional?

Yes, it was intentional. As i was going i was looking for ways to keep this PR reasonably sized and since the scope of tooltips is just the component i felt like it could get cut from this pr that focuses on the viz + app plugin. Happy to add them if that would be preferable!


> The scrolling is okay, but it would be really cool if you could scroll each column, independently. I have no idea if that's possible.

That would be super cool!!! I don't know either but something I'm on the lookout for. During the design phase, we had assumed that independent column scrolling would be hard, so we're hoping that implementing user-filters on nodes and edges will be sufficient. But we'll see... 😄 



> I'm also finding the text in the plot a bit hard to read. Maybe it's my browser (Firefox)?

The glyphs were pretty thin so I increased the font weight of all the text (see below). Is that a little easier to read? I could also increase the font size, too! Let me know what you think. 

Screen Shot 2023-11-17 at 6 27 23 AM

jernestmyers (approved):
These changes LGTM. Not seeing anymore wayward requests and the text is certainly easier to see!

dmfalke:
These updates look great. Very nice!

PR #564, opened 2023-10-24
Polishing updates for volcano plot

### LINKED ISSUE #558: Volcano: update legend text and pointer
**Problem:** 
Text in the legend is unclear. Also the cursor switches to the hand when over the legend which suggests it's interactive but the legend isn't interactive. (from slack)

**Suggested solution:**
Change the legend text to "Up in X" and "Down in X". Also keep the mouse the usual pointer while over the legend to prevent the user from thinking it's interactive.

asizemore:
Resolves #558 

This PR improves the volcano plot styling and functionality in the following ways:
- Adds the `isClickable` prop to legends in order to control the cursor, preventing the legend from seeming clickable when it isn't but still allowing for the tooltip. Nothing happens with this prop on the bubble legend but i added it anyways for consistency?
- Updates the volcano legend text based on feedback from Dan B
- Added more space for the x axis annotations
- Truncated the x axis annotations with a new `truncateWithEllipsis` util.

In addition, this PR
- fixes the Empty plot volcano story 
- adds a new story that displays the volcano inside of our plot container. Helpful for checking out how the plot margins will look on the site.
- deletes the relaxed microbe pic because we didn't end up using it for the compute service (leftover from last year!)

Testing:
- i tested a boxplot with long tick labels to make sure i didn't break tick labels with the new `truncateWithEllipsis` function, but feel free to test for yourself! 
- if you see anything else that looks iffy in terms of style or spacing in the volcano plot please let me know!


packages/libs/eda/src/lib/core/components/visualizations/relaxMicrobe.tsx
bobular, line 1: 😢

packages/libs/components/src/components/plotControls/PlotLegend.tsx
```diff
@@ -8,6 +8,7 @@ import PlotBubbleLegend, { PlotLegendBubbleProps } from './PlotBubbleLegend';
 
 interface PlotLegendBaseProps extends ContainerStylesAddon {
   legendTitle?: string;
+  isClickable?: boolean;
```
bobular, line 11: I think remove this (explanation in final comments)

bobular (changes requested):
Sorry for the late review!

I think you only need the `isClickable` configuration prop for the "contents" of the legend not the title. Yes we show a pointy cursor, but that is to encourage the user to wait for the tooltip, not for them to click on it. So I think remove the bits that change the cursor for the title and the legend container.

`PlotLegend` handles the title, while the `PlotXXXLegend` components handle the contents. All the `PlotXXXLegend` components have the `isClickable` prop, so you can still pass this from `VolcanoPlotVisualization` and have `PlotListLegend` respond to it.

Let's check that the mouse cursor behaviour of regular visualizations (e.g. scatterplot with both types of overlay) are the same as qa before merging.

asizemore:
Update: decided to give pointer cursor only to items when they have the checkbox!

Also looping in @moontrip who i should have looped in from the start!

bobular (commented):
Hi @asizemore 

I notice that for a vanilla visualization with a categorical legend, the whole legend panel has the :point_up_2: mouse cursor, including the title, which is not clickable.

I think the better solution would be to remove the `isClickable` prop from `PlotLegendBaseProps` and the `PlotLegend` component as a whole. In fact, remove `isClickable` from all the `PlotXXXLegend` components.  `PlotGradientLegend` and `PlotBubbleLegend` can hard-code `cursor: 'default'`.

Then in `PlotListLegend` the cursor attribute can be set something like this

`cursor: showCheckBox && item.hasData ? 'pointer' : 'default'`

asizemore:
@bobular Ready for another look!

The latest commit
- removes `isClickable`
- prevents checkbox items with no data (disabled checkbox items) from having the pointer and instead gives them the default cursor.

I checked the volcano legend and the vanilla scatterplot gradient and categorical legend - both functioned as expected. 

Let me know what you think!

bobular (approved):
Looks great. Confirmed functionality for mixed data/no-data checkbox legend :-)

PR #565, opened 2023-10-24
Incorporate p value floor from backend

### LINKED ISSUE #528: Volcano: empty plot sometimes because of infinity points
From slack, Dan B found that this request on HMP V3-V5 returns data, yet no plot appears

image

Could be that there is a point that breaks it? Not sure

asizemore:
Resolves #528 

Recent backend work allows the differential abundance request to now accept an optional parameter `pValueFloor` which then the backend uses to prevent pvalues from being so low we can't represent them (for example on my laptop 1e-400 just becomes 0). The backend also uses the pValueFloor to calculate what the associated floor for the adjusted p value should be. Finally, it returns both of these values to the frontend. Read more here: https://github.com/VEuPathDB/microbiomeComputations/pull/53

This PR makes use of all the above work by sending a p value floor to the backend, then incorporating the returned pValueFloor and adjustedPValueFloor into the volcano plot. The volcano plot uses these values in the tooltip and for plotting these points.

Some more points about this PR:
- I've changed `minPValueCap` to `pValueFloor` in order to be consistent with the backend.
- Since the adjustedPValueFloor should be calculated based on the pValueFloor, i've wrapped these together in one object to make it clear that they are related.
- the floor also affects when the max y axis is truncated, so you'll see updates around this logic. We don't want to show the truncation bar if we're already showing the capped y axis values (which are the max of the data anyways) 
- this PR also fixes the tooltip display names in the volcano stories

Testing:
- qa backend is ready!
- Not all diff abund computes will return these super small p values. One that i know does (and so is good for testing) is HMP V3-V5, Genus, Host Body Habitat, Colon, Skin (same as in the linked issue).


d-callan:
wondering if this is ready for review?

asizemore:
not ready for review yet!

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -188,6 +189,9 @@ export function DifferentialAbundanceConfiguration(
     configuration.differentialAbundanceMethod =
       DIFFERENTIAL_ABUNDANCE_METHODS[0];
 
+  // Set the pValueFloor here. May change for other apps.
+  if (configuration) configuration.pValueFloor = '1e-200';
```
jernestmyers, line 193: Would we rather this value come from the `DefaultStatisticsFloors` object?

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -415,6 +416,12 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
         ]
       : [];
 
+  // Record any floors for the p value and adjusted p value sent to us from the backend.
+  const statisticsFloors: StatisticsFloors = {
+    pValueFloor: (data.value && Number(data.value.pValueFloor)) ?? 0,
```
jernestmyers, line 421: Similarly, if our default `DefaultStatisticsFloors` object assigns `1e-200` for `pValueFloor`, then should this fallback value not also be `1e-200` (and be derived from that same object)?

jernestmyers (approved):
Just a couple questions re: the defaults, but otherwise things seem okay to me!

asizemore:
@jernestmyers your questions were deep indeed! After much pondering i've come up with the following:
1. A volcano plot totally by itself should not floor data (this is the same as pValueFloor = 0). If the plot assumes a nonzero floor and the data has a different floor, then one could make a quite confusing plot on accident. A floor is only really really needed if one of the points has pvalue = 0. To better address this case, i've udpated the volcano to throw an error if there is a  point with pvalue = 0 and no floor specified. This prevents someone from making a plot with a point plotted at infinity (not helpful) and helps them know what to do to make their plot more meaningful.
2. We do still want our computation to use a floor, so the floor in the diff abund plugin is not derived from the DefaultStatisticsFloors object.

I've also updated the comments to make this all cleaner. If you happen to take a look and have any more ideas or suggestions please let me know and i'll make a new branch!

PR #566, opened 2023-10-24
Bring back old FlyTo behaviour for now

bobular:
I wasn't too keen on the map automatically zooming to data whenever the user zoomed out to zoom level 1.  This brings back the old behaviour without too much bother.


dmfalke (approved):
This is great!

Issue #529, opened 2023-10-02
Volcano: add axis truncation warning


asizemore:
When we change the axis ranges in the volcano plot such that some points get cut off, we get the nice truncation bars but no dismissable warning at the range input. Can we please add the warning so that it's consistent with the other plots and to explain what the yellow bars mean?

PR #570, opened 2023-10-24
add truncation warning to Volcano Viz

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/529.

Truncation warnings are added for both independent and dependent axes. A screenshot is attached in the following:

microbiome volcano truncation warning01

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
@@ -264,30 +264,6 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
     }
   }, [data.value, vizConfig.dependentAxisRange, rawDataMinMaxValues]);
 
-  // determine default min/max for x and y
-  const defaultIndependentAxisRange = useMemo(() => {
-    if (!data.value) return undefined;
-    return {
-      min: rawDataMinMaxValues.x.min,
-      max: rawDataMinMaxValues.x.max,
-    };
-  }, [data.value, rawDataMinMaxValues.x]);
-
-  const defaultDependentAxisRange = useMemo(() => {
-    if (!data.value) return undefined;
-    // since y values are -log10, it needs to check min & max
-    return {
-      min: Math.min(
-        -Math.log10(rawDataMinMaxValues.y.min),
-        -Math.log10(rawDataMinMaxValues.y.max)
-      ),
-      max: Math.max(
-        -Math.log10(rawDataMinMaxValues.y.min),
-        -Math.log10(rawDataMinMaxValues.y.max)
-      ),
-    };
-  }, [data.value, rawDataMinMaxValues.y]);
-
   const significanceThreshold =
     vizConfig.significanceThreshold ?? DEFAULT_SIG_THRESHOLD;
   const effectSizeThreshold =
@@ -483,7 +459,7 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
     ...(data.value ? {} : EMPTY_VIZ_AXIS_RANGES),
   };
 
-  // set truncation flags: will see if this is reusable with other application
+  // set truncation flags
   const {
     truncationConfigIndependentAxisMin,
     truncationConfigIndependentAxisMax,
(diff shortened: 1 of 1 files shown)


packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -448,6 +483,64 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
...
+    }
+  }, [
+    truncationConfigIndependentAxisMin,
+    truncationConfigIndependentAxisMax,
+    setTruncatedIndependentAxisWarning,
+  ]);
+
+  useEffect(() => {
+    if (
+      // (truncationConfigDependentAxisMin || truncationConfigDependentAxisMax) &&
```
asizemore, line 526: nit: old comments can be removed

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -448,6 +483,64 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
     ...(data.value ? {} : EMPTY_VIZ_AXIS_RANGES),
   };
 
+  // set truncation flags: will see if this is reusable with other application
```
asizemore, line 486: nit: old comments can be removed

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -448,6 +483,64 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
...
+  const {
+    truncationConfigIndependentAxisMin,
+    truncationConfigIndependentAxisMax,
+    truncationConfigDependentAxisMin,
+    truncationConfigDependentAxisMax,
+  } = useMemo(
+    () =>
+      truncationConfig(
+        {
+          independentAxisRange: defaultIndependentAxisRange,
```
asizemore, line 496: Instead of making new default...AxisRange variables, can we just have this memo dependent on the `rawDataMinMaxValues`? The independent axis range would just be 
```
{
    min: rawDataMinMaxValues.x.min,
    max: rawDataMinMaxValues.x.max,
}
```
like you have above, and then the dependent axis range could be
```
{
    min: -Math.log10(rawDataMinMaxValues.y.max),
    max: -Math.log10(rawDataMinMaxValues.y.min)
}
```
.


I'm hoping that will keep the component simpler, since we already have a few variables that have to do with the range of the data or axes, and i have trouble keeping them all straight as it is :D 
What do you think?

asizemore (commented):
Thank you for taking this on @moontrip !

I had a few small comments and then a larger one that would be good to make sure we're on the same page about. In general I'm hoping to just keep the number of variables related to the axes and data ranges down to a minimum because there are already so many but let me know if im missing something or there's more to consider!

moontrip:
> Thank you for taking this on @moontrip !
> 
> I had a few small comments and then a larger one that would be good to make sure we're on the same page about. In general I'm hoping to just keep the number of variables related to the axes and data ranges down to a minimum because there are already so many but let me know if im missing something or there's more to consider!

Thank you for your feedbacks, @asizemore 👍 The reason why I introduced new variables like defaultIndependentAxisRange etc. was simply because other Viz also used the same variable names. But I can understand your concern so I followed your suggestion in the new update. 

Just a side note: I merged the latest main branch into this PR, but somehow I could not see the plot with the same example throughout this work (shown in the following screenshot). I could see an error message like below. IIRC, I saw a ticket concerning this recently (P-value floor), but I lost track of it.

`Unexpected backend type(s) encountered: Invalid value `null` supplied to adjustedPValueFloor (type: string) of 1 (type: Partial<{ pValueFloor: string, adjustedPValueFloor: string }>) of [root] (type: ({ effectSizeLabel: string, statistics: Array<Partial<{ effectSize: string, pValue: string, adjustedPValue: string, pointID: string }>> } & Partial<{ pValueFloor: string, adjustedPValueFloor: string }>))`

mbio data01

asizemore:
Thanks @moontrip !! Good catch! I'll go fix this bug hang on...

asizemore:
Glad you found the bug! The fix is here: https://github.com/VEuPathDB/web-monorepo/pull/579

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
```diff
@@ -460,6 +471,71 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
...
+          dependentAxisRange: {
+            min: Math.min(
+              -Math.log10(rawDataMinMaxValues.y.min),
+              -Math.log10(rawDataMinMaxValues.y.max)
+            ),
+            max: Math.max(
+              -Math.log10(rawDataMinMaxValues.y.min),
+              -Math.log10(rawDataMinMaxValues.y.max)
+            ),
+          },
```
asizemore, line 497: We know that the -log(y.max) will always be less than -log(y.min), so we can simplify the dependentAxisRange to:

```suggestion
          dependentAxisRange: {
            min:  -Math.log10(rawDataMinMaxValues.y.max),
            max: -Math.log10(rawDataMinMaxValues.y.min),
          },
```

asizemore (approved):
I had one more suggestion but otherwise looks great so i'll go ahead and approve! Thanks for your help!

packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
moontrip: Thanks @asizemore for your suggestion 👍  I applied your changes to the new commit, so I will merge this PR.

PR #576, opened 2023-10-27
Possible fix for reset/flyTo all data; fix animation other worlds bug

### LINKED ISSUE #575: Map - animation broken when not in "main world"
If, while zoomed out fully, you pan 360 degrees east or west into "another world" and then zoom in to a marker, the semantic zoom animation (aggregation/disaggregation) is not working.  There are also sometimes some distracting visuals where markers fly across the screen. It's likely related.

bobular:
Fixes #575 

But it looks like that won't get closed automatically because this doesn't merge to main?

### DIFF
--- a/packages/libs/components/src/map/SemanticMarkers.tsx
+++ b/packages/libs/components/src/map/SemanticMarkers.tsx
@@ -163,7 +163,8 @@ export default function SemanticMarkers({
     [consolidatedMarkers]
   );
 
-  useFlyToMarkers({ markers: refinedMarkers, flyToMarkers, flyToMarkersDelay });
+  // this should use the unadulterated markers (which are always in the "main world")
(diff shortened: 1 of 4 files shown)


bobular:
Need to test the following once #597 is addressed.

Megastudy.
Filter: species = Culex nigripalpus + Culex tarsalis
Bubble markers (e.g. sample.sex: female/all)

Get bubbles on the screen. Hit the fly-to reset button.

Then add Anopheles punctipennis to the filter.  It should look like this:
image

Then hit the fly-to button and it may end up looking like this (the markers are different sizes, slightly):
image

dmfalke (approved):

dmfalke:
> Fixes #575
> 
> But it looks like that won't get closed automatically because this doesn't merge to main?

We should add the "fixes..." comment to the parent pr

PR #577, opened 2023-10-30
Add more reference org filters

jernestmyers:
Resolves https://redmine.apidb.org/issues/50190

Added a filter for reference strains to below components. To be frank, I don't love the vertical bar that separates the toggles in the `OrganismFilter`, but I've left it for now.

`OrganismParam`:
image

`OrganismFilter`:
image


dmfalke (approved):
I'm also not crazy about the vertical bar, but we can see what others have to say.

PR #579, opened 2023-10-31
allow adjustedPValueFloor to also be null

### LINKED ISSUE #578: Volcano: handle null adjusted p value floor
Found in https://github.com/VEuPathDB/web-monorepo/pull/570

Sometimes the adjusted p value floor is null, when then is not handled well by the plot and gives us the dreaded empty plot. The adjpvaluefloor being null indicates that there is no floor, and that no points have been floored. Let's make that a reality in the component!

asizemore:
Resolves #578 

The backend sometimes sends an adjustedPValueFloor = null. This PR adds that option to the api.

### DIFF
--- a/packages/libs/eda/src/lib/core/api/DataClient/types.ts
+++ b/packages/libs/eda/src/lib/core/api/DataClient/types.ts
@@ -373,7 +373,7 @@ export const VolcanoPlotResponse = intersection([
   }),
   partial({
     pValueFloor: string,
-    adjustedPValueFloor: string,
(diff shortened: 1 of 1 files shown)


asizemore:
Only one review is necessary i think - just added the both of you in case there is lots else going on!

jernestmyers (approved):
A doozy!

moontrip (approved):
I was not quite sure if it was a frontend or backend fix but looks like frontend fix is just sufficient 👍 I checked that it worked just fine!

asizemore:
thanks team!!

PR #580, opened 2023-10-31
Simple record page

dmfalke:
This PR introduces a simplified version of a record page, that can be embedded in popups. Attributes and tables can be specified via query params.

Route paths for this page look like:
```
/embed-record/long_read_transcript/TALONT000034884_DS_2e2a681fa9?tables=SampleInfo
```

which results in a page that looks like:
image


aurreco-uga (approved):
i assume the links at the top work?  we shoudl review the UX (such as opening in new tab?)

dmfalke:
The links work. I forgot that the download link goes to a new page. We'll either need to hide that link, or make the download page work in popups (hide header/footer, etc).

PR #583, opened 2023-11-01
Update visible help styling

dmfalke:
This PR includes the following changes:
1. Puts visible help in an info box
2. Add global style rules for `<dl>` blocks: `<dt>` and `<dd>` are on the same line, and `<dt>` is in bold.

image


aurreco-uga (approved):
some improvements for when there is time:
- i dont think the  i  icon is needed within the box, nor the question mark icon by the param name when we provide inline description.
- the tree should not offer count  or search box when only under 10 or 15 nodes.

Issue #449, opened 2023-08-25
Map - new feature - user-control of geohash level


bobular:
At the moment the Leaflet zoom level is converted via a pre-defined mapping to geohash level.

It would be nice to offer the user some control over this - I was thinking a +2 +1 0 -1 -2 (or perhaps just +1 0 -1) adjustment that the user could apply via a new UI element.

I think that's basically it.  But if anyone can think of any unintended consequences, please comment.

danicahelb:
it would be nice to aggregate to the country level (ie, one marker per country) for individual studies where there are study sites in multiple countries. For these individual studies, it would be expected that most countries would NOT have data so the issue Bob raised above would not matter

bobular:
geohash_0 (containing the empty string `''` would be nice to have).  Easiest route is to add it to the database and make sure all the code (back and front end) doesn't treat it as a false value. Some other things might break though - grid lines would have to be disabled.

bobular:
Making a new ticket for aggregating by administrative region, will edit the title and reference this ticket in the new one.

PR #588, opened 2023-11-02
Multiple marker selection via BoundsDriftMarker and SemanticMarkers

bobular:
Note: targets https://github.com/VEuPathDB/web-monorepo/pull/534 

Hi @moontrip and @dmfalke 

I wanted to try my suggestion in https://github.com/VEuPathDB/web-monorepo/pull/513#issuecomment-1747598022 and I have found that it works pretty well and has removed quite a lot of code. I've also made it more 'reacty' and got rid of one jquery-like bit.

I have only sorted out the `@veupathdb/components` side of things - and fixed all the stories (hopefully reverting the old marker stories back to what they were before, so only the MarkerSelection story has selectedMarker logic).

I did introduce a `useEffect` to do pruning on `selectedMarkers` but I'd say this was OK. If you pan a selected marker off-screen, you would expect to trigger a side-effect maybe? Nothing is triggered if nothing needs pruning.

I've changed the spec of `selectedMarkers` back to `string[]` as per the original spec.

Some things to do

- [x] test chart markers (oops I forgot)
- [x] wire back into SAM and store the selectedMarkers in appState/markerConfiguration? - we could discuss with @dmfalke the best approach here (to avoid triplicate code in each MapType)
- [x] if possible fix the highlighting behaviour so that after the click, the highlighted marker is not dimmed - or... add a highlight around the popup (I think I prefer the latter) --> new standalone ticket https://github.com/VEuPathDB/web-monorepo/issues/617
- [x] disable multiple-marker in SAM via commenting or other easily reversible approach --> new ticket for wiring https://github.com/VEuPathDB/web-monorepo/issues/618
- [ ] diagnose/address popup refresh/flash due to changes in `selectedMarkers` which is likely related to marker churn in `SemanticMarkers`
- [ ] address double click issues


moontrip (commented):
@bobular Thank you for testing and refactoring my work. Now I understand what you had in mind! 👍 I have reviewed and tested this PR and found that it had some errors and SAM part was not modified yet. Thus, I branched out and made changes at https://github.com/VEuPathDB/web-monorepo/pull/589 for your convenience 😄: my PR will also work with SAM, but I did not consider your second to-do, appState/markerConfiguration, in there.

So here are my thoughts on some of your to-dos:

- test chart markers (oops I forgot): I tested this and there were some glitches so I fixed them at the PR https://github.com/VEuPathDB/web-monorepo/pull/589
- if possible fix the highlighting behaviour so that after the click, the highlighted marker is not dimmed - or... add a highlight around the popup (I think I prefer the latter): If I understand your speaking correctly, it is not the issue of marker click, but it is due to the mouse-over event for both popup and rectangular gray box. I didn't look into this carefully, but I guess that there might be a CSS effect to make marker dim for the mouse-over marker.

By the way, I noticed a minor(?) disadvantage on your approach using useEffect. Since selectedMarkers are updated via useEffect, there is some delay to update selectedMarkers array. As a side effect, when marker drift happens (aggregation/deaggregation, i.e., when geohash level is changed), highlights remain for a while, and then disappear: this is more evident when the number of markers are large (e.g., chart marker at marker selection storybook). But it may just be okay.

bobular:
> If I understand your speaking correctly, it is not the issue of marker click, but it is due to the mouse-over event for both popup and rectangular gray box. I didn't look into this carefully, but I guess that there might be a CSS effect to make marker dim for the mouse-over marker.

Agreed, it's pre-existing functionality that dims the marker so that the bounding box grey rectangle is more visible while the popup is shown. I think a yellow border around the popup would solve this for me.  At the moment the only feedback I get from highlighting is a dull yellow. More like "lowlighting" lol...

bobular:
I'll test the speed/responsiveness issue later. I wonder if we should have a snackbar that says "[Some of] your selected markers have been deselected because they have gone off-screen" or "Your marker selection has been cancelled because the marker aggregation level has changed due to zooming" (or similar!)

moontrip:
> > If I understand your speaking correctly, it is not the issue of marker click, but it is due to the mouse-over event for both popup and rectangular gray box. I didn't look into this carefully, but I guess that there might be a CSS effect to make marker dim for the mouse-over marker.
> 
> Agreed, it's pre-existing functionality that dims the marker so that the bounding box grey rectangle is more visible while the popup is shown. I think a yellow border around the popup would solve this for me. At the moment the only feedback I get from highlighting is a dull yellow. More like "lowlighting" lol...

@bobular I cannot think a quick solution off the top of my head for highlighting popup: it may be a bit tricky. Pop-up only shows when doing mouse-over a marker, so it may not be that visible in a general case.

And, dull yellow! 😄 Perhaps other color may also work, like (dull) orange 😄

bobular:
> By the way, I noticed a minor(?) disadvantage on your approach using useEffect. Since selectedMarkers are updated via useEffect, there is some delay to update selectedMarkers array. As a side effect, when marker drift happens (aggregation/deaggregation, i.e., when geohash level is changed), highlights remain for a while, and then disappear: this is more evident when the number of markers are large (e.g., chart marker at marker selection storybook). But it may just be okay.

I tested with the chart marker story too and I think it's OK - it only happens when aggregating on zooming out. It could even be seen as a feature? It shows you where your selected markers were for about 1 second. Then if a snackbar came up to say what had happened, I think that would be great. But Dave usually has other opinions on snackbars!

Tested also with donuts in SAM - aggregation behaviour is the same as in the story (though slower due to real-life data requests) - I think it's fine.

What does need addressing is the loss of marker double-click (should zoom-to-marker-bounds). Could you see if we can keep both click behaviours please?

moontrip:
> I'll test the speed/responsiveness issue later. I wonder if we should have a snackbar, triggered by the effect, that says "[Some of] your selected markers have been deselected because they have gone off-screen" or "Your marker selection has been cancelled because the marker aggregation level has changed due to zooming" (or similar!)

@bobular I did some test with SAM. It is more noticeable when zooming-out (small video is attached). I wonder how it will work with larger number of markers like mega study etc. but we may live with that anyway? :) Live site will be faster than my VM, so that delay may not be that noticeable in there.

https://github.com/VEuPathDB/web-monorepo/assets/12802305/a88aaa49-acc3-4341-bef4-1413137bb83a

bobular:
Thanks for the video - I've been recording one too.

I think what you demonstrate is absolutely fine. While I was recording I couldn't always replicate it. Sometimes the aggregation animation seemed broken (while the disaggregation was still animating fine). It may be related to https://github.com/VEuPathDB/web-monorepo/issues/575 

Anyway, here's my video, highlighting (pun intended) another small issue, and discussing the dim yellow issue too.


https://github.com/VEuPathDB/web-monorepo/assets/308639/893d5844-2001-4f1e-b4bc-4f59f5a545d4

packages/libs/components/src/map/SemanticMarkers.tsx
```diff
@@ -149,12 +155,30 @@ export default function SemanticMarkers({
...
+  useEffect(() => {
+    if (setSelectedMarkers && selectedMarkers) {
+      const prunedSelectedMarkers = selectedMarkers.filter((id) =>
+        consolidatedMarkers.find(({ props }) => id === props.id)
+      );
+
+      if (prunedSelectedMarkers.length < selectedMarkers.length)
+        setSelectedMarkers(prunedSelectedMarkers);
+    }
+  }, [consolidatedMarkers, selectedMarkers]);
```
dmfalke, line 168: I think using `useEffect` here is fine. However, I'm not convinced that we should do this (remove selected markers when they go offscreen). This means that a user cannot select markers that fall outside of the users viewport size, _at a given geohash_. Would it be "good enough" to clear marker selection when the geohash changes (and this would happen in the MapAnalysis component)?

bobular: If we keep them when they go off-screen then it's even more important to put something in the header to say how many things are currently selected. I'm not saying that's a bad idea, but it needs further UI thought: can you cancel your selections from the header also?

The use case for maintaining selections outside the viewport is presumably so the user can zoom into Australia, pick a few markers, pan over to South America and pick a few more.

At the moment the plan is to _add_ the marker-selection as a filter to the floating viz filters. The floating viz filters already include the viewport, though of course that could be removed if a selected-marker-based filter is being applied. 

I personally think the UX is a lot simpler if we restrict everything to what is currently on-screen/in-viewport.

If the current whole-world/continent leaflet zoom level isn't giving enough geohash resolution, there is always this ticket to help: https://github.com/VEuPathDB/web-monorepo/issues/449 (I have moved it to the new board)

bobular: I'm half wondering if we should be selecting gehash rectangles instead of markers. Perhaps with shift-click or ctrl-click. It feeds into the discussion of little filters and floating visualizations here: https://github.com/VEuPathDB/web-monorepo/issues/596

dmfalke: > I'm half wondering if we should be selecting gehash rectangles instead of markers. Perhaps with shift-click or ctrl-click. It feeds into the discussion of little filters and floating visualizations here: #596 (see point 3.)

I was just thinking the same thing!

dmfalke: I'm convinced about removing the selection.

moontrip:
@bobular Just for a note. I think I did comment somewhere that I will look into the popup location bug(?) but can't find it. Anyway. I have been trying to resolve that bug.

bobular:
Go ahead and merge this to the parent branch (and move the to-do list over to that PR), or re-target this PR to `main`.  Thanks!

moontrip:
> After merging #610 into this, please go ahead and merge this to the parent branch (and move the to-do list over to that PR), or re-target this PR to `main` and close #534. Thanks!

@bobular thank you for your suggestion! I think that it is better to keep this PR as a candidate to be merged into main so that we can easily keep track of our progress/comments. I will close the parent PR and leave this to be merged into main 👍

moontrip:
@bobular FYI, I found a solution to enable marker-double click again at https://github.com/VEuPathDB/web-monorepo/pull/615. It would be great if you can check it out when you are available.

bobular:
I've checked off the remaining items/made new tickets.

We could merge this when it is disabled (see new checkbox item in description).

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/DonutMarkerMapType.tsx
```diff
@@ -304,6 +308,7 @@ function MapLayerComponent(props: MapTypeMapLayerProps) {
   if (markerDataResponse.error)
     return <MapFloatingErrorDiv error={markerDataResponse.error} />;
 
+  // pass selectedMarkers and its state function
```
bobular, line 311: Oops we need to get rid of this comment

dmfalke (approved):

bobular:
I think we're in a relatively good place here, but we should not merge this before b66, which gives us time to address the remaining minor issues (minor to the user - major for us, probably!)

PR #589, opened 2023-11-03
fix errors and enable multi-selection marker for SAM

moontrip:
@bobular This was branched out from your branch (https://github.com/VEuPathDB/web-monorepo/tree/multiple-marker-selection-simplified) and intended to fix errors as well as enabling multi-selecting markers with SAM.

From my tests, this worked fine but please test this and merge: will be merged into your branch

### DIFF
--- a/packages/libs/eda/src/lib/map/MapVEu.scss
+++ b/packages/libs/eda/src/lib/map/MapVEu.scss
@@ -27,24 +27,22 @@
     }
   }
 }
-/* donut and bubble markers highlight */
-.highlight-donutmarker,
-.highlight-bubblemarker {
-  background-clip: padding-box;
-  border-radius: 50%;
-  /* add inset to reduce a gap between the element and the box-shadow */
-  box-shadow: 2.5px 2.5px 0px 6px rgba(255, 255, 0, 1),
-    2.5px 2.5px 0px 2px rgba(255, 255, 0, 1) inset;
-  z-index: 9999 !important;
-}
 
-/* chart marker highlight */
-.highlight-chartmarker {
-  background-clip: padding-box;
-  /* the rounded corner of the chart marker has 10px in radius */
-  border-radius: 10px;
-  /* add inset to reduce a gap between the element and the box-shadow */
-  box-shadow: -1px -0.5px 0px 6px rgba(255, 255, 0, 1),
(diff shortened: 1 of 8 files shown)


packages/libs/eda/src/lib/map/MapVEu.scss
bobular, line 1: Hi @moontrip, as you know, I'm no CSS wizard.

What's going on here with the apparent duplication between this file and `packages/libs/components/public/css/vb-popbio-maps.css` ?

In #588 I only [changed vb-popbio-maps.css](https://github.com/VEuPathDB/web-monorepo/pull/588/files#diff-28f7e79019ad4cc46a38827478ce9e21670348ff2beb8a67b1301dc1404826f5R267-R268) and the story seemed to work OK, but I much prefer the structured SCSS approach. Does the SAM use this? Can we remove the redundancy somehow?

bobular (approved):
This looks great. I'd really appreciate if you could sort the CSS redundancy if possible!

packages/libs/eda/src/lib/map/MapVEu.scss
moontrip: Good morning @bobular 👋 

Yes there are two CSS related stuffs. I may be wrong, but I think that we are not using vb-popbio-maps.css for EDA SAM? It can be easily checked by removing SCSS's highlight stuff and see if SAM's highlight works.

But I think that map-related storybook relies on vb-popbio-maps.css, beyond highlight stuff: again I may be incorrect hmm

moontrip: @bobular I tested the CSS and it seems to be independent each other. That is, if I remove highlight part from MapVEu.scss, then SAM does not recognize highlight-related CSS, which means SAM does not rely on vb-popbio-maps.css. I didn't test the opposite case, removing highlight-related CSS from vb-popbio-maps.css, but I can guess that it would not work for storybook because the storybook does not/cannot refer to EDA's MapVEu.scss 

So in summary, we need to keep both CSS and SCSS: one for story and the other for EDA SAM, separately. IIRC, that was the reason why I set up both CSS and SCSS.

bobular: OK, thanks - sounds like something we should discuss next Wednesday, and then make a ticket for if appropriate.
Go ahead and merge this and I'll take #588 out of draft so we can merge that too.

PR #590, opened 2023-11-03
Eda study type error

dmfalke:
This PR will throw errors caught when attempting to load eda study metadata, if it is not a `FetchClientError`. This change will cause io-ts error to get thrown, which is useful to prevent masking that error. The styling of the error modal is also updated.

image


jernestmyers (approved):
LGTM! 💪

PR #594, opened 2023-11-04
srt new icon refs to #52746

aurreco-uga:
i need to add a hook to access the current build/release number (here harcoded)..  but this is working!
Screen Shot 2023-11-04 at 5 30 33 PM


dmfalke (commented):
This looks good, but I'm not sure about the `newBuild` property on `RadioList`. It seems a little too specific to WDK features.

Maybe we could add a new component named `WdkFeature` that takes an optional `buildNumber` property, and performs the logic of including the icon? I'm just thinking of a way to maximize reuse.

You can get the current build with a hook like this:
```ts
function useCurrentBuildNumber(): number | undefined {
  return useWdkService(async wdkService => {
    const config = await wdkService.getConfig();
    return config.buildNumber;
  })
}
```

and can be used like this:
```ts
// type is `number | undefined`
const currentBuild = useCurrentBuild();
```

jernestmyers:
I opted for a low-cost solution for now.

dmfalke (approved):

PR #598, opened 2023-11-06
Wdk client 191 first table column sticky

chowington:
Closes https://github.com/VEuPathDB/WDKClient/issues/191


dmfalke:
This is looking really great. I have some comments/questions:

1. It looks like Mesa will only support one fixed column. I think we will have some use cases where we want 2 or 3 fixed columns. Would it be hard to support that?
2. The left/right cell borders disappeared. Is that intentional?
3. I'm not crazy about the sticky column having a gray background. Could we try something more subtle?
4. We might need to get feedback from others before merging. I'd be happy to put this on my dev site.

Here is a screenshot of the table in Firefox:

image

chowington:
@dmfalke, I implemented almost all of your suggestions. The only thing I couldn't figure out was keeping the side borders on the first column top cell. I played around a lot with the CSS, but couldn't get those borders to stay like the other first columns cells. Hopefully it doesn't stand out too much. What do you think?

dmfalke (commented):
Thanks for the update, @chowington! I'm only seeing the side border on the first column in Firefox (see screenshot). Also, I think we might need to consider preventing the table from being taller than the screen. Now that the table container scrolls horizontally, the horizontal scrollbar is off screen (I think the browser window was scrollable before, so it was visible at all times).  

image

dmfalke:
Adding `border-collapse: separate;` to the CSS rule for `.MesaComponent .DataTable table` seems to do the trick. Maybe give that a shot?

dmfalke (commented):
This is looking awesome! I think there might be one last change to make.

The top and bottom border of the table header seems to be lost in the eda workspace. Maybe this is when there aren't any fixed columns?:

image

I played with the CSS a little bit, and setting `border-collapse: collapse` fixes this in the eda workspace. I'm wondering if we can set the value of `border-collapse` based on if there are fixed columns? Or maybe there is something else causing this?

Sorry this keeps dragging on!

dmfalke (commented):
@chowington I hope you don't mind--I just pushed some updates that fix the issue with the cell borders. It also fixes the sticky header offset on genomic sites. Let me know what you think.

chowington:
@dmfalke Thanks for your work. I still see a couple of bugs: There's a pixel column of bleed-through on the left of the first header cell, and the last cell of the sticky column loses its borders when scrolling horizontally.

image

dmfalke:
> @dmfalke Thanks for your work! I still see a couple of bugs: There's a pixel column of bleed-through on the left of the first header cell, and the last cell of the sticky column loses its borders when scrolling horizontally.
> 
> image

Ah, good catch! Will try to fix it.

dmfalke:
@chowington I just pushed another commit. Let me know how it looks to you.

chowington:
> @chowington I just pushed another commit. Let me know how it looks to you.

Looks good! I fixed yet another few issues. Final review? 🤞

dmfalke (approved):
:rocket:

Issue #553, opened 2023-10-15
EDA studies menu:  ability to collapse UDs, or show them below?


aurreco-uga:
<img width="1253" alt="EDAStudyMenu-collapseUDs" src="https://github.com/VEuPathDB/web-monorepo/assets/16354313/56c3cecb-42fc-4ad3-b4c9-548f0b4ac945">

PR #600, opened 2023-11-06
Add collapsible section to studies menu

jernestmyers:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/553

~~NOTE: This has not been added to mbio's menu. Awaiting review before doing so.~~

UPDATE: This has now been added to mbio's menu as well.

https://github.com/VEuPathDB/web-monorepo/assets/69446567/8ed1e6b6-6aa1-40ac-bb19-0fd0221ed41f


dmfalke (commented):
This looks great. Implementation looks great. There is a voice in my head saying that we should use a native `<details>` element... you know, leverage the platform and all. Thoughts?

jernestmyers:
> This looks great. Implementation looks great. There is a voice in my head saying that we should use a native `<details>` element... you know, leverage the platform and all. Thoughts?

I ran into issues with styling `<details>` elements in the past, but a quick read on MDN suggests I **_shouldn't have_**...

We also have `CollapsibleSection` in WDKClient, which I opted against due to a prior memory of limitations (without recollection of _**what**_ limitations)...

dmfalke:
Fair enough!

dmfalke:
This looks great, but I think curated studies should default to open. I suspect clinepi outreach would prefer that.

dmfalke (approved):
:rocket:

Issue #597, opened 2023-11-06
Map - trouble in Bubble Town...


bobular:
I noticed some discrepancies when testing global reset/fly-to-data - which I'll describe later.

[ This appears to be a post-refactor issue. The live site map doesn't even remember the marker mode when duplicating a tab, so it's not possible to reproduce there. ]

First a much easier way to reproduce a bubble problem is to get some bubble markers on the map and then duplicate the browser tab. All the markers end up teeny tiny:
![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/581d2961-e8ff-468f-9566-98f924b152a7)

The next comment will hopefully explain how to reduce the more complex fly-to related issue.

bobular:
OK, so I can't reproduce it in current `main`. Might be something specific to the branch I'm working on.

bobular:
That sounds like the same thing, thanks @moontrip

bobular:
Closed thanks to #604  :-)

Issue #603, opened 2023-11-07
Map - bubble marker size instability


bobular:
Need to test the following once #597 is addressed.

Megastudy.
Filter: species = Culex nigripalpus + Culex tarsalis
Bubble markers (e.g. sample.sex: female/all)

Get bubbles on the screen. Hit the fly-to reset button.

Then add Anopheles punctipennis to the filter.  It should look like this:
![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/5396015a-37e9-4260-9fc5-38a7066de20d)

Then hit the fly-to button and it may end up looking like this (the markers are different sizes, slightly):
![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/fab108da-0b95-4e05-822c-e6d390bcc69e)

It shouldn't be different. Suspect it's related to #597

_Originally posted by @bobular in https://github.com/VEuPathDB/web-monorepo/issues/576#issuecomment-1795560618_

bobular:
Easier reproduction:

Bubble mode as above.
Filter: species = Culex nigripalpus
Hit the fly-to-data button or get N. America in view.
Note the "snowman" in Florida
![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/9b621f82-d38a-4654-a54f-9381b11dd97a)

Now add Anopheles punctipennis to the filter.
New bubbles should appear but the snowman should stay the same.

Now nudge the map slightly and watch the markers shrink when the response comes back. (They shouldn't.)

If you remove Anopheles punctipennis, and then re-add it. The problem does NOT occur again, unless you clear the react-query cache (hit the "Clear" button in the react-query panel (flower icon)).

PR #604, opened 2023-11-07
fix initial bubble marker bug

moontrip:
Hi @dmfalke (and @bobular):

As commented at https://github.com/VEuPathDB/web-monorepo/issues/597#issuecomment-1795524714 about a bug concerning the bubble marker, I tried to find the reason and fix the bug. After delving into the issue, I realized that it was because marker data is prefetched via useQuery while the legendData via useLegendData hook is still being undefined. This leads the bubble markers to be not shown: more specifically all the marker data have 0 diameters & no colors due to undefined bubbleValueToColorMapper & bubbleValueToDiameterMapper functions at legendData. After sleeping on this, I finally found two possible solutions as follows:

a) Based on the pre-existing comment by Dave, "FIXME Don't make dependent on legend data", I tried to move legendData and the relevant parts to use legendData, i.e., finalMarkerData, out of useMarkerData, i.e. move them to the parent where useMarkerData is called, and then adjusted the parent accordingly. Based on my tests, this worked fine.

b) That being said, an idea came to my mind as an alternative solution after a while. So, basically this can also be considered as an issue of not calling useQuery again (refetch) when legendData is ready (not undefined), thus I changed the conditions of "enabled" option used in the useQuery by additionally checking if legendData is undefined. Certainly, this is much simpler solution than a) and I confirmed that this also worked out.

Due to simplicity, I made PR for b). What do you think @dmfalke?

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
@@ -595,11 +595,6 @@ function useMarkerData(props: DataProps) {
 
   const { numeratorValues, denominatorValues } = configuration;
 
-  const disabled =
-    numeratorValues?.length === 0 ||
-    denominatorValues?.length === 0 ||
-    !validateProportionValues(numeratorValues, denominatorValues);
-
   const studyEntities = useStudyEntities();
   const dataClient = useDataClient();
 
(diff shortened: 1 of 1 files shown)


dmfalke (approved):
Good detective work! I think this is a fine solution. Thanks! 

I know this is in draft mode, but I think it's ready to merge.

moontrip:
@dmfalke 🕵️ 
Thank you for your review and confirmation! 👍 I will merge this to main after updating this branch with the latest main

bobular:
Any reason not to close #597, @moontrip?

I'm looking at #603 now but that can be a separate issue!

moontrip:
@bobular Honestly, I am not quite sure if my fix is related to your ticket, #597. Can you please test it with your working branch?

bobular:
I guess this fixes the first bug/screenshot in #597 and then I started rambling on about a second bug which I have described in more detail in #603.  I'll edit #597 and close it.

PR #609, opened 2023-11-07
Bubble marker mode filter sensitivity

### LINKED ISSUE #582: Map - bubble marker proportion config should be filter-sensitive
If you filter on some species and then go to bubble marker config (variable=species) then the default numerator and denominator contain all the species (unfiltered). But the drop-downs are filter-sensitive (see screenshot). Both should show the filtered options, I think.

We need to figure out what to do when the filter changes:
1. reset to all numerator / all denominator
2. do a more sophisticated check to see what selections can be perserved

image

### LINKED ISSUE #603: Map - bubble marker size instability

bobular:
Fixes #582 
Fixes #603 

This PR also address some instability in marker sizes because `legendData` wasn't used as a dependency for the markers request.


packages/libs/eda/src/lib/map/analysis/utils/defaultOverlayConfig.ts
```diff
@@ -24,7 +24,6 @@ import { BubbleMarkerConfiguration } from '../MarkerConfiguration/BubbleMarkerCo
 
 export interface DefaultBubbleOverlayConfigProps {
   studyId: string;
-  filters: Filter[] | undefined;
```
bobular, line 27: This wasn't used so I removed it!

packages/libs/eda/src/lib/map/analysis/hooks/standaloneMapMarkers.tsx
bobular, line 1: Potentially remove the whole file and move the two exported items elsewhere.

```
packages/libs/eda/src/lib/map/analysis/mapTypes/shared.ts:import { GLOBAL_VIEWPORT } from '../hooks/standaloneMapMarkers';
packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BarMarkerMapType.tsx:import { ChartMarkerPropsWithCounts } from '../../hooks/standaloneMapMarkers';
```

bobular:
Hi @dmfalke - I'm keeping this in draft mode because there is one more thing to think about before merging.

What do we do with numerator/denominator selections when the user filters directly on the bubble overlay variable?

For example, I have markers configured to show `female/(male+female+unknown)`.  Then I filter on `sex=[male,mixed]` - the interface currently shows this:
image
and
image

I think the simplest UX is to reset the selections to `all/all` when filters change on this variable, directly.

So, I suggest we need a, you guessed it, effect in `useOverlayConfig` to watch for changes in `overlayVariable.vocabulary` and reset `numeratorValues` and `denominatorValues` to `undefined` when that happens, with a snackbar to tell the user!  The only problem is that every function that uses `useOverlayConfig` would need to be passed `updateConfiguration` also. Maybe as a second argument to `useOverlayConfig`, to avoid messing with the types too much?

What are your thoughts?

bobular:
For the imminent b66 release @dmfalke and I decided not to merge the more comprehensive fix to this problem (#643) because it isn't fully resolved to our satisfaction and may not be in time.

The temporary solution put forward here is that invalid numerator/denominator configurations can be fixed by selecting the correct values but the invalid configurations are not policed and the potentially erroneous back end request will still be made.

After the release we will address it properly.

Dave, please could you standardise the error message reporting? Errors are currently displayed in the Legend - but it would be great if they could use what you did in #651 - TIA!

dmfalke:
> Dave, please could you standardise the error message reporting? Errors are currently displayed in the Legend - but it would be great if they could use what you did in #651 - TIA!

Sure thing. I won't get to this until much later tonight, or tomorrow morning.

bobular:
This is hopefully good enough for b66.  Difficult to test at the moment due to back end issues.

The UX is NOT ideal. Changes in filters can invalidate the bubble marker proportion config, and this causes a broken map-markers response (and a nice pink message in the middle of the screen) but the user is not told that they can fix this by adjusting the bubble marker configuration.

dmfalke (approved):
Let's do this!

PR #610, opened 2023-11-08
address popup position issue when clicking marker

moontrip:
This will possibly address an issue of popup positions when selecting/deselecting marker, as shown in the video, https://github.com/VEuPathDB/web-monorepo/pull/588#issuecomment-1792724120. Note that this is branched out from https://github.com/VEuPathDB/web-monorepo/tree/multiple-marker-selection-simplified.

After 🕵️, it was noticed that the popup positions somehow kept changing when selecting/deselecting marker, which were used for calculating popup offsets, perhaps because of regenerating markers due to click events. A solution I used is to replace the popup-based positions with the marker-based positions because the latter remains consistent regardless of click events. An example video is attached in the following. 

https://github.com/VEuPathDB/web-monorepo/assets/12802305/83100fdd-3169-4ea4-842d-7928c3132651


bobular:
I hopefully merged main into all the parent branches and up to this one...  So that I could run it with the qa.vectorbase back end...

bobular (commented):
The popup is now in the right position, which is great!

Would you be able to investigate why it flashes every two seconds please?

bobular (approved):
I decided to approve so we can merge all this down to one branch away from `main`.
It's getting too complicated.
Let's look at the flashing popups next week.

moontrip:
@bobular thank you for your review! 👍

PR #615, opened 2023-11-09
address marker double-click

moontrip:
@bobular 

This will address the issue found by Bob, the loss of marker double-click, https://github.com/VEuPathDB/web-monorepo/pull/588#issuecomment-1792658637. 

After examining the issue, I think that the issue arose that by having marker selection with mouse click, marker double-click  could not be distinguished from marker single-click, in other words, double-click was treated as two times of single-click due to the single-click event for marker selection. Thus, a solution I found is to check the number of mouse click and then the current event of marker single-click (i.e., marker selection) is restricted accordingly, which is probably the simplest solution among another possible solutions. The number of mouse click can be checked via event.originalevent.detail. In this manner, the function for marker single-click to select/deselect marker is wrapped with a condition when event.originalevent.detail = 1 only. I have tested several times and this seems to work properly. Below is a sample video that shows single-click and then double-click.

https://github.com/VEuPathDB/web-monorepo/assets/12802305/3c209db2-9dac-48fa-8f69-9211c6a68063


packages/libs/components/src/map/BoundsDriftMarker.tsx
```diff
@@ -280,26 +280,28 @@ export default function BoundsDriftMarker({
...
-          prevSelectedMarkers.filter((id: string) => id !== props.id)
-        );
-      } else {
-        // select
-        setSelectedMarkers((prevSelectedMarkers: string[]) => [
-          ...prevSelectedMarkers,
-          props.id,
-        ]);
+    // check the number of mouse click and enable function for single click only
+    if (e.originalEvent.detail === 1) {
```
bobular, line 284: This is great. It seemed to me that it might not be a robust solution but `originalEvent.detail` is core Leaflet so it should be fine. No need to check for browser compatibility etc, which I suspected at first (from my first naive Google results).

bobular (approved):
LGTM!

packages/libs/components/src/map/BoundsDriftMarker.tsx
moontrip: @bobular thank you for your review 👍 I have tested both firefox and chrome browsers and they worked fine. I will merge this into the parent branch

bobular:
PS. I found it wasn't working for me (in Firefox at least).

console.log shows the first click still gets handled, but we'll look into it next week.

PR #621, opened 2023-11-09
Prevent marker popups flashing

### LINKED ISSUE #616: Map - marker mouse-over popups are flashing every two seconds
This is currently an issue on `main` on qa.vectorbase.org - in Firefox and Chrome at least.

Sorry no video!

As-yet unmerged work on [multiple marker selection](https://github.com/VEuPathDB/web-monorepo/pull/588), [popup placement](https://github.com/VEuPathDB/web-monorepo/pull/610) and [click handling](https://github.com/VEuPathDB/web-monorepo/pull/615) are not the cause (though note that the popup placement (now fixed) alternates every two seconds).

May be related to #573 ?

It's a recent regression though, just not sure exactly when/what caused it.

bobular:
Fixes #616 

TBH I'd like to give this some more thought and testing before merging, so I'm making this a draft.

Massive thanks to @moontrip for debugging the location of the problem.


bobular:
The popups still flash when we pan 360 degrees around the globe.

bobular:
This is fixed now, but with a deep-equals check to prevent unwanted re-renders. Better than the infinite loop currently on qa though!

I think one day we should revisit the marker animation, 360 degree world-wrapping and zoom-to-data. It's not very robust!

packages/libs/components/src/map/SemanticMarkers.tsx
```diff
@@ -124,8 +127,18 @@ export default function SemanticMarkers({
...
+      // Only check the props - hopefully more efficient!
+      // If there are any function props (there aren't right now) these
+      // will be compared with referential-equals.
+      if (
+        !isEqual(
+          recenteredMarkers.map(({ props }) => props),
+          prevMarkers.map(({ props }) => props)
+        )
+      )
+        setPrevMarkers(recenteredMarkers);
```
dmfalke, line 141: Can we rename `prevMarkers` to `prevRecenteredMarkers`, and `setPrevMarkers` to `setPrevRecenteredMarkers`?

bobular: Sure

dmfalke (approved):
looks good to me!

PR #630, opened 2023-11-10
Eda:  better handle single app mode

dmfalke:
This PR updates the logic related to `singleAppMode` in the EDA workspace. Previously, we assumed the first `computation` in the `analysis.descriptor.computations` array would match the `singleAppMode` computation type. The precludes the possibility that the analysis was created in some other context, or that the analysis contains other computations.

Now, we will look for any analysis that matches the `singleAppMode` type. If one is not found, one will be created before rendering the viz tab. Additionally, `singleAppMode` is being passed to `MapAnalysis`, to ensure that the computation is always created when the website is configured to use `singleAppMode`.

This is to enable using both the EDA Workspace view, and the Standalone Map view, for a single analysis object.

ETA this also addresses an issue with the SAM where a new analysis is not saved when the marker config is updated. This was fixed by adding a way to override the `createIfUnsaved` argument for the AnalysisState setters.

### DIFF
--- a/packages/libs/eda/src/lib/core/hooks/analysis.ts
+++ b/packages/libs/eda/src/lib/core/hooks/analysis.ts
@@ -47,18 +47,25 @@ export type AnalysisState = {
   analysis?: Analysis | NewAnalysis;
   /** Error object related to loading an analysis */
   error?: unknown;
-  /** Set the display name of the analysis. See @Setter */
+  /** Set the display name of the analysis. See {@link Setter}. */
   setName: Setter<Analysis['displayName']>;
-  /** Set the description of the analysis. */
+  /** Set the description of the analysis. See {@link Setter}. */
   setDescription: Setter<Analysis['description']>;
+  /** Set the notes of an analysis. See {@link Setter}. */
   setNotes: Setter<Analysis['notes']>;
+  /** Set the isPublic flag of an anlysis. See {@link Setter}. */
   setIsPublic: Setter<Analysis['isPublic']>;
+  /** Set the filters of an analysis. See {@link Setter}. */
   setFilters: Setter<Analysis['descriptor']['subset']['descriptor']>;
-  /** Set the list of configured computations of the analysis. See {@link Setter} */
+  /** Set the list of configured computations of the analysis. See {@link Setter}. */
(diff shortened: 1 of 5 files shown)


bobular (commented):
The code looks great but I'm still getting
`Error: Error: Could not find any computation app.` when loading a map analysis (old or brand new) in regular EDA.

From EDA to map seems to work fine.

Haven't looked into why, sorry!

bobular:
I think this is because I used a VectorBase map analysis.

We still have this line
```
      apps = apps.filter((app) => app.projects.includes(projectId));
```
in `ComputationRoute`.

So this is likely due to the fact that none of the "vanilla" pass-through visualisations are "allowed" for VectorBase (`app.projects`)

I'll test with a clinepi back end and then likely approve.

bobular (approved):
Worked fine both directions with clinepi as expected!

dmfalke:
> I think this is because I used a VectorBase map analysis.
> 
> We still have this line
> 
> ```
>       apps = apps.filter((app) => app.projects.includes(projectId));
> ```
> 
> in `ComputationRoute`.
> 
> So this is likely due to the fact that none of the "vanilla" pass-through visualisations are "allowed" for VectorBase (`app.projects`)
> 
> I'll test with a clinepi back end and then likely approve.

I think the problem is that you have `REACT_APP_SINGLE_APP_MODE=pass` set in your `.env.local` file. Try to comment that out when using VectorBase.

dmfalke:
> > I think this is because I used a VectorBase map analysis.
> > We still have this line
> > ```
> >       apps = apps.filter((app) => app.projects.includes(projectId));
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in `ComputationRoute`.
> > So this is likely due to the fact that none of the "vanilla" pass-through visualisations are "allowed" for VectorBase (`app.projects`)
> > I'll test with a clinepi back end and then likely approve.
> 
> I think the problem is that you have `REACT_APP_SINGLE_APP_MODE=pass` set in your `.env.local` file. Try to comment that out when using VectorBase.

I suppose we can log a warning in the developer console, and render the default UI

PR #641, opened 2023-11-14
Add links to EDA and SAM to switch between the two views

### LINKED ISSUE #625: Integrate SAM into ClinEpiDB (Phase I)
Redmine ticket: https://redmine.apidb.org/issues/52726

We need to be able to allow the user community to access curated ClinEpiDB studies on the SAM. 

Can we either (i) replace the Map viz tool to a linkout to SAM or (ii) add a new button (ie, viz type) on the viz tab that links out to SAM? In either case, the tool should be labeled `Explore on Interactive Map` and should have a `Beta` flag.

When the button is clicked, the study the user was working on should show in SAM in a **NEW tab** (it is important that ClinEpiDB stays open as a separate tab).

All filters that had been applied in ClinEpiDB should also be applied in the SAM

_Phase 2 (which won’t happen for a long time) would be to rework SAM so it is not stand-alone and have it replace the map viz tool. So, we would remove all functionality from the left menu EXCEPT for Configure Map_

dmfalke:
Closes #625 

This PR adds buttons to switch between the SAM and EDA views for a given analysis.

Screenshot from 2023-11-14 13-56-49

Screenshot from 2023-11-14 13-19-18

### DIFF
--- a/packages/libs/components/src/map/SemanticMarkers.tsx
+++ b/packages/libs/components/src/map/SemanticMarkers.tsx
@@ -10,7 +10,7 @@ import { AnimationFunction, Bounds } from './Types';
 import { BoundsDriftMarkerProps } from './BoundsDriftMarker';
 import { useMap } from 'react-leaflet';
 import { LatLngBounds } from 'leaflet';
-import { debounce } from 'lodash';
+import { debounce, isEqual } from 'lodash';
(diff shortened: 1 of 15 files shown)


jernestmyers (approved):
🟢

PR #643, opened 2023-11-14
Draft: handle incompatible numerator/denominator selections without snackbar effects

bobular:
Branched from #609 

The main thing here is that `useOverlayConfig` returns undefined if the numerator/denominators are incorrect in two ways
1. the existing condition that numerators must be present in denominator
2. all numerators and denominators must be in the variable's vocabulary (which can change now, due to filters)

When the overlay config is "bad" then the bubbles and bubble legend data requests are disabled and the user is directed to make amends.

How to test
1. at the moment, it's best to use the tick data from qa.vectorbase back end
2. for me that's https://localhost:3000/maps/DS_e18287e335/new
3. get bubble markers on `biological sex` with `female / female+unknown`
4. then add a filter on `biological sex = male, unknown`
5. go back to bubble config to see what's going on

Some rough edges still need to be worked out
- [ ] ideally we need to stop showing the markers when the config is bad, and the legend(s) need to show something helpful, rather than the previous legend data (I'm thinking that all map legends should have a :gear: icon in the bottom right corner which simply opens up the maptype configuration panel)
- [ ] how can we get the `useQuery`s to emphatically return "no data" so that we don't keep showing the previousData? I tried a `return undefined` in a queryFn and it didn't like that.
- [ ] basically we want the map to look the same after changing filters so that bubble proportion configs are invalid
       - directly after invalidation
       - after switching map types and back to bubble again (or refreshing the browser)
- [x] must test lineplot proportions and check not broken! - see how filter sensitivity works there
- [ ] check how many back end requests are made - are any causing back end errors?

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -138,6 +138,10 @@ import { Tooltip } from '@veupathdb/components/lib/components/widgets/Tooltip';
 import { FloatingLineplotExtraProps } from '../../../../map/analysis/hooks/plugins/lineplot';
 
 import * as DateMath from 'date-arithmetic';
+import {
+  invalidProportionText,
+  validateProportionValues,
+} from '../../../../map/analysis/utils/defaultOverlayConfig';
 
 const plotContainerStyles = {
   width: 750,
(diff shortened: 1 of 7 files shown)


bobular:
Tested lineplot (as a floater).

It looks like the new approach would help here. We just have to tell the user what the red outline means and decide whether or not to disable plotting when things are invalid (probably should!)

image

dmfalke (commented):
I'm still reviewing, but a preliminary thought:

If the config is invalid, in `useMarkerData` and `useLegendData`, maybe we can set the `disabled` flag for the `useQuery` calls, and just return something indicating that the configuration is invalid.

dmfalke:
I think the general UX for this is good. Once you are able to get rid of the markers, I think this will be in very good shape.

A suggestion for the legend when the config is bad is to show the invalid proportion error, along with a link/button to open the config menu.

I also think we should keep the invalid values, even if not selected, and to disable them if not selected. I think it will make what's going on more obvious to the user.

bobular:
Superseded by https://github.com/VEuPathDB/web-monorepo/pull/671
This should NOT be merged!

PR #644, opened 2023-11-15
Map marker popup highlight

moontrip:
@bobular I have tried to implement to highlighting marker popup when selecting a marker, as you discussed at https://github.com/VEuPathDB/web-monorepo/pull/588, though not quite sure if it is still valid. Anyway, a video capture is attached in the following:

https://github.com/VEuPathDB/web-monorepo/assets/12802305/1382d452-98d2-486b-8d0b-019796edf145

* this is branched out from https://github.com/VEuPathDB/web-monorepo/pull/588.


bobular:
Hi @moontrip - I tested and it works OK. There's one "flash" after clicking which it would be nice to investigate.

What I don't understand is this pull request/branch behaviour.  There are far too many diffs with the parent branch. I merged `main` into both this branch and the parent and it didn't help. Do you have any idea what's going on?

It would make this a lot easier to review!

bobular:
Actually `git diff origin/multiple-marker-selection-simplified` on the commandline shows just three files with a few differences - which I assume is the work you did here!

But `git diff multiple-marker-selection-simplified` shows 62 files different.  This PR shows 30.  Very strange.

moontrip:
> Hi @moontrip - I tested and it works OK. There's one "flash" after clicking which it would be nice to investigate.
> 
> What I don't understand is this pull request/branch behaviour. There are far too many diffs with the parent branch. I merged `main` into both this branch and the parent and it didn't help. Do you have any idea what's going on?
> 
> It would make this a lot easier to review!

Hi @bobular The flash phenomenon is, as I mentioned before, due to the two updates (regenerations) of `refinedMarkers` at `SemanticMarkers` component when clicking/unclicking a marker: caused by the changes of the two dependencies, `consolidatedMarkers`and `selectedMarkers`. It is a similar symptom with the previous popup flashing due to constant update of `consolidatedMarkers` that resulted in continuous update for `refinedMarkers`, thus I think that this is inevitable. 

As for the git stuff, I guess that for whatever reason, either my or your local branch was diverged from remove branch: it could be either this branch or its parent branch (`multiple-marker-selection-simplified`). So, I tried to re-update both branches, and now they seem to be alright: this branch now shows its changes only.

bobular (approved):
Thanks for sorting the github weirdness - nice to see just three files changed :-) 

I agree that changes in `selectedMarkers` are causing some of the flashing - I added a small commit to prevent every click on the map (outside markers) causing a flash in the popups a second or so later.

But that doesn't solve the wider problem. But let's merge this and just work on the parent branch.

Also, I still find that double-click is selecting a marker.  If you zoom all the way in to one marker, and then double click it, does it highlight for you? It does for me.

moontrip:
> Thanks for sorting the github weirdness - nice to see just three files changed :-)
> 
> I agree that changes in `selectedMarkers` are causing some of the flashing - I added a small commit to prevent every click on the map (outside markers) causing a flash in the popups a second or so later.
> 
> But that doesn't solve the wider problem. But let's merge this and just work on the parent branch.
> 
> Also, I still find that double-click is selecting a marker. If you zoom all the way in to one marker, and then double click it, does it highlight for you? It does for me.

@bobular My pleasure :) 
Honestly, I didn't test such an edge case, zooming all the way into one marker and trying double-click, and I cannot test it as the backend seems not to be ready or has an issue (500 server error). I will check it when it will be available.

Issue #642, opened 2023-11-14
ClinEpiDB - Strange behavior with Studies menu section collapsed state


dmfalke:
I noticed that the collapsed state of the sections in the Studies menu has a strange interaction when the filter is used. Perhaps it's related to React keys? See the screencast below:

[Screencast from 2023-11-14 15-30-17.webm](https://github.com/VEuPathDB/web-monorepo/assets/365139/907900f7-26ea-42be-9a79-a0e540fdb917)

aurreco-uga:
looks like when filtering it changes the section state: if closed, it opens, or if opened it closes it..

PR #646, opened 2023-11-15
Fix annoying react key errors in EDA site homepage

jernestmyers:
While working on https://github.com/VEuPathDB/web-monorepo/issues/642 I noticed that I introduced a key error with my `CollapsibleDetailsSection` work. I am addressing that particular key error and the rest of that ticket on a separate branch, but I went ahead and resolved the other two annoying key errors re: `News` and `CardList`.


dmfalke (approved):
Good catch!

PR #647, opened 2023-11-15
Render transcript filter in SRT for genes only

jernestmyers:
Resolves https://github.com/VEuPathDB/web-monorepo/pull/523#issuecomment-1812597494

**NOTE**: Pulling the data to check if it's a `transcript` feels brittle. Is there a more robust way?

### DIFF
--- a/packages/libs/web-common/src/components/reporters/SequenceFormFactory.jsx
+++ b/packages/libs/web-common/src/components/reporters/SequenceFormFactory.jsx
@@ -8,6 +8,8 @@ import { ComponentsList } from './SequenceFormElements';
 import * as ComponentUtils from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
 import * as ReporterUtils from '@veupathdb/wdk-client/lib/Views/ReporterForm/reporterUtils';
 import './ReporterForms.scss';
+import { useWdkService } from '@veupathdb/wdk-client/lib/Hooks/WdkServiceHook';
+import { getResultTypeDetails } from '@veupathdb/wdk-client/lib/Utils/WdkResult';
 
 const SINGLE_TRANSCRIPT_VIEW_FILTER_VALUE = {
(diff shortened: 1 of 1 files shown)


dmfalke:
> **NOTE**: Pulling the data to check if it's a `transcript` feels brittle. Is there a more robust way?

I thought the record class object included the available view filters, but I'm not seeing that information. So, this might be the only reliable way.

packages/libs/web-common/src/components/reporters/SequenceFormFactory.jsx
```diff
@@ -90,7 +90,11 @@ const createSequenceForm = (
       includeSubmit,
       viewFilters,
       updateViewFilters,
+      resultType,
     } = props;
+    const shouldRenderTranscriptFilter =
+      resultType?.step?.recordClassName &&
+      resultType?.step?.recordClassName === 'transcript';
```
dmfalke, line 97: You should use https://github.com/VEuPathDB/web-monorepo/blob/b6b240468f82ccc6fd2051b1ff2a72782efb7857/packages/libs/wdk-client/src/Utils/WdkResult.ts#L47-L48 to get the `recordClassName` of the `resultType`:

```ts
const resultTypeDetailsResult = useWdkService(wdkService => getResultTypeDetails(wdkService, resultType));
const shouldRenderTranscriptFilter = resultTypeDetailsResult.value?.recordClassName === 'transcript';
```

dmfalke (changes requested):
This looks good. See my suggestion.

packages/libs/web-common/src/components/reporters/SequenceFormFactory.jsx
jernestmyers: I'm getting an infinite loop with this approach. Any idea why?

dmfalke (approved):

PR #649, opened 2023-11-16
Deal with react-query errors and retry settings

### LINKED ISSUE #637: client keeps asking for the marker endpoint after a 500
(there might be an issue already with this bug)

when generating a 500 on teh marker endpoint , the client keeps asking for it. logs fill rapidly 

here is a current way to generate a 500:
https://qa.vectorbase.org/vectorbase.b66/app/workspace/maps/SEqWRUH/import

donut-date-stddev-bug

dmfalke:
Fixes #637 

This PR prevents useQuery from retrying queries when they fail. It also prevents useQuery from refetching queries when the browser window is refocused. Finally, the `isLoading` props for legend uses the `isFetching` property of a useQuery result.

### DIFF
--- a/packages/libs/eda/src/lib/map/MapVeuContainer.tsx
+++ b/packages/libs/eda/src/lib/map/MapVeuContainer.tsx
@@ -67,6 +67,10 @@ export function MapVeuContainer(mapVeuContainerProps: Props) {
         keepPreviousData: true,
         // We presume data will not go stale during the lifecycle of an application.
         staleTime: Infinity,
+        // Do not attempt to retry if an error is encountered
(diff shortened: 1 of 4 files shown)


dmfalke:
I should probably add some more details. 

The default value of retry is 3, which means an errored query will be retired 3 times before it's disabled.

There are 4 places that call the map-markers query, which explains why we see 4 requests (the first attempt, and the three retries.

bobular (approved):
LGTM. I didn't test though!

dmfalke:
> LGTM. I didn't test though!

I'll take it!

PR #653, opened 2023-11-17
add maaslin citation and link

### LINKED ISSUE #639: Volcano: add maaslin citation to the title
Add citation for maaslin2 to the title so folks know what we're using!

https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1009442

So the volcano title would be something like
"Differential abundance computed using Maaslin2 (Mallick et al., 2021) with default parameters."
Status: ready to implement!

asizemore:
Resolves #639 

Adds the Maaslin citation and link via some dangerous html.

Screen Shot 2023-11-17 at 7 18 18 AM

I checked the other apps that use plot subtitles (ranked abundance box and scatter) and those subtitles look good.


packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -164,7 +170,15 @@ function DifferentialAbundanceConfigDescriptionComponent({
 // Include available methods in this array.
 // 10/10/23 - decided to only release Maaslin for the first roll-out. DESeq is still available
 // and we're poised to release it in the future.
-const DIFFERENTIAL_ABUNDANCE_METHODS = ['Maaslin']; // + 'DESeq' in the future
+type DifferentialAbundanceMethodCitations = { Maaslin: string };
+const DIFFERENTIAL_ABUNDANCE_METHOD_CITATIONS: DifferentialAbundanceMethodCitations =
+  {
+    Maaslin:
+      '<a href="https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1009442">(Mallick et al., 2021)</a>',
+  }; // + deseq paper in the future
```
dmfalke, line 178: You can use `as const` to make typescript infer a `Record` type using the defined properties as string literals. So, no need for the type `DifferentialAbundanceMethodCitations`.
```suggestion
const DIFFERENTIAL_ABUNDANCE_METHOD_CITATIONS =
  {
    Maaslin:
      '<a href="https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1009442">(Mallick et al., 2021)</a>',
  } as const; // + deseq paper in the future
```

dmfalke (changes requested):
I hope you don't mind, but I took the privilege of reviewing this :)

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -164,7 +170,15 @@ function DifferentialAbundanceConfigDescriptionComponent({
 // Include available methods in this array.
 // 10/10/23 - decided to only release Maaslin for the first roll-out. DESeq is still available
 // and we're poised to release it in the future.
-const DIFFERENTIAL_ABUNDANCE_METHODS = ['Maaslin']; // + 'DESeq' in the future
+type DifferentialAbundanceMethodCitations = { Maaslin: string };
```
bobular, line 173: This could be `{ Maaslin: ReactNode }`

Though be aware that `undefined` is part of this type.

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -164,7 +170,15 @@ function DifferentialAbundanceConfigDescriptionComponent({
...
+const DIFFERENTIAL_ABUNDANCE_METHOD_CITATIONS: DifferentialAbundanceMethodCitations =
+  {
+    Maaslin:
+      '<a href="https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1009442">(Mallick et al., 2021)</a>',
```
bobular, line 177: Then this could be regular JSX, e.g. basically without the outer single quotes.

packages/libs/eda/src/lib/core/components/visualizations/OutputEntityTitle.tsx
```diff
@@ -26,10 +26,11 @@ export function OutputEntityTitle({ entity, outputSize, subtitle }: Props) {
           : 'No entity selected'}
       </span>
       {subtitle && (
-        <div className={cxSubtitle()} style={{ color: gray[700] }}>
-          <br />
-          {subtitle}
-        </div>
+        <div
```
bobular, line 29: Then I think you can just revert this change and go back to `{subtitle}` and get out of the Danger Zone...

packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
```diff
@@ -94,7 +94,13 @@ export const plugin: ComputationPlugin = {
     volcanoplot: volcanoPlotVisualization.withOptions({
       getPlotSubtitle(config) {
         if (DifferentialAbundanceConfig.is(config)) {
-          return `Differential abundance computed using ${config.differentialAbundanceMethod} with default parameters.`;
+          return `Differential abundance computed using ${
```
bobular, line 97: This would be regular JSX too, e.g.
```
return <span>Differential abundance computed using {method} {CITATIONS[method]} with default parameters</span>;
```

bobular (changes requested):
I think we could live a bit less dangerously!

If this type in layouts/types.ts
```
export interface TitleOptions {
  getPlotSubtitle?(config: unknown): string | undefined;
}
```
was changed so the function returns `ReactNode` then we could pass around strings or JSX fragments (`string` and `undefined` are a subset of ReactNode) 

Related suggestion are inline!

dmfalke:
> (string and undefined are a subset of ReactNode)

FYI `undefined` is not a subset of `ReactNode`, but `null` is!

bobular:
> > (string and undefined are a subset of ReactNode)
> 
> FYI `undefined` is not a subset of `ReactNode`, but `null` is!

My sources (@types/react/index.d.ts) say otherwise
```
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

dmfalke:
> > > (string and undefined are a subset of ReactNode)
> > 
> > 
> > FYI `undefined` is not a subset of `ReactNode`, but `null` is!
> 
> My sources (@types/react/index.d.ts) say otherwise
> 
> ```
>     type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
> ```

I stand corrected! I was confusing this with allowable return values of a React Component, which is `ReactElement | null`:

```ts
    interface FunctionComponent<P = {}> {
---->   (props: PropsWithChildren<P>, context?: any): ReactElement<any, any> | null;
        propTypes?: WeakValidationMap<P> | undefined;
        contextTypes?: ValidationMap<any> | undefined;
        defaultProps?: Partial<P> | undefined;
        displayName?: string | undefined;
    }
```

asizemore:
Thank you both and @dmfalke I'm glad you're here, too!

I incorporated the suggested changes and updated the `abundance` and `differentialabundance` plugins so that their subtitles are now react nodes. The `correlation` plugin i left as a string (so nice that ReactNode includes them all!). The plots with subtitles (abundance scatter, abundance box, volcano, bipartite network) all have nice looking subtitles now!

The only thing i didn't do that maybe needs some doing is 
> Though be aware that undefined is part of this type.

Do i need to incorporate `undefined` somehow? Or just be aware that this is a possibility?

dmfalke:
> Do i need to incorporate `undefined` somehow? Or just be aware that this is a possibility?

As long as you aren't using this as the return value of a component, it should be fine as-is.

dmfalke (approved):

bobular (approved):

bobular:
I haven't tested due to issues today but it LGTM!

PR #655, opened 2023-11-17
Fix breakage from variableId -> collectionId changes

### LINKED ISSUE #652: Mbio: Ranked abundance scatter shows overlay input but it should not
Current behavior (from qa. any study):

Screen Shot 2023-11-17 at 7 04 19 AM

Desired behavior (from live):
Screen Shot 2023-11-17 at 7 05 52 AM

In the ranked abundance scatter, the y axis and the overlay variable are fixed because they're provided by the computation. However, on qa we're seeing that the user still has the option to pick an overlay variable. When they do, the plot regenerates but it does not take that user-selected var into account. Said another way, we're basically showing an option that does nothing and creates confusion by conflating a user-selected input with a fixed input (generated from the computation).

I noticed that the other compute apps seems to have the correct inputs fixed, so i think the problem is specific to the scatter in ranked abundance.

jernestmyers:
Resolve #652


asizemore (approved):
Looks great! Thanks for getting to this so quickly!
The legend and complete cases table also look good - great catch. In fact even the vizs are working faster with this branch so idk how you did that but thank you!

PR #656, opened 2023-11-17
Studies menu filter bug

### LINKED ISSUE #648: Studies dropdown menu: My studies and Curated studies sections always should be open (remove ability to close)
On qa.restricted.clinepi, the studies dropdown looks like this, with `My studies` closed initially and `Curated studies` opened initially
image

If I use the search bar (ex: search for avenir), this causes `My studies` to disappear and `Curated studies` to close (such that I cannot see the study I searched for without opening it up)
image

Please change the functionality of this so that `My studies` and `Curated studies` are always open (there is no need to close these sections).
If I search for a study, please keep the `My studies` and `Curated studies` headers, but just show the studies that meet the search below the headers

jernestmyers:
Resolves #648 and #642 

A screencast of the new behavior can be found [in this comment](https://github.com/VEuPathDB/web-monorepo/issues/648#issuecomment-1814608485).

An outline, approved by EDA UX folks, of the new behavior is as follows:
- the sections for `My studies` and `Curated studies` are expanded by default
- collapsible sections only exist if/when a user has "DIY" studies
- searching will expand both sections, even if the user has collapsed them
- if a search filters away all studies from a section, that section will be removed (but the section header will persist for the remaining section)


packages/libs/wdk-client/src/Components/Display/CollapsibleDetailsSection.tsx
```diff
@@ -1,17 +1,24 @@
-import React, { ReactNode, useState } from 'react';
+import React, { ReactNode, useEffect, useState } from 'react';
 
 interface Props {
   summary: ReactNode;
   collapsibleDetails: ReactNode;
   initialShowDetailsState?: boolean;
+  expandDueToFiltering?: boolean;
```
dmfalke, line 7: To be honest, I'm not crazy about this new prop. It might be time to consider making this a fully controlled component, and remove (e.g., "lift") the `showDetails` state. I'm not sure how much work this would entail. (How many different places would need to be updated to track this state?)

dmfalke (commented):
I think the behavior is good. Please take a look at my comment about "lifting" the state of the `CollapsibleDetailSection` component.

packages/libs/wdk-client/src/Components/Display/CollapsibleDetailsSection.tsx
jernestmyers: Yeah, great point. I'll push a change shortly. Shouldn't be a big load given that it's a new component being used in only 4 instances: two sections in clinepi studies menu, two in mbio.

packages/libs/wdk-client/src/Components/Display/CollapsibleDetailsSection.tsx
```diff
@@ -1,17 +1,18 @@
-import React, { ReactNode, useState } from 'react';
+import React, { ReactNode } from 'react';
 
 interface Props {
   summary: ReactNode;
   collapsibleDetails: ReactNode;
-  initialShowDetailsState?: boolean;
+  showDetails: boolean;
+  setShowDetails: React.Dispatch<React.SetStateAction<boolean>>;
```
dmfalke, line 7: Can we change this to the following? I feel like the current type is an implementation detail.

```suggestion
  setShowDetails: (showDetails: boolean) => void;
```

packages/sites/clinepi-site/webapp/js/client/component-wrappers/SiteHeader.jsx
```diff
@@ -26,10 +26,37 @@ export default function SiteHeaderWrapper() {
...
+    // for now, we default to each studies section being open and the useEffect
+    // ensures that the sections open when searching
+    const [expandUserStudies, setExpandUserStudies] = useState(true);
+    const [expandCuratedStudies, setExpandCuratedStudies] = useState(true);
+    useEffect(() => {
+      if (searchTerm && searchTerm.length > 0) {
+        setExpandUserStudies(true);
+        setExpandCuratedStudies(true);
+      }
+    }, [searchTerm, setExpandUserStudies, setExpandCuratedStudies]);
```
dmfalke, line 38: Instead of using a `useEffect` hook here, you can make a callback function that calls `setSearchTerm` and both `setExpand*Studies` functions. I think it's a little bit clearer, and it will also remove a render cascade: calling a `setState` setter in a `useEffect` callback, due to a different `useState` value changing, will incur two renders, since the `useEffect` callback will be called _after_ the first state has been updated.

dmfalke (commented):
Very nice! I just have a couple more suggestions.

PR #662, opened 2023-11-21
temporarily turn off correlation

### LINKED ISSUE #619: Mbio: Temporarily turn off correlation app for b66 release
We don't want to release the correlation app just yet, so let's just turn it off for the release. We can turn it back on afterwards so we can continue development!

Iirc, just removing the correlation line from the index file will do the trick 😄

asizemore:
Resolves #619 

Turns off the correlation app so that it will not get released with b66.

Can be reverted after the release!

From PR (Correlation should show up between diff abund and Distributions, but it's not there!)
Screen Shot 2023-11-21 at 6 04 03 AM


dmfalke (approved):

asizemore:
Thank you! I'll wait to merge until we're just about to do the monorepo release that will be for b66, just so that the correlation app can stay turned on as long as possible.

PR #664, opened 2023-11-21
update default effect size and var names

### LINKED ISSUE #663: Mbio: Volcano update default effect size threshold to 1
Discussed in the mbio channel on nov 16, we were asked to please change the default effect size threshold to +/- 1 (currently it is +/- 2). 

Context: Originally we had 2 as the cutoff because that is common practice for deseq. When we changed from computing differential abundance with deseq to maaslin, the x axis value changed from log2 fold change to the maaslin-specific model coefficient which we can generally call their computed "effect size". A maaslin effect size of 1 is approximately a log2 fold change of 2. 
The above means that for now just changing the default threshold to 1 will be sufficient. But it does suggest that in the future when we add new methods, each method might want to set its own threshold.

asizemore:
Resolves #663 

Updated the default effect size threshold to 1 (effectively +/- 1). Also renamed the var to make the concept more general. "Effect size" is the general term for that measure, whereas "model coefficient" (from maaslin) and "log2FoldChange" (from deseq) are specific to the methods.

Here's a pic
Screen Shot 2023-11-21 at 9 57 34 AM

Note - any testing needs to be done with a cached completed compute bc they're failing right now. The above was with diabimmune

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
@@ -63,8 +63,8 @@ import { truncationConfig } from '../../../utils/truncation-config-utils';
 // use Notification for truncation warning message
 import Notification from '@veupathdb/components/lib/components/widgets//Notification';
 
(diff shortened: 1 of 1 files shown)


jernestmyers (approved):
🟢

PR #665, opened 2023-11-21
Prototype transcript filter in ortho/paralogs tables in gene page

jernestmyers:


dmfalke (approved):
This looks pretty good! Performance isn't great when the table has several thousand rows, but that's more of a rendering issue (paging or windowing would help tremendously).

dmfalke:
FYI, I will merge this asap

PR #671, opened 2023-11-23
Improve UX for bubble marker proportion configuration when filters are active on the variable

### LINKED ISSUE #670: Map - improve UX of bubble marker categorical proportion filter sensitivity
Continues work on #582 which was merged in a suboptimal state for b66.

There's a problem with values disappearing from the numerator and denominator dropdowns when a filter is active.

Here we will fix it by showing the filtered out values as disabled values in the dropdown.

The user will probably be prompted to review each dropdown to confirm the current active selections.

The markers (and legends) also need to be disabled when the proportion configuration becomes invalid.

bobular:
Fixes #670 

- [x] show "filtered away" values as greyed out in num/denom dropdowns
- [x] tidy up validation logic by putting validation check into `useOverlayConfig`
- [x] don't show markers when the proportion is invalid
- [x] show something in the legends when no markers are shown for this reason. Note that I will be adding a ticket to add a :gear: (gear) icon to the legends, which will open up the active marker config.
- [ ] one question remains... is the checked greyed out checkbox a problem? (it cannot be deselected)
      - do we want to modify ValuePicker/SelectList so that disabled items have greyed out labels but active deselectable checkboxes (but not reselectable?)
- [ ] use `ReactNode` for `invalidProportionText` and make it two paragraphs? Oh, but `PluginError` can't handle that at present. I feel this should be a separate ticket?

Screenshot from 2023-11-23 18-02-39

### DIFF
--- a/packages/libs/eda/src/lib/core/types/study.ts
+++ b/packages/libs/eda/src/lib/core/types/study.ts
@@ -121,6 +121,10 @@ const Variable_Base = t.intersection([
   }),
   t.partial({
     vocabulary: t.array(t.string),
+    // filter-sensitive versions of StudyEntity data have modified
+    // vocabularies - this is where we keep a copy of the original
+    // (note that this never comes from the back end)
+    fullVocabulary: t.array(t.string),
   }),
 ]);
 
--- a/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
@@ -34,10 +34,7 @@ import {
 } from '../../MarkerConfiguration/MapTypeConfigurationMenu';
 import { BubbleMarkerIcon } from '../../MarkerConfiguration/icons';
 import { useStandaloneVizPlugins } from '../../hooks/standaloneVizPlugins';
-import {
-  getDefaultBubbleOverlayConfig,
-  validateProportionValues,
-} from '../../utils/defaultOverlayConfig';
+import { getDefaultBubbleOverlayConfig } from '../../utils/defaultOverlayConfig';
 import {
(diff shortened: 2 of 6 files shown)


packages/libs/eda/src/lib/core/types/study.ts
```diff
@@ -121,6 +121,10 @@ const Variable_Base = t.intersection([
   }),
   t.partial({
     vocabulary: t.array(t.string),
+    // filter-sensitive versions of StudyEntity data have modified
+    // vocabularies - this is where we keep a copy of the original
+    // (note that this never comes from the back end)
+    fullVocabulary: t.array(t.string),
```
bobular, line 127: This is probably the most contentious change!

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
```diff
@@ -100,27 +100,28 @@ function BubbleMapConfigurationPanel(props: MapTypeConfigPanelProps) {
 
   // If the variable or filters have changed on the active marker config
   // get the default overlay config.
-  const activeOverlayConfig = useOverlayConfig({
+  const { overlayConfig, isValidProportion } = useOverlayConfig({
```
bobular, line 103: **active**OverlayConfig didn't seem absolutely necessary so I didn't bother renaming the new object return value

dmfalke: That's a holdover from an earlier version of this code base

dmfalke (approved):
I think this looks great! We might want to have a UX review at some point, for refinement. But, I think this is merge-worthy. Maybe wait until Wednesday or Thursday, since we're still trickling in updates for b66 (unless you don't mind this going into b66).

Issue #611, opened 2023-11-08
Map - add handling of date variables for bubble markers


bobular:
Don't forget to remove the temporary disabling here: https://github.com/VEuPathDB/web-monorepo/pull/495

bobular:
Done via https://github.com/VEuPathDB/web-monorepo/pull/672 - not sure why it wasn't linked here

PR #672, opened 2023-11-24
#611 bubble marker date handling

bobular:
The deal with [this change](https://github.com/VEuPathDB/web-monorepo/pull/672/commits/bae036808f9d94c25abd7ddca68d1ade042eb182) is that previously, `useMarkerData` was doing two things inside its `useQuery` `queryFn`:
1. fetching the raw data
2. mapping it to fully fledged marker data using diameter/color mapping functions

In order to prevent inconsistencies in marker sizing/coloring, it was [previously necessary](https://github.com/VEuPathDB/web-monorepo/commit/e02a4b4d17fbad5b4917ab1db277e4d0e5eb409c) to use legend data as a "dependency" for the `useQuery`, but this meant repeating the raw data request unnecessarily.
So I have split things out into separate hooks and then combine them in a new `useMarkerData` that spoofs a `useQuery` return value via a regular `useMemo`. It deserves some critical review I think!

Another possible contentious point is the addition of `valueType` to `aggregationConfig`. This saves a lot of fetching the full `overlayVariable` in various places and/or passing a lot of values around functions, but has the disadvantage of diverging from the back end type definitions, and therefore there are two `omit()`s to remove it before bubble-related requests are made.


dmfalke (approved):
This is a nice improvement. I did struggle with how to handle a query depending on another, while still propagating the query results. I couldn't really find any convention for this. It's as if we want a utility that can take 2+ queries and merge them. I guess https://tanstack.com/query/latest/docs/react/reference/useQueries sort of does that.

dmfalke:
> a query depending on another

Technically, these queries are independent. But, we want to combine the results. I think I'm overthinking things :slightly_smiling_face:

packages/libs/eda/src/lib/core/api/DataClient/types.ts
```diff
@@ -852,10 +852,18 @@ export const BubbleOverlayConfig = type({
...
+      type({
+        overlayType: literal('continuous'),
+        aggregator: keyof({ mean: null, median: null }),
+      }),
+      partial({
+        // optional for backwards compatibility with user db
+        // and because it must not be sent to the back end
+        // in StandaloneMapBubblesRequestParams and StandaloneMapBubblesLegendRequestParams
+        valueType: keyof({ number: null, date: null }),
+      }),
```
bobular, line 865: So you didn't mind this? @dmfalke

dmfalke: Sorry, I didn't give this much thought. If we don't care about `valueType`, then we can remove it. `io-ts` will not throw if there are excess properties.

bobular: Thanks. I've removed it from the io-ts type and made the appropriate adjustments. Will merge now.

PR #674, opened 2023-11-27
Fix unsync'ed "Currently viewing" marker in All Analyses table.

### LINKED ISSUE #515: Map - My Analyses "currently viewing" seems to get out of date
I switch to another analysis but the My Analyses table doesn't reflect this with the "(currently viewing)" text.
Probably just a missing hook dependency?

image

bobular:
Fixes #515 

Added the missing dependency.

However, a related issue is that editing the active analysis' name in either the table or the header does not cause the other one to update.  I looked into it briefly but didn't find an obvious answer.  Should we try to fix that too?


dmfalke (approved):
LGTM!

PR #682, opened 2023-11-28
Add files via upload

danicahelb:


dmfalke:
A better place for these might be in https://github.com/VEuPathDB/StaticContent/tree/master/assets/images.

PR #692, opened 2023-12-02
Fix #687 wrong output entity in map header stats

### LINKED ISSUE #687: Map - header counts not correct when marker variable is a study variable
The map is showing collection sites when a study variable is used as overlay.  Those are the numbers in the markers. It can't show study counts unless we are happy with double counting.

So the numbers in the header are wrong and confusing - particularly the View: count (which is collection sites).

image

Despite the new-look header counts for the megastudy coming soon (ticket coming very soon, promise...) this should be fixed in case any non-megastudy datasets have a non-root geo-entity.

bobular:
Fixes #687 

Bubble uses `useCommonData` and donut/bar propagate the `outputEntityId` through its `useMarkerData` hooks. I can standardise (to `useCommonData`?) if it's important!


dmfalke (approved):
I think this is great. Should bubbles also propagate the output entity, and such, like the other markers?

bobular:
Thanks @dmfalke 

It would be easier to use the `useCommonData` approach in all cases actually. 

Did you have a reason for suggesting the alternative?

Issue #624, opened 2023-11-09
Map - Legend title should have fixed width and wrap or use ellipses


bobular:
Too much space wasted here!

![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/dd58568d-438d-466a-8049-37984070f7ca)

PR #699, opened 2023-12-04
ellipsis of legend title and tooltip for SAM and minWidth

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/624

Similar to EDA legend, the panel title at Draggable panel now has ellipsis (25 characters like EDA) and mouseover tooltip (screenshot). Also, minWidth is set to the Draggable panel so that it cannot be too narrow for bubble marker: set to be 120px based on Country variable and considering the width of scale bar.

SAM legend title ellipsis01


bobular (approved):
This great. Tested and code looks good.

PR #701, opened 2023-12-05
Add legends for bipartite network

### LINKED ISSUE #668: Add legend for bipartite network

asizemore:
Resolves #668 

Adds two legends to the bipartite network visualization. One legend for the stroke/link width and one for the link color. The link width legend required an update to PlotLegend so that it could handle the varying stroke widths.

I added to legends instead of putting them all in one because i think it's more clear that these are two dimensions of data that we're conveying when the two data representation dimensions are separated. 

I gave the css a go for the multi-legend container, but would be grateful for some fierce looks there especially!

Screen Shot 2023-12-05 at 10 08 05 AM


packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
```diff
@@ -237,7 +261,72 @@ function BipartiteNetworkViz(props: VisualizationProps<Options>) {
...
+  const colorLegendItems: LegendItemsProps[] = [
+    {
+      label: 'Positive correlation',
+      marker: 'line',
+      markerColor: twoColorPalette[1],
+      hasData: true,
+      lineThickness: '3px',
+    },
+    {
+      label: 'Negetive correlation',
```
bobular, line 304:
```suggestion
      label: 'Negative correlation',
```

packages/libs/components/src/components/plotControls/PlotListLegend.tsx
```diff
@@ -23,6 +23,7 @@ export interface LegendItemsProps {
   group?: number;
   rank?: number;
   hideFromLegend?: boolean;
+  lineThickness?: string;
```
bobular, line 26: // add comment to say what kind of string, e.g. '0.15em' is the default

packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
```diff
@@ -135,6 +147,17 @@ function BipartiteNetworkViz(props: VisualizationProps<Options>) {
     ])
   );
 
+  // Adjust link stroke widths so that it spans an appropriate range for this viz.
+  const dataStrokeWidths =
+    data.value?.bipartitenetwork.data.links.map((link) =>
+      Number(link.strokeWidth)
```
bobular, line 153: Curious what the deal is with strokeWidth being a string. I know that's the API/io-ts definition.

I guess it's nothing to do with dates.

Do we need to protect against non-numbers? Are there expected non-number values? Or maybe just a comment. Just worried about it evaluating to `NaN` at runtime at some point.

packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
```diff
@@ -135,6 +147,17 @@ function BipartiteNetworkViz(props: VisualizationProps<Options>) {
     ])
   );
 
+  // Adjust link stroke widths so that it spans an appropriate range for this viz.
```
bobular, line 150: Pedantically we're not adjusting here, we're making an adjustor function

bobular (changes requested):
This is great.

I think this (see below) needs extra verbiage - needs to say "(absolute?) correlation coefficient" somewhere. We should tame the decimals too. These are the only two things for the "requested changes". Everything else looks great with a few minor suggestions.

image

packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
asizemore: Its a good question. We've been generally just converting numbers to strings for transferring the data lately in order to retain precision. But in this specific case the strokeWidth should always be a number. I left a comment!

I'm remembering you made a comment in a separate PR that instead of strokeWidth it should really be linkWeight. I think i like that and need to check out how wide reaching a change that would require. But hopefully that would clarify this spot as well!

asizemore:
Legend title updated (in _both_ correlation apps this time!) and decimals tamed! Other comments also addressed!

Screen Shot 2023-12-11 at 2 47 38 PM

bobular (approved):

bobular (approved):
Looks great. I wouldn't bother with the `strokeWidth` back end API change but leave a note somewhere that it might be useful to do, one day, when we have nothing on!

Issue #484, opened 2023-09-07
Donut/barplot marker categorical config table "No Results" placeholder should be spinner most of the time


bobular:
You can see this when switching from 

Show counts for:
Filtered -> Visible

It may also appear other times.

The spinner is our normal loading indicator. The No Results should be reserved for when a response has come back empty.

PR #703, opened 2023-12-06
add spinnner for categorical table at SAM

moontrip:
This addresses #484. 

Spinner component was also modified to accommodate a prop of overriding styles: positioning etc.

A sample video capture is included: since it is hard to capture, I just used hard refresh (Ctrl + F5) to reload data, so white background/screen is shown for a while.

https://github.com/VEuPathDB/web-monorepo/assets/12802305/1b89f6b4-8446-4d13-8fac-1e58fac390b0


bobular:
Hi @moontrip 

I've been doing a bit of testing and it looks like there is still the "no results" table when switching from categorical to continuous variables, e.g. in the video below


https://github.com/VEuPathDB/web-monorepo/assets/308639/b75ca3dc-9691-4cee-a1ba-91ea677b2ae3

bobular (commented):
There's a general UX issue here. Maybe @dmfalke would like to chime in.

Do we want spinners on every plot and table that changes? And do we want that between every variable change? We can do that by changing the `isLoading` to `isFetching` in DK's PR here.  But then we'd need to add a spinner to the marker preview graphic too.

That may be too many spinners. People might get hypnotised...

Alternatively we could just use the spinners for the first new table render (the `isLoading` approach taken here).

It would be nice to have some feedback after changing the variable though. Maybe a small spinner to the right of the variable picker?

moontrip:
> Hi @moontrip
> 
> I've been doing a bit of testing and it looks like there is still the "no results" table when switching from categorical to continuous variables, e.g. in the video below

Hi @bobular For the issue you found, I firstly tried to use the approach you mentioned at Slack, but it did not work out. Thus, I examined it more carefully, and finally found the culprit. It is because of the hook, `useDistributionOverlayConfig()` that also uses react-query, that renders additional delay beyond `useCategoricalValues()` hook. Due to this reason, the `isAllCategoricalValuesLoading` prop becomes false for a moment before overlayConfiguration is loaded/fetched: in this case, `mesa` table is called due to the logic I used at the `CategoricalMarkerConfigurationTable` component.

Thus, a solution I made is to additionally check `overlayConfiguration.isFetching` at `isAllCategoricalValuesLoading` prop, which seems to work properly from my tests.

bobular:
> Maybe everything below the variable selector in the config panel should be a spinner until all of the data is loaded?

Do you mean the _first_ data is loaded?  Or do you mean spinners for every fetch of new data for new variables etc?

bobular (approved):
Thanks - that solved it.

I'll make a [new ticket](https://github.com/VEuPathDB/web-monorepo/issues/726) for a general overhaul of update/spinner behaviour in this panel. It's not very consistent.

dmfalke:
> > Maybe everything below the variable selector in the config panel should be a spinner until all of the data is loaded?
> 
> Do you mean the _first_ data is loaded? Or do you mean spinners for every fetch of new data for new variables etc?

I was thinking we should show it every time a new variable is selected, and we would show it until both the distribution and summary marker data are fetched.

bobular:
On Thu, Dec 14, 2023 at 3:46 AM Dave Falke ***@***.***> wrote:

> Maybe everything below the variable selector in the config panel should be
> a spinner until all of the data is loaded?
>
> Do you mean the *first* data is loaded? Or do you mean spinners for every
> fetch of new data for new variables etc?
>
> I was thinking we should show it every time a new variable is selected,
> and we would show it until both the distribution and summary marker data
> are fetched.
>

Thanks, that's almost what we have now in this PR.

The summary marker is silently updated, however.

I made a new ticket to standardise the spinners.

moontrip:
> Thanks - that solved it.
> 
> I'll make a [new ticket](https://github.com/VEuPathDB/web-monorepo/issues/726) for a general overhaul of update/spinner behaviour in this panel. It's not very consistent.

Thank you for your review and test @bobular 👍 Since we have a new ticket for the behavior of the spinner, I will merge this.

PR #706, opened 2023-12-06
EDA variable collection selector

dmfalke:
### Description

This PR introduces a new `VariableCollectionSelectList` component. It includes some factoring to provide more flexibility in how collections for a study are presented. It also includes some changes to improve type safety.

### Details

An overview of the changes are:
- Update `SingleSelect` to allow items to be grouped
- Add `VariableCollectionSelectList`, which uses the new group feature of `SingleSelect`
- Change compute config type to `Partial` in many cases, to better reflect how users gradually choose values for the config.
- Change default compute config from `undefined` to `{}` in many cases.
- Add helper to convert an `io-ts` `Partial` codec to a `Type` codec
- Change the compute plugin function `isComputeValid` to `isComputeComplete` to better reflect user flow.
- Residual updates, based on changes above.

### Screenshots

image

---

**Question:** Should `VariableCollectionSelectList` should be renamed to `VariableCollectionSingleSelect`?


packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -152,11 +155,14 @@ export function AbundanceConfiguration(props: ComputationConfigProps) {
   } = props;
   const studyMetadata = useStudyMetadata();
   // Include known collection variables in this array.
-  const collections = useVariableCollections(studyMetadata.rootEntity);
+  const collections = useVariableCollections(
+    studyMetadata.rootEntity,
+    isNotAbsoluteAbundanceVariableCollection
```
d-callan, line 160: :100:

packages/libs/eda/src/lib/core/types/visualization.ts
```diff
@@ -52,36 +52,39 @@ export const Visualization = intersection([
...
   computationId: string,
+  displayName: t.union([t.string, t.undefined]),
   descriptor: type({
     type: string,
     configuration: unknown,
   }),
   visualizations: array(Visualization),
 });
 
+export function makeComputationWithConfigDecoder<T>(
```
d-callan, line 76: i think im just not following this generally.. what is this doing?

packages/libs/eda/src/lib/map/analysis/hooks/standaloneVizPlugins.ts
```diff
@@ -101,7 +101,7 @@ export function useStandaloneVizPlugins({
 
     const pluginBasics = {
       configurationComponent: ZeroConfigWithButton,
-      isConfigurationValid: t.undefined.is,
+      isConfigurationComplete: t.undefined.is,
```
d-callan, line 104: how does this relate to our decision to not have undefined as default compute config (in favor of `{}`) in each of the apps?

d-callan: is a standalone viz one without a compute? not sure im following..

packages/libs/eda/src/lib/map/analysis/hooks/standaloneVizPlugins.ts
```diff
@@ -154,5 +154,5 @@ export function useStandaloneVizPlugins({
         },
       },
     };
-  }, [selectedOverlayConfig]);
+  }, [overlayHelp, selectedOverlayConfig]);
```
d-callan, line 157: what is overlayHelp?

d-callan (approved):
i didnt test this but i think its beautiful, and would be happy to if needed

asizemore:
Wow it looks sooo much nicer! Thanks Dave!

I've only tested so far but have two comments:
1. The following screenshot is from diabimmune. Diabimmune has two assay enetites, so we can see the select list gets long. In the next year or so, it's quite possible we have a study with three assay entities, which would make the list even longer. Have you thought about how this list would scroll? It wasn't obvious to me because we'd want a user to see the headings, though i guess having them be collapsible in the future would be one solution? Not crucial for now just wanted to ask!
Screen Shot 2023-12-11 at 6 47 58 AM

2. Looks like the correlation assays didn't get the updated select list. All the other apps looked good!


And to your question about naming, i prefer the `VariableCollectionSingleSelect` because that makes the relationship with the `SingleSelect` component more clear.

packages/libs/coreui/src/components/inputs/SingleSelect.tsx
```diff
@@ -89,18 +99,75 @@ export default function SingleSelect<T>({
...
+  value: T;
+  indexOffset: number;
+  indexOfFocusedElement: number;
+}
+function OptionGroup<T>(props: OptionGroupProps<T>) {
+  const {
+    itemGroup,
+    onKeyDown,
+    onSelect,
+    value,
```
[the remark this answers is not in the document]
dmfalke: This is the value of the selected item. I suppose `selectedValue` could be more clear. The current name is meant to reflect web APIs, which typically just use `value`. It is also the same name we use for `SelectList`.

packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -21,21 +21,24 @@ import { useMemo } from 'react';
...
 
 export type AbundanceConfig = t.TypeOf<typeof AbundanceConfig>;
 // eslint-disable-next-line @typescript-eslint/no-redeclare
-export const AbundanceConfig = t.type({
+export const AbundanceConfig = t.partial({
   collectionVariable: VariableCollectionDescriptor,
   rankingMethod: t.string,
 });
 
+const CompleteAbundanceConfig = partialToCompleteCodec(AbundanceConfig);
```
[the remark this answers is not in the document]
dmfalke: > are we saying that when we declare some type, it should always reflect possible states rather than the desired state? like is the reason were unhappy w how things were bc a user might simply not have set a rankingMethod yet, even though one is required for the compute?


Essentially, yes. Otherwise we end up with intermediate states that are not reflected by our types. I think we were getting around this with a type cast, and I kept running into runtime errors.

I think this bears more discussion, tbh. There are other ways to approach this, which we should enumerate and evaluate. We should also be consistent with visualizations, in which we don't explicitly declare a "complete" state. Instead we just check that the props we want are defined in the relevant places.

I guess a question to ask is, is this too clever? It also doesn't cover every case (see differential abundance, where we have to drill further into the `comparator` object).

packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -21,21 +21,24 @@ import { useMemo } from 'react';
...
+
 export const plugin: ComputationPlugin = {
   configurationComponent: AbundanceConfiguration,
   configurationDescriptionComponent: AbundanceConfigDescriptionComponent,
-  createDefaultConfiguration: () => undefined,
-  isConfigurationValid: AbundanceConfig.is,
+  createDefaultConfiguration: () => ({}),
```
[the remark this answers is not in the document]
dmfalke: `undefined` doesn't confirm to the config type.

packages/libs/eda/src/lib/map/analysis/hooks/standaloneVizPlugins.ts
dmfalke: Yes, these standalone vizs use a "compute" without config

packages/libs/eda/src/lib/core/components/variableSelectors/VariableCollectionSelectList.tsx
```diff
@@ -0,0 +1,72 @@
...
+  collectionPredicate?: (collection: CollectionVariableTreeNode) => boolean;
+  value?: VariableCollectionDescriptor;
+  onSelect: (value?: VariableCollectionDescriptor) => void;
+}
+
+export function VariableCollectionSelectList(props: Props) {
+  const { collectionPredicate, onSelect, value } = props;
+  const entities = useStudyEntities();
+
+  const items = useMemo(() => {
```
asizemore, line 20: Why get the collections here within the component, instead of passing in an array of collections? In the apps it looks like we already have `collections`, which is a filtered list of collections, no?

dmfalke: I went back-and-forth on this, and ultimately decided on this, to be more consistent with other variable selectors. Generally speaking, it's probably better to pass things in, rather than look them up on React Context (hidden dependencies, and all).

At this point, it appears the only reason we are getting collections in the apps is to ensure the study has collections. Does mbio have any such studies?

dmfalke:
> Wow it looks sooo much nicer! Thanks Dave!
> 
> I've only tested so far but have two comments:
> 
>     1. The following screenshot is from diabimmune. Diabimmune has two assay enetites, so we can see the select list gets long. In the next year or so, it's quite possible we have a study with three assay entities, which would make the list even longer. Have you thought about how this list would scroll? It wasn't obvious to me because we'd want a user to see the headings, though i guess having them be collapsible in the future would be one solution? Not crucial for now just wanted to ask!
> 
> 
> Screen Shot 2023-12-11 at 6 47 58 AM

I was assuming the entire list would scroll. I will add a story to see what it looks like. I'm hesitant to go w/ collapsible headings (I don't really like them in most cases). But a search box might be a good idea, if the list gets very long?

> 
>     2. Looks like the correlation assays didn't get the updated select list. All the other apps looked good!

Oops! I will fix.

> 
> 
> And to your question about naming, i prefer the `VariableCollectionSingleSelect` because that makes the relationship with the `SingleSelect` component more clear.

Ah, great. I will make the change!

dmfalke:
I just pushed some changes to `SelectList`:
1. Group labels are sticky when scrolling
2. Fixed an issue with focus indexing
3. Change focus styling to use a neutral background (the dotted outline look glitchy to me)

[Screencast from 2023-12-11 12-34-36.webm](https://github.com/VEuPathDB/web-monorepo/assets/365139/93eff696-ec63-44ce-a713-808fd915b0a7)

dmfalke:
The selector component has been renamed, and correlation apps are using the new selector component.

packages/libs/eda/src/lib/core/components/variableSelectors/VariableCollectionSelectList.tsx
asizemore: Ahh got it. Makes sense to me!

Mbio has no such studies. All the guts of mbio data are always in a collection so even in the future its very very unlikely we'll ever load data without a collection!

asizemore (approved):
looks great!! Thanks @dmfalke !

packages/libs/eda/src/lib/core/components/variableSelectors/VariableCollectionSelectList.tsx
dmfalke: Cool! Do you think it's okay to remove the current check we're doing?

The UX of the current code probably isn't great: if a study does not have any collections, the entire viz tab will show an error message. If we remove the check, instead they will see an empty `SingleSelect` list, with no way to run the compute, which doesn't seem _too_ bad.

asizemore: Yeah i agree, the ux after removing the check sounds better. It'd be disruptive enough to (hopefully) make a user contact us about it but not so disruptive that they would stop doing anything at all on the site!

I made an [issue](https://github.com/VEuPathDB/web-monorepo/issues/723) to improve the no-collection ux. I know i need to do t[his related one for correlation soon](https://github.com/VEuPathDB/EdaNewIssues/issues/731), so if it turns out i see an easy way to do both i will!

asizemore: oh i just read your final comment that you already removed the check. Awesome thank you!

PR #707, opened 2023-12-06
Add ability for user to threshold edges by correlation value or p value

### LINKED ISSUE #704: Bipartite network: let user set thresholds for edges
The backend is ready for the frontend to give it pvalue and correlation magnitude thresholds. Let's add them in as viz controls!

asizemore:
Resolves #704 

Allows a user to set the correlation coefficient threshold as well as the p-value threshold. Both of these are sent to the data service who will use these values to remove edges (and nodes) in the network. 

https://github.com/VEuPathDB/web-monorepo/assets/11710234/43795e23-d629-49ae-b333-94af8e16f790


dmfalke:
Oh, I have a pretty naive question. Is it completely necessary to make a new request based on p-value? Is that a static value that can be included with the data and then filtered on the front end?

asizemore:
Good question! In fact both the correlation and p value are static. Both could be filtered on in the frontend. We picked the backend because the amount of data we could reduce sending to the frontend is huge (hundreds of nodes and thousands of edges), but also because the correlation and pvalue filtering affect any network metrics we calculate (we're planning to calculate degree pretty soon). We had assumed network metrics would be more quickly calculated on the backend, which is why then it seemed like the correlation and pvalue filters should also go to the backend.

Does that make sense? Is there a better way to do it?

dmfalke:
That makes perfect sense. Thanks for the explanation!

PR #710, opened 2023-12-06
Fix two minor typos in page that tells you you've opened a non-map study in the map

bobular:


dmfalke (approved):
Thanks for catching these typos! :grimacing:

PR #719, opened 2023-12-08
Improve map error handling and fix some logic related to marker config and legend updates

### LINKED ISSUE #708: Map - 'Could not get overlay config' error for continuous overlay vars in no-data situations
Megastudy

1. select a time range in slider
2. select a different range in subset interface
3. select date variable for donut or bar (and maybe bubble) marker

The error comes from the following in `packages/libs/eda/src/lib/map/analysis/mapTypes/shared.ts`
```
  const overlayConfigResult = useDistributionOverlayConfig({
    studyId,
    filters,
    binningMethod,
    overlayVariableDescriptor: selectedVariable,
    selectedValues,
  });

  if (overlayConfigResult.error) {
    throw new Error('Could not get overlay config');
  }
```

Which is causing a 500 on the `/eda/filter-aware-metadata/continuous-variable` endpoint with the error message:
"Could not generate continuous variable metadata" from the request

```
{
	"config": {
		"metadata": [
			"binRanges"
		],
		"variable": {
			"entityId": "OBI_0000659",
			"variableId": "EUPATH_0043256"
		}
	},
	"filters": [
		{
			"entityId": "OBI_0000659",
			"max": "2019-12-31T00:00:00.000Z",
			"min": "2014-01-01T00:00:00Z",
			"type": "dateRange",
			"variableId": "EUPATH_0043256"
		},
		{
			"entityId": "OBI_0000659",
			"max": "2001-03-17T00:00:00Z",
			"min": "1994-09-30T00:00:00Z",
			"type": "dateRange",
			"variableId": "EUPATH_0043256"
		}
	],
	"studyId": "VBP_MEGA"
}
```

### LINKED ISSUE #713: Map - donut/bar marker continuous marker config preview (and legend) is not using global viewport
To reproduce, just keep an eye on the marker config preview as you move around the map. It should not change, but it does. There's no spinner (a bug/oversight perhaps?) so keep watching carefully!

It is definitely not showing "all filtered data"!

A likely fix is as simple as this (don't pass any `boundsZoomLevel`). Seems to work for me.

DonutMarkerMapType.tsx and BarMarkerMapType.tsx
```
  const previewMarkerResult = useMarkerData({
    studyId,
    filters,
    studyEntities,
    geoConfigs,
//    boundsZoomLevel: appState.boundsZoomLevel,
    selectedVariable: configuration.selectedVariable,
    binningMethod: configuration.binningMethod,
    selectedValues: configuration.selectedValues,
    valueSpec: 'count',
  });
```

image

bobular:
Fixes #708
Fixes #713 

See #708 for use cases that were broken.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BarMarkerMapType.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BarMarkerMapType.tsx
@@ -37,6 +37,8 @@ import {
   DistributionMarkerDataProps,
   defaultAnimation,
   isApproxSameViewport,
+  isNoDataError,
+  noDataErrorMessage,
   useCategoricalValues,
   useCommonData,
   useDistributionMarkerData,
@@ -160,7 +162,6 @@ function ConfigPanelComponent(props: MapTypeConfigPanelProps) {
     filters,
     studyEntities,
     geoConfigs,
-    boundsZoomLevel: appState.boundsZoomLevel,
     selectedVariable,
     selectedValues,
     binningMethod,
@@ -356,7 +357,9 @@ function MapLayerComponent(props: MapTypeMapLayerProps) {
   });
 
   if (markerData.error && !markerData.isFetching)
-    return <MapFloatingErrorDiv error={markerData.error} />;
+    return isNoDataError(markerData.error) ? null : (
+      <MapFloatingErrorDiv error={markerData.error} />
+    );
 
   // pass selectedMarkers and its state function
   const markers = markerData.markerProps?.map((markerProps) => (
@@ -384,14 +387,8 @@ function MapLayerComponent(props: MapTypeMapLayerProps) {
 }
 
 function MapOverlayComponent(props: MapTypeMapLayerProps) {
-  const {
-    studyEntities,
-    studyId,
-    filters,
-    geoConfigs,
-    appState: { boundsZoomLevel },
-    updateConfiguration,
-  } = props;
+  const { studyEntities, studyId, filters, geoConfigs, updateConfiguration } =
+    props;
   const configuration = props.configuration as BarPlotMarkerConfiguration;
   const findEntityAndVariable = useFindEntityAndVariable();
   const { variable: overlayVariable } =
@@ -412,7 +409,6 @@ function MapOverlayComponent(props: MapTypeMapLayerProps) {
     studyId,
     filters,
     geoConfigs,
-    boundsZoomLevel,
     selectedVariable: configuration.selectedVariable,
     binningMethod: configuration.binningMethod,
     dependentAxisLogScale: configuration.dependentAxisLogScale,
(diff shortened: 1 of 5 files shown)


packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BarMarkerMapType.tsx
```diff
@@ -356,7 +357,9 @@ function MapLayerComponent(props: MapTypeMapLayerProps) {
   });
 
   if (markerData.error && !markerData.isFetching)
-    return <MapFloatingErrorDiv error={markerData.error} />;
+    return isNoDataError(markerData.error) ? null : (
+      <MapFloatingErrorDiv error={markerData.error} />
+    );
```
bobular, line 362: Above is where we ignore certain types of error and show no markers or spinners on the map (returning `null`)

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
```diff
@@ -304,6 +299,15 @@ export function useDistributionMarkerData(props: DistributionMarkerDataProps) {
     },
     enabled: overlayConfig != null,
```
dmfalke, line 300: I wonder if this should be something like
```suggestion
    enabled: !overlayConfigResult.isFetching,
```

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
```diff
@@ -365,3 +369,23 @@ export const GLOBAL_VIEWPORT = {
...
   },
 };
+
+/**
+ * Error handling helpers
+ */
+
+const noDataPatterns = [
+  /did not contain any data/,
+  /Could not generate continuous variable metadata/,
```
dmfalke, line 379: When does this happen?

dmfalke (commented):
This looks good. I have a couple of questions below.

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
bobular: I'm pretty sure both are caused by completely empty subsets - this is the error the back end throws to prevent the R stuff blowing up.

The first one comes from the `map-markers` endpoint and the second one from `continuous-variable-metadata`

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
bobular: Maybe `overlayConfig != null && !overlayConfigResult.isFetching`?

This would make sure the main marker data request doesn't get made with stale overlayConfig data perhaps?

What was your reasoning or hunch?

dmfalke: Yes, that was my thinking.

dmfalke (approved):

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
bobular: Added some comments to help future coders

bobular:
Thanks for your review. I've added your `enabled:` suggestion and tested it and didn't spot any problems.

Issue #450, opened 2022-02-18
EDA ClinEpi: add scale to display labels


danicahelb:
We are using “scale” annotation in the owl file to indicate whichever type of log (log2, log10, ln, etc.) the data has been loaded in.
"Scale" information should be added to the display label, in a similar manner to how units are being added. 

If the label has units, the type of log should be added in the parentheses & separated from the units by a comma. For example: Alpha-1-antitrypsin (ln, mg/g)

see also https://redmine.apidb.org/issues/43947

ryanrdoherty:
@dmfalke Backend changes are now in dev and can be put on QA by upgrading eda-subsetting in tagger to 3.11.0.

PR #721, opened 2023-12-11
add scale to EDA's display labels

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/450.

It was hard to find where the scale info (metadata) is used but at least I could manage to find one study, Mal-ED study. After checking the codes carefully, I realized that an easy solution was to change `variableDisplayWithUnit()` so that all (or most) Vizs show scale info at various display labels such as axis title, legend title, variable coverage table etc. Below are screenshot with Scatterplot for before and after this PR.

I saw that there were cases where units are empty while scale exists. In that case, I just used scale with parenthesis: e.g., Polio serotype... in the "After" screenshot. 

### **- Before**
MAL-ED scale01

### **- After**
MAL-ED scale01-PR


packages/libs/eda/src/lib/core/utils/variable-display.ts
```diff
@@ -1,9 +1,29 @@
...
+        return variable.displayName + ' (' + variable.units + ')';
+      }
+    } else {
+      if (variable.scale && variable.scale != null) {
+        return variable.displayName + ' (' + variable.scale + ')';
+      } else {
+        return variable?.displayName;
+      }
+    }
+  } else return variable?.displayName;
```
dmfalke, line 28: There is quite a bit of nesting of if-statements here. I think we can reduce the this significantly

```suggestion
  if (variable == null) return;
  if (!NumberVariable.is(variable)) return variable.displayName;
  
  const parentheticalParts = [ variable.units, variable.scale ].filter(Boolean);
  
  if (parentheticalParts.length) return `${variable.displayName} (${parentheticalParts.join(', ')})`;

  return variable.displayName;
```

dmfalke (changes requested):
Overall, this looks good. I added a suggested change, to simplify the code. Let me know what you think. Thanks!

moontrip:
> Overall, this looks good. I added a suggested change, to simplify the code. Let me know what you think. Thanks!

Hi again @dmfalke 👋 Thank you for your review and suggestion! I saw the technique, array.filter(Boolean), somewhere before and I think it would work quite well in this case. Will check it out next week 👍

moontrip:
@dmfalke I have addressed your feedbacks. It would be great if you can check it out 👍

dmfalke (approved):

moontrip:
Thank you for your review @dmfalke 👍

PR #741, opened 2023-12-19
Minor polishing updates for correlation

### LINKED ISSUE #727: Allow the user to pick correlation method

### LINKED ISSUE #728: Prevent the threshold inputs from allowing ridiculous options and unify language

asizemore:
Resolves #727 
Resolves #728 

This PR addresses a few minor polishing updates for the correlation app

- [ ] Allows user to select method (spearman vs pearson)
- [ ] Prevents the thresholds from being ridiculous and unifies language
- [ ]

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/Plugins.scss
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/Plugins.scss
@@ -18,6 +18,16 @@
     }
   }
 
+  &-CorrelationOuterConfigContainer {
+    display: flex;
+    flex-direction: column;
+    gap: 0.75em;
+    span {
+      white-space: nowrap;
+    }
+    text-transform: capitalize;
+  }
+
   &-InputContainer {
     display: flex;
     gap: 0.75em;
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
@@ -18,6 +18,7 @@ import { bipartiteNetworkVisualization } from '../../visualizations/implementati
 import { variableCollectionsAreUnique } from '../../../utils/visualization';
 import PluginError from '../../visualizations/PluginError';
 import { VariableCollectionSelectList } from '../../variableSelectors/VariableCollectionSingleSelect';
+import SingleSelect from '@veupathdb/coreui/lib/components/inputs/SingleSelect';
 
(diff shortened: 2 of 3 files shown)


dmfalke:
@asizemore I think the layout looks great! I did a quick trial run (using qa.microbiomedb.org), and the "link width" legend has a bunch of `NaN`s

image

dmfalke (commented):
Code changes look great, btw!

asizemore:
>  the "link width" legend has a bunch of NaNs

For some calculations we're getting nans for the correlation value. Originally i thought 'oh that's a backend issue'. It is, and there's a ticket [here](https://github.com/VEuPathDB/microbiomeComputations/issues/73#issuecomment-1876921195) for it. But also the frontend should not show these NaNs in case some get through. In fact, a few weeks ago @bobular even suggested the legend be more robust in handling stroke widths that might not be numbers, and i naiively thought 'oh no, that'll never happen' 😆 .

So anyways, I'll make a fix for handling NaNs and update you when it's done!

d-callan:
so i said in the backend ticket, but will here as well.. i think the NaNs are a legitimate result. sometimes we have data, but not enough to produce a reliable correlation coefficient. id be inclined to excluding these edges from the diagram, building the legend as though they dont exist as well, and possibly reporting somewhere how many pairs we did this for if we think thatd be helpful.

also worth noting there is a ticket to consider adding options to pre-filter variables, which i think is a related but separate issue. its possible, if we do choose to pre-filter variables, that this issue goes away.. but its also possible it wont if we decide to offer the user control over the pre-filtering.

asizemore:
@bobular and @dmfalke based on @d-callan 's feedback i think there are a few options for handling NaNs in the frontend that i want to think through a bit. If you don't mind, i'd like to put that work in a separate PR. Mind if we just go ahead and review the layout, node labels, and threshold inputs for this PR? If it's helpful, the BONUS and HMP WGS studies usually produce non-NaN results

bobular (approved):
The pedant in me would like these two legend headings to be consistent:
image

I think the main headings should be "Link width" and "Link color".

Perhaps they should be "Link width (absolute correlation coefficient)" and "Link color (correlation direction)"? (And possibly remove the word "correlation" from the link color labels.)

Overall - looks great and happy to merge with/without my suggestions.

I found a scalability issue that I will Slack you about.

dmfalke (approved):

asizemore:
@bobular i went ahead and fixed the capitalization for all apps using that lovely `capitalize` lodash fn. I also unified the legend text and increased the default correlation coefficient as we discussed in slack.

Thank you both!

Issue #567, opened 2023-09-14
study access service: make data access dashboard button more prominent


danicahelb:
we should “unhide” the button to open the data access dashboard… we've now handed over access management to the teams and the current location is too hidden

Make a new tab labeled “Manage study” that is only visible to provider-managers (and VEuPathDB staff) when logged in. Styling of this tab should be different so that this tab (only visible to managers) is differentiated from the other tabs (always visible to all users). Perhaps make the tab the same coloring as the button?

Once the new tab is implemented, delete the button from the study details tab

<img width="1428" alt="image" src="https://github.com/VEuPathDB/EdaNewIssues/assets/25450900/922a575e-6779-4053-b651-09296f34a677">

danicahelb:
All of the information on the study access page (ie, https://clinepidb.org/ce/app/study-access/DS_8e1f623542) should be on the new "Manage study" tab

<img width="666" alt="image" src="https://github.com/VEuPathDB/service-dataset-access/assets/25450900/01aa5749-2298-4a17-bfbc-5fcfb4487b83">

dmfalke:
@danicahelb what if we move the button next to the study name in the analysis header, and skip making the new tab?

![image](https://github.com/VEuPathDB/web-monorepo/assets/365139/7cbf109e-b51f-4094-8a8f-7c72e8a539bb)

danicahelb:
yes, that is fine

danicahelb:
can we push this to the live site as soon as it is QA'd (ie, not wait to b67 release)?

dmfalke:
> can we push this to the live site as soon as it is QA'd (ie, not wait to b67 release)?

That should be doable. @chowington I can handle that, since the process is pretty verbose.

PR #742, opened 2023-12-19
Move data access dashboard button next to study title

chowington:
Closes https://github.com/VEuPathDB/web-monorepo/issues/567


packages/libs/eda/src/lib/workspace/EDAWorkspaceHeading.tsx
```diff
@@ -62,50 +64,68 @@ export function EDAWorkspaceHeading({
...
+              }}
+            >
+              <div>
+                <FloatingButton
+                  themeRole="primary"
+                  text="New analysis"
+                  tooltip="Create a new analysis"
+                  textTransform="capitalize"
+                  size="medium"
+                  // @ts-ignore
```
dmfalke, line 99: Can this be removed?

dmfalke (commented):
This looks pretty good, but I think the button text is too big, and it should match the site theme. Maybe use a coreui button for this?

image

packages/libs/eda/src/lib/workspace/EDAWorkspaceHeading.tsx
chowington: Ah, yep!

chowington:
@dmfalke I was thinking we might want to retain the styling of the button (at least for now) since it's moving to a new place and won't be where users are used to it. What do you think? Maybe just shrink the text a bit?

dmfalke:
I definitely think we should shrink the font size. I suspect folks will also want it to match the rest of the button styles, but I also see your point.

@danicahelb @aurreco-uga do you have an opinion?

dmfalke:
@chowington after some intense reflection, I think you're right.

asizemore:
Just to keep it exciting, i'll disagree. My proposal is to use the Filled version of the coreui button and the theme color. The biggest reason is because what do we do the next time we need to make something more prominent? We'll have to make it more prominent than a button that's already broken the site style, so we'll have to break the site style bigger next time. This is a downhill road.

Now remember i am so far not at all involved in this ticket and we had already accepted the above proposal so please feel free to ignore this 😄

chowington:
@asizemore The idea was to only keep the prominent style for a little while so users could find it at its new location; however I think using CoreUI's FilledButton works just as well. I've updated the style, as well as added an info box at the button's previous location saying it's been moved. What do y'all think?

image

dmfalke:
I think this looks great. What do you think about adding a little more space (doubling, maybe) between the study name and the button?

asizemore:
Looks great! I agree with @dmfalke that a little more spacing would help. Probably just enough so that the button is about one "space" (like the space between words) from the end of the study name?

chowington:
How about this?

image

danicahelb:
Looks great @chowington! 

Please change the verbiage to `Study management dashboard` or more simply `Management dashboard`

dmfalke (approved):
Changes look great. Pending changes related to Danica's request, feel free to merge.

Issue #734, opened 2023-12-15
Map - click (not on marker), pan or zoom on map should hide the left hand menu/panel.


bobular:
Self explanatory I hope!

PR #743, opened 2023-12-19
hide left panel and map viz at SAM when map events occur

moontrip:
This addresses https://github.com/VEuPathDB/web-monorepo/issues/734. I think that in conjunction with left panel, it is better to close opened map Viz as well, thus this PR included that part as well.

Here is a video shot for a demo:

https://github.com/VEuPathDB/web-monorepo/assets/12802305/c9ce92c7-e119-4318-abfc-1a2c6d51bc60

### DIFF
--- a/packages/libs/components/src/map/MapVEuMap.tsx
+++ b/packages/libs/components/src/map/MapVEuMap.tsx
@@ -335,9 +335,6 @@ function MapVEuMapEvents(props: MapVEuMapEventsProps) {
         zoomLevel: mapEvents.getZoom(),
       };
       onBoundsChanged(boundsViewport);
-
-      // map zoom occurred
-      if (setIsMapEvents != null) setIsMapEvents(true);
     },
     moveend: () => {
       onViewportChanged({
@@ -352,9 +349,6 @@ function MapVEuMapEvents(props: MapVEuMapEventsProps) {
         zoomLevel: mapEvents.getZoom(),
       };
       onBoundsChanged(boundsViewport);
-
-      // map panning occurred
-      if (setIsMapEvents != null) setIsMapEvents(true);
     },
     baselayerchange: (e: { name: string }) => {
       onBaseLayerChanged && onBaseLayerChanged(e.name as BaseLayerChoice);
@@ -363,7 +357,7 @@ function MapVEuMapEvents(props: MapVEuMapEventsProps) {
     click: () => {
       if (setSelectedMarkers != null) setSelectedMarkers(EMPTY_MARKERS);
 
-      // map click occurred???
+      // map click occurred
       if (setIsMapEvents != null) setIsMapEvents(true);
     },
   });
--- a/packages/libs/components/src/map/MapVEuMap.tsx
+++ b/packages/libs/components/src/map/MapVEuMap.tsx
@@ -154,7 +154,7 @@ export interface MapVEuMapProps {
   /* selectedMarkers setState (for on-click reset) **/
   setSelectedMarkers?: React.Dispatch<React.SetStateAction<string[]>>;
   children?: React.ReactNode;
-  /*  setIsMapEvents is used to check if map events such as click, zoom, and panning occured */
+  /* setIsMapEvents is used to check if map events such as click, zoom, and panning occured */
   setIsMapEvents?: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
--- a/packages/libs/components/src/map/MapVEuMap.tsx
+++ b/packages/libs/components/src/map/MapVEuMap.tsx
@@ -154,8 +154,10 @@ export interface MapVEuMapProps {
   /* selectedMarkers setState (for on-click reset) **/
   setSelectedMarkers?: React.Dispatch<React.SetStateAction<string[]>>;
   children?: React.ReactNode;
-  /* setIsMapEvents is used to check if map events such as click, zoom, and panning occured */
-  setIsMapEvents?: React.Dispatch<React.SetStateAction<boolean>>;
+  /* hide map side panel */
+  onMapClick?: (value: boolean) => void;
+  onMapDrag?: (value: boolean) => void;
+  onMapZoom?: (value: boolean) => void;
 }
 
 function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
@@ -179,7 +181,9 @@ function MapVEuMap(props: MapVEuMapProps, ref: Ref<PlotRef>) {
     interactive = true,
     defaultViewport,
     setSelectedMarkers,
-    setIsMapEvents,
+    onMapClick,
+    onMapDrag,
+    onMapZoom,
   } = props;
 
   // use a ref to avoid unneeded renders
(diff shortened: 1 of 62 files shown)


moontrip:
Hi @bobular your feedbacks at Slack are addressed so it would be great if you can check out.

packages/libs/components/src/map/MapVEuMap.tsx
dmfalke, line 1: Instead of the `setIsMapEvents` prop, I would rather see callback props for each event. The current approach requires state to track when an event is triggered, and it also doesn't indicate which event was triggered.

The props could be as simple as

```typescript
interface MapVEuMapProps {
  // ...
  onClick: () => void;
  onDrag: () => void;
  onZoom: () => void;
}
```

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
dmfalke, line 1: Following on to what I said above, the callback used for each event handler should just call `appState.setIsSidePanelExpanded(false)`.

dmfalke (changes requested):
@moontrip I made some comments recommending a different approach.

The upshot is that `MapVEuMap` should take callback props for each event, and `MapAnalysis` should pass a callback that updates the appState so that `isSidePanelExpanded = false`.

moontrip:
> @moontrip I made some comments recommending a different approach.
> 
> The upshot is that `MapVEuMap` should take callback props for each event, and `MapAnalysis` should pass a callback that updates the appState so that `isSidePanelExpanded = false`.

Hi @dmfalke 👋 Thank you for your suggestions. I will test them next week 👍 Have a great weekend!

dmfalke:
> Hi @dmfalke 👋 Thank you for your suggestions. I will test them next week 👍 Have a great weekend!

🤘

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
bobular: I agree with Dave!

moontrip:
Hi @dmfalke and @bobular 

I have addressed your feedbacks with some changes. One question I have is that actually we do not necessarily pass three props such as onDrag, onZoom, and onClick because a map event (drag, zoom, or click) is detected at corresponding MapVEuMapEvents in the MapVEuMap component. So, perhaps we can simply pass single prop such as setIsSidePanelExpanded(), to MapVEuMapEvents /MapVEuMap?

dmfalke:
> Hi @dmfalke and @bobular
> 
> I have addressed your feedbacks with some changes. One question I have is that actually we do not necessarily pass three props such as onDrag, onZoom, and onClick because a map event (drag, zoom, or click) is detected at corresponding MapVEuMapEvents in the MapVEuMap component. So, perhaps we can simply pass single prop such as setIsSidePanelExpanded(), to MapVEuMapEvents /MapVEuMap?

I think we should keep the three props. That gives us flexibility in the future, in case we want to change which events cause the side panel to be collapsed.

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -303,6 +313,9 @@ interface MapVEuMapEventsProps {
   onBoundsChanged: (bondsViewport: BoundsViewport) => void;
   onBaseLayerChanged?: (newBaseLayer: BaseLayerChoice) => void;
   setSelectedMarkers?: React.Dispatch<React.SetStateAction<string[]>>;
+  onMapClick?: (value: boolean) => void;
+  onMapDrag?: (value: boolean) => void;
+  onMapZoom?: (value: boolean) => void;
```
dmfalke, line 318: Let's keep these generic and call them with any arguments. In other uses cases, it wouldn't be clear what the boolean `value` means.

```suggestion
  onMapClick?: () => void;
  onMapDrag?: () => void;
  onMapZoom?: () => void;
```

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -900,6 +898,9 @@ function MapAnalysisImpl(props: ImplProps) {
                     defaultViewport={defaultViewport}
                     // for multiple markers cancelation of selection only
                     setSelectedMarkers={setSelectedMarkers}
+                    onMapClick={setIsSidePanelExpanded}
+                    onMapDrag={setIsSidePanelExpanded}
+                    onMapZoom={setIsSidePanelExpanded}
```
dmfalke, line 903: Instead of passing the in the `useState` setter function, use a wrapper function that calls it:

```ts
const closePanel = useCallback(() => setIsSidePanelExpanded(false), []);
```

```suggestion
                    onMapClick={closePanel}
                    onMapDrag={closePanel}
                    onMapZoom={closePanel}
```

dmfalke (changes requested):

moontrip:
> > Hi @dmfalke and @bobular
> > I have addressed your feedbacks with some changes. One question I have is that actually we do not necessarily pass three props such as onDrag, onZoom, and onClick because a map event (drag, zoom, or click) is detected at corresponding MapVEuMapEvents in the MapVEuMap component. So, perhaps we can simply pass single prop such as setIsSidePanelExpanded(), to MapVEuMapEvents /MapVEuMap?
> 
> I think we should keep the three props. That gives us flexibility in the future, in case we want to change which events cause the side panel to be collapsed.

Thank @dmfalke for your comment and suggestions. I applied them to the new commit 👍

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -154,6 +154,10 @@ export interface MapVEuMapProps {
   /* selectedMarkers setState (for on-click reset) **/
   setSelectedMarkers?: React.Dispatch<React.SetStateAction<string[]>>;
   children?: React.ReactNode;
+  /* close left-side panel when map events happen */
```
bobular, line 157: We shouldn't be documenting behaviour of the map code here. The callbacks can do *anything*!

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -871,9 +872,9 @@ function MapAnalysisImpl(props: ImplProps) {
                   <MapSidePanel
                     isExpanded={appState.isSidePanelExpanded}
                     isUserLoggedIn={userLoggedIn}
-                    onToggleIsExpanded={() =>
-                      setIsSidePanelExpanded(!appState.isSidePanelExpanded)
-                    }
+                    onToggleIsExpanded={() => {
+                      setIsSidePanelExpanded(!appState.isSidePanelExpanded);
+                    }}
```
bobular, line 877:
```suggestion
                    onToggleIsExpanded={() =>
                      setIsSidePanelExpanded(!appState.isSidePanelExpanded)
                    }
```

I'd go back to the original. Maybe you added logging here but best to not make unnecessary changes.

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -343,13 +362,19 @@ function MapVEuMapEvents(props: MapVEuMapEventsProps) {
         zoomLevel: mapEvents.getZoom(),
       };
       onBoundsChanged(boundsViewport);
+
+      // close left-side panel when drag/panning event happens
```
bobular, line 366: Same comment about this comment (and the ones above and below). Best to remove.

packages/libs/components/src/map/MapVEuMap.tsx
moontrip: Thank you for your review and suggestions @bobular I got your point and remove it.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
moontrip: complied!

packages/libs/components/src/map/MapVEuMap.tsx
moontrip: all removed!

bobular (approved):
Looks great, thanks!

dmfalke (approved):
Looks great!

moontrip:
Thanks @bobular and @dmfalke for your reviews and feedbacks 👍

PR #746, opened 2024-01-02
Use download reporters for standalone srt tool

dmfalke:
### Description

Addresses https://redmine.apidb.org/issues/52747

Update SRT page to use question/reporter combos. This will aid in the removal of the cgi based SRT.

### Screenshots

image

image

image

image


aurreco-uga (approved):

dmfalke:
Out of curiosity, should there be a limit on the number of input IDs? Since we now allow strategies and baskets, the number of IDs can be huge.

PR #751, opened 2024-01-05
allow plot references to be null or omitted

d-callan:
wip

i made this change so i could test some backend changes locally, but would like to know if people have reasons why i shouldnt merge it. mostly im collecting thoughts/ opinions right now, if anyone has any.

i made a change in https://github.com/VEuPathDB/EdaDataService/pull/342 that means sometimes well see empty plot refs. theyre just variables that i am keeping track of for one reason or another but arent actually in the plot. in this case it was so they could be used to properly impute zeroes in the popbio megastudy.

there is also a pr to discuss what these should actually look like as json https://github.com/VEuPathDB/veupathUtils/pull/33

### DIFF
--- a/packages/libs/eda/src/lib/core/api/DataClient/types.ts
+++ b/packages/libs/eda/src/lib/core/api/DataClient/types.ts
@@ -87,17 +87,20 @@ const variableSpec = type({
 });
 
 export type PlotReferenceValue = TypeOf<typeof plotReferenceValue>;
-const plotReferenceValue = keyof({
-  xAxis: null,
-  yAxis: null,
-  zAxis: null,
-  overlay: null,
-  facet1: null,
-  facet2: null,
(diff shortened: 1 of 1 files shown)


dmfalke:
I don't see any issue with merging this. I'm not sure the `nullType` is needed, unless the back end is providing a `null` value in the JSON.

d-callan:
> I don't see any issue with merging this. I'm not sure the nullType is needed, unless the back end is providing a null value in the JSON

it is currently, and https://github.com/VEuPathDB/veupathUtils/pull/33 should fix that. but this would let us be reverse compatible/ let me keep testing before that gets merged

bobular:
I can't see any unintended consequences within the front end code for this, so go for it!

dmfalke (approved):

PR #753, opened 2024-01-08
allow studies with one entity in correlation metadata app

asizemore:
The correlation metadata x assay app previously only allowed studies with 1) at least one entity and 2) entities with ids we have information about in the app. That's great for curated studies, but user studies have only one entity and their entity name isn't something we can predict. That causes an issue for this app because the user datasets will never pass the requirements for the app so won't be able to use it, even though they should.

This PR updates the logic in the correlation assay x metadata app so that user datasets can use it. In short, if a study has only one entity (usually a user dataset, but could be anything), then we'll look for continuous metadata in all the variables _that are not already in a collection_. We're making the assumption that collections contain only assay-type data, no metadata. In some distant future where a user can make their own collections, this assumption will no longer be valid, but it seems like that future is probably years away so i'm letting our smarter, cleverer future selves figure that one out :) 

I'm going to keep this in draft form until the matching backend is ready, but the frontend is totally testable as-is.


dmfalke:
This sounds like a good approach. Has any thought been put into annotation collections so we can determine if they are assay vs metadata collections? Are there other collection types on the horizon?

asizemore:
thats a good question. I've been talking to john b about it. One idea so far is to use the assumption that all collections are for assays. This is true, for now. I don't know of any plans for collections that aren't assays for the near future, so maybe that approach would at least improve the code for a bit... but we'll see what jb says about adding an isAssay annotation (or similar) to assay entities.

asizemore:
Update - all the backend changes are in! This is now testable from QA

asizemore:
@dmfalke lmk if you need a user dataset to test

dmfalke:
> @dmfalke lmk if you need a user dataset to test

If you have one to share, that would be great.

dmfalke (approved):
This looks great. I tested using the user dataset that you shared with me, and the correlation app was available.

asizemore:
Awesome thanks Dave!

Issue #658, opened 2023-11-17
Map - categorical marker configuration has inconsistent handling of "all other values"


bobular:
You should be able to reproduce this with any low-cardinality variable.
(I think high-cardinality are better behaved but I will need to check...)

1. donut marker mode
2. choose a low-cardinality variable anew
3. unselect one checkbox - see that the legend does not contain "All other values"
4. keep unchecking - notice the numbers in the markers get smaller (this is desired)
5. when no values are checked the numbers are zero (also desirable, I guess)
6. re-checking markers will reverse the process.
7. so far so good
8. now click the "uncheck all" header checkbox
9. "All other values" now shows in the legend. This isn't bad but it's inconsistent!
10. check some values - notice that the "All other values" remains and the numbers in the markers stay constant (also a good thing)
11. when all values are checked, "All other values" disappears (desirable) and we are effectively back to step 2. 

It is confusing that sometimes you get "All other values" and other times you don't.

For low cardinality variables, e.g. "male", "female" it would be annoying if "All other values" always kicked in. When deselecting "female" you would simply get "All other values" replacing the female donut segments. So the solution is probably to make sure AOV doesn't show for low-cardinality vars, but see the tickets below.

This is strongly related to future feature enhancement tickets
* https://github.com/VEuPathDB/web-monorepo/issues/511
* https://github.com/VEuPathDB/web-monorepo/issues/510
* and more... https://github.com/VEuPathDB/web-monorepo/issues?q=is%3Aissue+is%3Aopen+categorical+marker+configuration

- [x] check high cardinality vars
       - these are consistent - "All other values" appears all the time.

PR #754, opened 2024-01-08
Don't show 'All other values' for low-cardinality variables

chowington:
Fixes https://github.com/VEuPathDB/web-monorepo/issues/658


bobular (approved):
Looks good, and tested fine. There's a bit of repetition that could perhaps be factored out?

Issue #760, opened 2023-11-29
SAM: limit decimal points in mouseover markers


danicahelb:
There are way too many decimal points in mouseover of markers. Perhaps limit to 2 decimal points? Or 3 sig figs?
<img width="285" alt="image" src="https://github.com/VEuPathDB/EdaNewIssues/assets/25450900/d0e7733a-352a-4e16-a381-708a8c98ab68">

PR #772, opened 2024-01-16
limits decimal points for SAM barplot marker

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/760.

Screenshots are attached for before and after this PR. SCORE S. mansoni Cluster Randomized Trial is used.

# before
SAM bar decimal points1

# after
SAM bar decimal points2

### DIFF
--- a/packages/libs/components/src/plots/Barplot.tsx
+++ b/packages/libs/components/src/plots/Barplot.tsx
@@ -121,7 +121,12 @@ const Barplot = makePlotlyPlotComponent(
               orientation: orientation === 'vertical' ? 'v' : 'h',
               opacity: calculatedOpacity,
               type: 'bar',
-              text: showValues ? el.value : undefined,
+              // two decimal points for y values
+              text: showValues
+                ? el.value.map(
+                    (value: number) => `${Number.parseFloat(value.toFixed(2))}`
+                  )
+                : undefined,
               textposition: showValues ? 'auto' : undefined,
               marker: {
                 color: el.color,


packages/libs/components/src/plots/Barplot.tsx
```diff
@@ -121,7 +121,12 @@ const Barplot = makePlotlyPlotComponent(
               orientation: orientation === 'vertical' ? 'v' : 'h',
               opacity: calculatedOpacity,
               type: 'bar',
-              text: showValues ? el.value : undefined,
+              // two decimal points for y values
+              text: showValues
+                ? el.value.map(
+                    (value: number) => `${Number.parseFloat(value.toFixed(2))}`
```
dmfalke, line 127: I think some unnecessary type conversions can be removed.

```suggestion
                    (value: number) => value.toFixed(2)
```

dmfalke (commented):
@moontrip let me know what you think of my suggestion below.

packages/libs/components/src/plots/Barplot.tsx
moontrip: Hi @dmfalke The reason why I introduced Number.parseFloat() is that I wanted to remove unnecessary zeros added to an integer value. For example, for the value of 10, 10.toFixed(2) becomes 10.00. I think it is better to show 10 only in that case?

dmfalke: Ah, that makes sense. Thanks for the explanation.

dmfalke (approved):

packages/libs/components/src/plots/Barplot.tsx
moontrip: My pleasure 😃 Thank you for your review again 👍

PR #774, opened 2024-01-17
Feature 747 bpnet nodata ux

### LINKED ISSUE #747: Improve UX when no data passes the thresholds
Right now if we change the thresholds to be so high that all results are filtered out, we get a viz that looks like

Screen Shot 2024-01-03 at 1 11 34 PM
(from Bonus, correlation assay v metadata)

Which is not ideal. Let's instead have it look like...

Screen Shot 2024-01-03 at 1 10 33 PM

So we see text in the empty plot that tells the user how to fix the issue, also we can highlight the input boxes in warning yellow

asizemore:
Resolves #747 

In progress. Adds a nice way to render a custom node when the network is totally empty. However, i'm stuck on getting those darn input boxes to change outline colors. I know they _can_ do it because they turn red with an error, but i gotta figure out how to make them yellow with a warning.

### DIFF
--- a/packages/libs/components/src/plots/BipartiteNetwork.tsx
+++ b/packages/libs/components/src/plots/BipartiteNetwork.tsx
@@ -5,6 +5,7 @@ import { Graph } from '@visx/network';
 import { Text } from '@visx/text';
 import {
   CSSProperties,
+  ReactNode,
   Ref,
   forwardRef,
   useImperativeHandle,
@@ -42,6 +43,8 @@ export interface BipartiteNetworkProps {
   showSpinner?: boolean;
   /** Length of node label text before truncating with an ellipsis */
   labelTruncationLength?: number;
+  /** Additional error messaging to show when the network is empty */
+  emptyNetworkContent?: ReactNode;
 }
 
 // Show a few gray nodes when there is no real data.
@@ -71,6 +74,7 @@ function BipartiteNetwork(
     containerClass = 'web-components-plot',
     showSpinner = false,
     labelTruncationLength = 20,
+    emptyNetworkContent,
   } = props;
 
   // Use ref forwarding to enable screenshotting of the plot for thumbnail versions.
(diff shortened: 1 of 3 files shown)


bobular (commented):
The empty plot and yellow/orange warnings look great.

Before I get to that, I have a general problem with the interface.
If I make a new viz, set the input data and correlation method, then it still says it's not configured:
image

If I change the variance to 1 and then back to the default zero I am allowed to click on the blue button.

image

I don't know what these parameters mean and how to set them. I assume the defaults are OK to use. I'll see if there's a simple fix.

bobular:
I don't know if it's a simple fix but

I'm not sure why we're using `CompleteCorrelationAssayAssayConfig`

```
export const CorrelationAssayAssayConfig = t.partial({
  collectionVariable1: VariableCollectionDescriptor,
  collectionVariable2: VariableCollectionDescriptor,
  correlationMethod: t.string,
  prefilterThresholds: FeaturePrefilterThresholds,
});

const CompleteCorrelationAssayAssayConfig = partialToCompleteCodec(
  CorrelationAssayAssayConfig
);
```
would something like this would be more appropriate?
```
export const CorrelationAssayAssayConfig = t.intersection([
 // required params
 t.type({
  collectionVariable1: VariableCollectionDescriptor,
  collectionVariable2: VariableCollectionDescriptor,
  correlationMethod: t.string,
 }),
 // optional params
 t.partial({
  prefilterThresholds: FeaturePrefilterThresholds,
 })
]);
```

That may have downstream consequences (it does, I just tried it).

bobular (approved):
That's not really related to this issue, so I've looked over this code and it looks great!

dmfalke:
> I don't know if it's a simple fix but
> 
> I'm not sure why we're using `CompleteCorrelationAssayAssayConfig`
> 
> ```
> // eslint-disable-next-line @typescript-eslint/no-redeclare
> export const CorrelationAssayAssayConfig = t.partial({
>   collectionVariable1: VariableCollectionDescriptor,
>   collectionVariable2: VariableCollectionDescriptor,
>   correlationMethod: t.string,
>   prefilterThresholds: FeaturePrefilterThresholds,
> });
> 
> const CompleteCorrelationAssayAssayConfig = partialToCompleteCodec(
>   CorrelationAssayAssayConfig
> );
> ```
> 
> would something like this would be more appropriate?
> 
> ```
> export const CorrelationAssayAssayConfig = t.intersection([
>  // required params
>  t.type({
>   collectionVariable1: VariableCollectionDescriptor,
>   collectionVariable2: VariableCollectionDescriptor,
>   correlationMethod: t.string,
>  }),
>  // optional params
>  t.partial({
>   prefilterThresholds: FeaturePrefilterThresholds,
>  })
> ]);
> ```
> 
> That may have downstream consequences (it does, I just tried it).

The idea behind "complete" configs is that they can be used to run a compute. It may be that the "complete" config here needs to be relaxed so that prefilters are optional.

asizemore:
Thank you both! 
You're right the config with defaults there is a little messed up. We had planned to have those Prefilter inputs up for only a few days for testing before taking them down for the release, so we originally didn't put in much effort to the ux. Now that they'll live another two weeks though i think it's worth making sure they're not a pain to use! I'll address that in another PR.

PR #779, opened 2024-01-17
Remove compute id from route

### LINKED ISSUE #778: Visualizations: Remove compute id from route
We currently include the computation id as a part of a visualization's route. Since a visualization can move from one computation to another when modifying the compute configuration, visualization URLs are unstable.

Instead, we should remove the computation id from the route (and add a redirect for URLs that still contain the computation id).

dmfalke:
Closes #778 

This PR removed the computation ID from the url of a visualization. This makes the url stable when changing the configuration of a compute.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -80,7 +80,7 @@ interface Props {
  */
 export function VisualizationsContainer(props: Props) {
   const { baseUrl } = { ...props };
-  const { url } = useRouteMatch();
+  // const { url } = useRouteMatch();
 
   const currentStudyRecordId = useStudyRecord().id[0].value;
   const studiesForPerformanceWarning = [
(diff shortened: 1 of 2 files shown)


packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -320,7 +320,7 @@ export function NewVisualizationPicker(props: NewVisualizationPickerProps) {
     computation,
     geoConfigs,
     onVisualizationCreated = function (visualizationId, computationId) {
```
asizemore, line 322: i think we can remove `computationId` from the arguments

packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
```diff
@@ -80,7 +80,7 @@ interface Props {
  */
 export function VisualizationsContainer(props: Props) {
   const { baseUrl } = { ...props };
-  const { url } = useRouteMatch();
+  // const { url } = useRouteMatch();
```
asizemore, line 83: can we remove the commented code?

packages/libs/eda/src/lib/workspace/ComputationRoute.tsx
```diff
@@ -210,26 +199,38 @@ export function ComputationRoute(props: Props) {
...
-                  // These are routes for the computation instances already saved
-                  const computation =
-                    props.analysisState.analysis?.descriptor.computations.find(
-                      (c) => c.computationId === routeProps.match.params.id
+                  const result =
+                    props.analysisState.getVisualizationAndComputation(
+                      routeProps.match.params.visualizationId
                     );
+                  if (result == null) {
+                    return null;
```
asizemore, line 217: Should we return something like `<div>Cannot find visualization</div>` here?

asizemore (approved):
Awesome! Thanks for doing this!
I just had a couple questions, but nothing looked crazy and the testing went well so i'll go ahead and approve. I tested with a new analysis and one from April of this past year, using my dev site as the backend.

dmfalke:
Thanks for the review! I went ahead and made the changes you suggested. I will merge this soon!

PR #789, opened 2024-01-18
draft some correlation prefilter inputs

d-callan:
so im not worried about this being 'good' or 'pretty'. i dont think this is a thing to release, just a thing to let people play w prefilters. then we can decide what we do and dont like for defaults, whether to offer user controls, if there are other things wed like to prefilter on, etc. then at that point presumably this gets reverted or obviated by some newer change.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
@@ -1,4 +1,7 @@
-import { useFindEntityAndVariableCollection } from '../../..';
+import {
+  FeaturePrefilterThresholds,
+  useFindEntityAndVariableCollection,
+} from '../../..';
 import { VariableCollectionDescriptor } from '../../../types/variable';
 import { ComputationConfigProps, ComputationPlugin } from '../Types';
 import { capitalize, partial } from 'lodash';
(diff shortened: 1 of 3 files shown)


packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
```diff
@@ -203,6 +211,65 @@ export function CorrelationAssayAssayConfiguration(
...
           </div>
+          <div className={cx('-CorrelationOuterConfigContainer')}>
+            <H6>Prefilters</H6>
+            <div className={cx('-InputContainer')}>
+              <span>Proportion non-zero</span>
+              <NumberInput
+                minValue={0}
+                maxValue={1}
+                step={0.01}
+                value={configuration.prefilterThresholds?.proportionNonZero}
```
asizemore, line 222: Consider changing to
```
value={configuration.prefilterThresholds?.proportionNonZero ?? DEFAULT_PROPORTION_NON_ZERO_THRESHOLD}
```

asizemore: and similar for others so that they have defaults.

asizemore: Right now the defaults don't appear until one Prefilter is updated.

asizemore:
Some thoughts for later, since this is just getting us started:

1. We'll have to figure out how to prevent the boxes from jumping when we put in a value that's not allowed
2. The number of parameters right now isn't really ridiculous. Yeah it goes far to the left but it's not crazy. If we leave them there, the Prefilters div should wrap underneath Input Data. I'd also explore making an expandy section for this called Optional Advanced Parameters.
3. The Prefilters should come pre-populated with our defaults

asizemore (approved):
Left a suggestion but i think it's fine as-is for testing purposes. Exciting!!

d-callan:
fwiw i dont think wed want both variance and std dev for real, i prefer std dev, but wanted to collect others thoughts and play w them both. so in reality wed be adding 1 or 2 new inputs

PR #790, opened 2024-01-18
Improve handling of bipartite legend for plots w/ few links

### LINKED ISSUE #768: Bipartite network: adjust legend when only a few edges exist
Found with a user dataset while working on a separate issue, but this problem would affect all datasets. 

The stroke width legend is set up to always show 4 legend items. This doesn't make a lot of sense when there are four or fewer links, as seen below.
Screen Shot 2024-01-12 at 5 33 48 AM

Options
1. If there are four or fewer links, make one legend item per link. Just straight up show the actual value and stroke width for each link
2. If there are 3-4 links, no change. So a user would still see a range of equally spaced stroke widths from max to min, even if the 1 or 2 non max/min stroke widths are not exactly items in the legend. If there are 2-1 links, just make one legend item per link. 

I'm leaning toward option 1 because stroke widths are difficult enough to visually discern, so why not just give folks the actual matched stroke width in the legend. But I think @chowington and @d-callan might have opinions.

jernestmyers:
Resolves #768 

Screenshots from Bangladesh:

Phylum > Spearman
image

Species > Pearson
image

Genus > Spearman
image


jernestmyers:
> Clever solution! Thanks for taking this on @jernestmyers !
> 
> I don't really expect this to happen, but could we protect ourselves against duplicate weights by looking at only how many distinct values are in `dataWeights`? Maybe we make a new `uniqueDataWeights` array and use that instead of `dataWeights` where appropriate here in the legend items calculation? Although it looks like that would only change two lines... anyways what do you think?

Ah yes, that's a good idea! I'll add some logic to guard against duplication.

asizemore (approved):
LGTM!! 🏂

PR #791, opened 2024-01-19
Restrict inputs for functional associations app

### LINKED ISSUE #749: Correlation - restrict inputs for Functional Associations
Dan B asked us to make the functional associations app specifically about taxa vs functional assays. That means we need to
1. Restrict the data 1 input to just taxa
2. Restrict the data2 input to just pathways and EC abundances

asizemore:
Resolves #749 

Restricts the collection inputs for the Functional Associations correlation app. The first collection input should be restricted only to taxa, and the second only to functional data. In our case, functional data means pathway or gene abundance data.

With these tough restrictions, often all collections on an entity might get filtered out, which would leave a lonely entity name in the `VariableCollectionSelectList`. I've updated that component to remove entities with no items (collections).

Testing - use my dev site as a backend. QA backend is being updated today.


dmfalke (approved):
Looks good to me!

PR #792, opened 2024-01-19
Remove `unknown` types

dmfalke:
This PR removes `unknown` types from `Mesa` type definitions.

A big change is that `MesaColumn` now takes two generic types: `Row` and `Key`. The `Row` type is required, and the default for `Key` is `keyof Row`. This covers most use cases, and allows us to better infer the type of `value` in `CellProps`.

Getting these types right is tricky, because `Mesa` tries to allow any data structure to be used for a row, and to allow the consumer to compute the rendered value however it sees fit.

In most cases, we simply use objects as rows, and the keys of those objects as column keys, which this PR attempts to make easy.

### DIFF
--- a/packages/libs/coreui/src/components/Mesa/types.ts
+++ b/packages/libs/coreui/src/components/Mesa/types.ts
@@ -1,8 +1,8 @@
-import React, { CSSProperties } from 'react';
+import React, { CSSProperties, ReactElement, ReactNode } from 'react';
 
-export interface MesaStateProps<Row, Key extends string> {
+export interface MesaStateProps<Row, Key = keyof Row> {
   rows: Row[];
-  columns: MesaColumn<Key>[];
+  columns: MesaColumn<Row, Key>[];
   filteredRows?: Row[];
   uiState?: {
     sort?: MesaSortObject;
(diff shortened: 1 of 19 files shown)


packages/libs/coreui/src/components/Mesa/types.ts
```diff
@@ -35,33 +35,48 @@ export interface MesaStateProps<Row, Key extends string> {
...
 
-interface MesaAction {
+interface MesaAction<Row, Key> {
   selectionRequired?: boolean;
   element: React.ReactNode;
-  callback?: unknown;
-  handler?: unknown;
+  callback?: (row: Row, columns: MesaColumn<Row, Key>[]) => void;
+  handler?: (
+    seection: Row[],
```
jernestmyers, line 64: Nit: typo!

packages/libs/coreui/src/components/Mesa/types.ts
```diff
@@ -35,33 +35,48 @@ export interface MesaStateProps<Row, Key extends string> {
     selectedNoun?: string;
     selectedPluralNoun?: string;
     searchPlaceholder?: string;
-    deriveRowClassName?: unknown; // a function
-    renderEmptyState?: unknown; // a function
-    isRowSelected?: unknown; // a callback function
+    deriveRowClassName?: (row: Row) => string | undefined; // a function
+    renderEmptyState?: () => ReactNode; // a function
+    isRowSelected?: (row: Row) => boolean; // a callback function
```
jernestmyers, line 40: Can we remove these comments here and elsewhere now that the types are more clear?

jernestmyers (approved):
This looks great and compiled nicely when spinning up a local `genomics-site` and `clinepi-site` dev server.

I left just a couple nits.

dmfalke:
Thanks for the review, @jernestmyers. I've addressed your feedback.

PR #793, opened 2024-01-19
Add "Export to SVG" button to plots

### LINKED ISSUE #780: Export plots as SVG
This was a common user request at a recent workshop

dmfalke:
Closes #780 

This PR adds a button to export plots as an SVG file. Plots made with plotly.js are using `Plotly.toImage`. Plots made with visx are using a custom function to export the SVG directly.

When exporting visx plots, the backgound is transparent. All of the elements are there, so it's possible to open the exported svg file in an editor and give it a white background. I think it would be better to make it white by default, but I'm not sure how to do that. Maybe @asizemore has some ideas?

The button is using coreui styling, and is positioned to the bottom-right of the plot:

image

It would be great to get some testing done with this. Specifically, does this meet the requirements for being able to modify the plot for other uses? Should we allow the user to choose other formats (png, jpeg, etc)? It might make sense to get this on qa soon, so that Dan Beiting, et al, can play with it.


asizemore:
I'll check out the code tomorrow!

dmfalke:
> I've just tested so far, but the things ive tested (box, volcano, network) work great! I've been opening them in illustrator and they look really nice. The volcano opens with a black box over it, which is odd but once i deleted it the whole plot was there. This doesn't bother me, personally, because i'm used to weird objects showing up in svgs when i download them from places. In the same vein, it doesn't bother me at all that the background is not white. In fact the first thing i (again, just me personally) do when i would download an svg like this is to immediately delete the background. When these are going into figures or presentations, the background rectangle just gets in the way.

Thanks, this is very useful feedback! :smile: 

> The one question i had was it looks like the bipartite network has the Export button within the plot div. Was that intentional? All the others, even the visx volcano, are just below the bottom of the plot panel.

This was an oversight. I will fix it up.

dmfalke:
> This was an oversight. I will fix it up.

Upon further investigation, the styling for bpnet viz looks a little different than volcano viz. I'll try to fix it up in a bit.

asizemore:
I forgot to say, another thing i like about the button and its placement is that it makes expanding download options easy. For example if we also had a button for download png, and one for download data, etc., it's obvious where those all go.

packages/libs/components/src/plots/PlotlyPlot.tsx
```diff
@@ -95,6 +100,8 @@ function PlotlyPlot<T>(
     spacingOptions,
     showSpinner,
     showNoDataOverlay,
+    showExportButton,
+    exportFileName,
```
asizemore, line 104: Sanity check:
At first i thought it could be helpful to set a default `exportFileName` to cover us in the case of `showExportButton=true` but no `exportFileName` is given. However, since our `PlotlyPlot` component is only eventually called by `makePlotlyPlotComponent` which requires a `displayName` that becomes `exportFileName`, we should never hit the situation of an emty `exportFileName`. 
Yes?

packages/libs/components/src/plots/ExportPlotToImageButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+interface Props {
+  filename?: string;
+  toImage: ToImage;
+}
+
+export function ExportPlotToImageButton(props: Props) {
+  const { filename = 'plot', toImage } = props;
+  const ext = 'svg';
```
asizemore, line 21: Should this ext be based on `toImage` props?

asizemore: i s'pose we might not be able to do that since `toImage` is a function though

asizemore: oh maybe bc `toImage` is a function... i see

packages/libs/components/src/plots/ExportPlotToImageButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+        icon={Image}
+        themeRole="primary"
+        textTransform="none"
+        size="small"
+      />
+    </div>
+  );
+}
+
+async function downloadPng(
```
asizemore, line 43: what do you think about calling this function something more general like `downloadImage`, since we also use it to download svgs?

packages/libs/components/src/plots/ExportPlotToImageButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  toImage: ToImage,
+  filename: string,
+  ext: ToImageOpts['format']
+) {
+  const imgUrl = await toImage({ height: 450, width: 750, format: ext });
```
asizemore, line 48: just a note - initiall i was wary of setting the height and width because the bipartite networks have a height based on the number of nodes. However i gave it a shot with a super tall network and it worked great!

dmfalke:
> I forgot to say, another thing i like about the button and its placement is that it makes expanding download options easy. For example if we also had a button for download png, and one for download data, etc., it's obvious where those all go.

I was envisioning a dropdown with a "Download" or "Export" label. Thoughts?

asizemore:
> I was envisioning a dropdown with a "Download" or "Export" label. Thoughts?

That works for me!

dmfalke:
I pushed a change to allow the user to choose their export format:

image

packages/libs/components/src/plots/PlotlyPlot.tsx
dmfalke: There is a default in the button component: https://github.com/VEuPathDB/web-monorepo/blob/203e823ebbcb920f36cbc17846aebd2d3f1a3c62/packages/libs/components/src/plots/ExportPlotToImageButton.tsx#L20

packages/libs/components/src/plots/ExportPlotToImageButton.tsx
dmfalke: Yes, good catch! It's updated now

packages/libs/components/src/plots/ExportPlotToImageButton.tsx
dmfalke: Okay, great! We can make this more configurable later, if needed.

asizemore:
Looking good!
Testing now:
- boxplot: works great!
- bipartite network: I wonder if the height and width are getting us in trouble here. The png that's exported just shows a portion of the network. Also oddly the export button shows up way on the right. The svg export works great!
- volcano: looks great. The black box still appears in the svg but it's removable and all the plot still exists so i think it's fine. I googled a bit about it and it sounds like there's likely a transparency thing with that specific viz that exporting to svg doesnt like. It's possible it's also a browser thing but i got the same result with brave and safari. So who knows!
- mosaic: perfect!

Some screenshots:
Screen Shot 2024-01-31 at 6 56 52 AM
Screen Shot 2024-01-31 at 7 00 00 AM

I think at this point just the button placement and png output for the bipartite network are the only things that are worth updating if you don't mind! I'm still not worried about the black box on the volcano plot, since the underlying plot is still there.

I'm sorry for the slowness in review!

dmfalke:
Thanks for the review!

Regarding button placement for bipartite, this is what I see:

image

I still don't know how to get a bipartite graph, so I didn't look at the output images. Maybe you can share an analysis that has a good bipartite viz?

asizemore:
I should have also mentioned, to get the Generate Results button to show up, just modify any of those Prefilter Inputs. We introduced a bug in those prefilter inputs and i haven't had a chance to fix it yet

dmfalke:
I have made the following changes, which I hope address the immediate issues:

- Added a warning message for PNG
- Included error handling for the export button
- Modified some layout issues with the export button, including ensuring facet plots look okay

---

### Screenshots


**Facet plots**
image

**Error**
image

dmfalke:
I made an issue to look into PNG issues: https://github.com/VEuPathDB/web-monorepo/issues/818

asizemore (approved):
Good call on the facets!
I tested the network, boxplot, scatterplot, and lineplot this time.

While testing i noticed the following, but i think it's okay if these become new issues:
- the bipartite network png crops part of the plot
- the bipartite network's layout now pushes the plot over to the right of the legends instead of below the legends, until we shrink the screen
- the scatterplot svg doesn't capture points. I couldn't figure out why. I tried with a plot that only had a few points and still i got a rasterized image.

https://github.com/VEuPathDB/web-monorepo/assets/11710234/d9848cc1-40ee-4eb0-901f-a2008a77f2db



If you're happy with these becoming new issues, please let me know and i'll write them up!

dmfalke:
I wonder if the issue with scatter is that we're using webgl to render the points.

I'm not sure what changes would have impacted the layout for bipartite. I'll take a quick look.

Also, I forgot that I was going to look at changing dimensions for bipartite. 

I will spend a little bit of time trying to address these issues. I'll post an update by this afternoon.

dmfalke:
Okay, I pushed fixed for the issues you mentioned:

- bipartite image cropping
- bipartite viz legends not above plot
- scatter points in svg (for plots w/ < 1000 points)

Let me know what you think :crossed_fingers:

asizemore:
Thanks @dmfalke !

Scatterplots: check! Points are just an image for plots with more than 1000 points, and they are svg objects when plots have < 1000 points.

Bipartite network button placement: check! Looks good!

Bipartite network png: check! Also looks great!

lgtm! 🎸

PR #794, opened 2024-01-20
Little filter refactor

bobular:
This PR introduces a new hook `useLittleFilters` which components can use to easily get hold of referentially stable arrays of little filters, or big+little filters, for the specific combination of little filters needed.  Current types of little filter are
* 'viewport'
* 'time-slider'
* 'marker-config'
* 'selected-markers' (coming soon)

Due to the final checklist item below, some work was done on the filter-aware study variable metadata modification in `packages/libs/eda/src/lib/core/hooks/workspace.ts`. There can now be two filters for the same variable: one big filter and potentially another 'marker-data' little filter. These can get passed to `useStudyEntities` so the metadata modification now makes use of all filters.

Previously there was a `find` to get the first filter corresponding to the variable of interest. The vocabulary of the variable was set to the filter's `stringSet` (or date, number etc)

Now the vocabulary of the variable is filtered using the values in the filter's `stringSet` and this is repeated for all filters, this narrowing down to the most restrictive of the filters.

A similar process now applies to variable distribution ranges for range-based filters.

To dos:

- [x] fix variable change behaviour of marker-config LFs - need to fetch vocabulary/range metadata explicitly I think
- [x] check that marker-config LF is stable upon map drag
- [x] figure out backwards compatibility (littleFilters are only updated on _changes_) - even switching between marker modes is going to have this problem... (this was a problem with the initial approach, now abandoned)
- [x] test floaters when (hopefully) megastudy data is back to normal
- [x] remove `filtersIncludingViewport` prop
- [x] remove `analysisState` prop from `useLittleFilters` if no there's no anticipated need (I suppose it could get the regular `filters` from there, instead of from the dedicated prop at the moment?)
- [x] update Bar and Bubble marker modes
- [x] don't forget continuous variables for marker-config LFs
- [x] <s>do header entity counts with filters and `useEntityCounts`, not marker data - optimise `useEntityCounts` to filter on entities, if appropriate</s> won't do - leave this to sub-studies branch
- [x] figure out if we need marker-config LFs if there is already a big filter on the marker variable - answer: we probably do. the big filter could filter down to 'male' and 'female' and the marker config itself could have just 'male' checked by the user, so that filter needs to be applied too. yes we are doing this
- [x] filter-sensitive variable metadata wrangling (e.g. reduced vocabs on filtered cateogoricals) should maybe handle multiple filters on the same variable (see previous item) - an alternative is to put more specific filters first in the combined arrays returned by `useLittleFilters`. It's not just marker-configs, imagine filtering on date and also using a timeslider (can test with histogram floater) - decision was to apply all set filters in series to the vocabulary, and to merge range filters to narrowest (largest min and smallest max).
- [x] port TimeSliderQuickFilter to useQuery


packages/libs/eda/src/lib/map/analysis/littleFilters.ts
```diff
@@ -0,0 +1,50 @@
+import { Filter, StudyEntity, Variable } from '../../core';
+import { useMemo } from 'react';
+import { AppState } from './appState';
+import { GeoConfig } from '../../core/types/geoConfig';
+import { useDeepValue } from '../../core/hooks/immutability';
+import { VariableDescriptor } from '../../core/types/variable';
+
+export interface useLittleFiltersProps {
```
dmfalke, line 8: Please make this start with an uppercase letter

dmfalke (changes requested):
I think this looks great! I think the latest changes make a lot of sense. See my requested change below.

packages/libs/eda/src/lib/map/analysis/littleFilters.ts
bobular: All done - thanks for looking

bobular:
There's a bit of state/dependency handling I'd like to fix before merging. Switching between map modes causes a refresh of the timeline, which should be cached by react-query so I'll take a look...

bobular:
It wasn't using useQuery. Is now.

I found a relatively convenient way to port it. I could look at other usePromise code some time when I am twiddling my thumbs... (like never!)

dmfalke:
> It wasn't using useQuery. Is now.
> 
> I found a relatively convenient way to port it. I could look at other usePromise code some time when I am twiddling my thumbs... (like never!)

Awesome! Honestly, I'd like to port all `usePromise` cases to `useQuery`. Hmm... what if `usePromise` were to use `useQuery`? I doubt it's straightforward, due to the `queryKey` property, but would be nice.

dmfalke (approved):
Actually, I'm just going to approve this. Let me know if you want me to look at anything else. Nice work!

bobular:
Just double checked that the useQuery replacement is valid.

It's fine. The post-processing of the data uses only the data itself. If the post-processing had used other dependencies, it would have been more complicated. I'm pretty sure a lot of visualization response processing falls in the more complicated category, unfortunately!

Issue #777, opened 2024-01-17
Mesa - first column of tables w/ dragable columns cannot be clicked (to sort)


jernestmyers:
This appears to be a regression. I can click to sort the first column of a dragable and sortable Mesa table on the live site, but not when working off a local dev site from `main`. Note that all other columns remain sortable. This regression seems to only affect the first column of tables that allow users to drag/reorder the columns.

To reproduce:
- spin up a local `clinepi-site` dev server off the `main` branch
- navigate to the `Study summaries table` (either from the `Studies` header menu or from the study cards container)
- in the `Curated studies` table, try to change the sorting of the first column and nothing happens (note that other columns remain sortable)

dmfalke:
We should also make "locked" columns not moveable.

PR #796, opened 2024-01-22
Restore interactvitiy of Data Table sticky column headers

chowington:
Fixes https://github.com/VEuPathDB/web-monorepo/issues/777


jernestmyers (approved):
Nice, and so simple!

Issue #745, opened 2023-12-20
Map - add gear icons to bottom right of legend(s) that opens up marker configuration panel


bobular:
If the marker configuration panel is already open then maybe show a snackbar 'The marker configuration panel is already open'?

Not for this ticket, but we should think about it at the same time:
Floating plots ought to have something similar to open up the "Supporting plots" panel, but I don't think it should be a gear icon (because that would be confused with configuring the plot) - maybe some kind of list item in the header bar and we should probably use the plot's name in the floating plot DraggablePanel header.

Edit to add: or maybe a bigger overhaul of supporting plots should be discussed before adding something equivalent to the gear icons.

bobular:
Hi @moontrip - following on from https://github.com/VEuPathDB/web-monorepo/pull/743 I think we should change the default behaviour of "open the marker config panel when opening a new map".  I don't think we want this any more.  I think when we have the gear icons, it will be clear enough to the user how to customise the markers.  At the moment in a new analysis, the side panel opens, and then closes after the first "flyTo".

I assume this just involves setting `isSidePanelExpanded: false` in `defaultAppState` in appState.ts.

Issue #785, opened 2024-01-17
Map - refine 'landing behaviour' of shared/public analyses


bobular:
For April 2024 if possible.

Was just on a call with our tick data provider friends and they love the detailed analysis platform as researchers but would like it much simplified for their users who are general public, hospital and vet doctors and potentially policy makers.

Shared and/or public analyses are part of the answer to this.

If we can pre-configure views to show certain things, these can be linked from their website.

At the moment we have the following issues with imported analyses:

* left panel is open on marker configuration (unwanted)
* floating plot is not in the "hide controls" state we left it in
* floating plot is not in the position it was left in
* floating plot size is not what it was left in
* (legend position(s) also not preserved, but perhaps less of a problem)
* if the floating plot has a "proportion" y-axis and the floater's controls are hidden, there is no information to tell the user what that proportion is (the label is just "proportion")

This might be better as a parent ticket for smaller, well-defined tasks.

bobular:
Related to #275

bobular:
This is now mostly done in #275.
Need to think more about the floater, hidden controls, resizing etc before closing this completely I think, but great start.

bobular:
Moved remaining issues to b69 nice to have - and will require discussion first

PR #797, opened 2024-01-23
Add gear icon to SAM legend etc

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/745. Below is a list of works involved in this PR.

- [x] not to open map nav panel initially
- [x] making gear icon component
- [x] adding gear icon to all SAM legends (donut, barplot, and bubble markers)
- [x] properly arranging gear icon
- [x] mouse-over tooltip on the icon
- [x] functionality to open the map nav panel when pressing the icon
- [x] displaying snackbar when the nav panel is already open

Note: 
1) I thought that the icon was better placed at title area, not the bottom right. 
2) I used "warning" style for snackbar, but not quite sure if it is better than other styles: there are four categories such as "info", "success", "warning", and "error".  

A screenshot is attached:
SAM legend gear icon01


bobular (commented):
Hi. This is a great start. I have a few suggestions that will hopefully maintain maintainability!

`DraggablePanel` currently has higher level app-specific verbiage in it ("Marker configuration panel is already open") and an app-specific prop `isSidePanelExpanded`. It would be best if the `onClick` was just `onClick={openPanel}` and the `openPanel` callback defined in `MapAnalysis` could do all the snackbar logic. Then there's no need to pass `isSidePanelExpanded`.

I think it's OK for all `DraggablePanel`s to have an optional `SettingsButton` - I would rename its `openPanel` prop to `onSettingsButtonClick` and probably add a `settingsButtonTooltip` prop. But then we still have to pass two props around everywhere.

Would it be easier just to have a `headerButtons` prop to `DraggablePanel`? We would create the button `openMarkerConfigButton` in `MapAnalysis` which already has tooltip and onClick handler it needs. We would then just pass `openMarkerConfigButton` as a prop into the XXXMarkerMapType plugins.

We also need to set `activeSideMenuId` in the onClick handler so that the marker config panel actually opens. At the moment it will open the filter panel if that was the last thing opened.

moontrip:
Thank you for your review and detailed comments, @bobular 👍 I will check it out carefully next week. Have a great weekend!

moontrip:
I have addressed most of your feedbacks and suggestions. Please find my comments in the following that start from **DK:**

> `DraggablePanel` currently has higher level app-specific verbiage in it ("Marker configuration panel is already open") and an app-specific prop `isSidePanelExpanded`. It would be best if the `onClick` was just `onClick={openPanel}` and the `openPanel` callback defined in `MapAnalysis` could do all the snackbar logic. Then there's no need to pass `isSidePanelExpanded`.

**DK:** That makes sense. I have complied with.

> I think it's OK for all `DraggablePanel`s to have an optional `SettingsButton` - I would rename its `openPanel` prop to `onSettingsButtonClick` and probably add a `settingsButtonTooltip` prop. But then we still have to pass two props around everywhere.
> 

**DK:** This and below one are conflicted each other. Thus, I employed the next one

> Would it be easier just to have a `headerButtons` prop to `DraggablePanel`? We would create the button `openMarkerConfigButton` in `MapAnalysis` which already has tooltip and onClick handler it needs. We would then just pass `openMarkerConfigButton` as a prop into the XXXMarkerMapType plugins.
> 

**DK:** Made a headerButton component at MapAnalysis and passed it to DraggablePanel

> We also need to set `activeSideMenuId` in the onClick handler so that the marker config panel actually opens. At the moment it will open the filter panel if that was the last thing opened.

**DK:** I could manage to figure this out and now corresponding marker config panel is open regardless of the previous position.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -765,6 +765,56 @@ function MapAnalysisImpl(props: ImplProps) {
...
+    setIsSidePanelExpanded(true);
+  };
+
+  const openPanel = useCallback(() => {
+    appState.isSidePanelExpanded
+      ? enqueueSnackbar('Marker configuration panel is already open', {
+          variant: 'info',
+        })
+      : openPanelBehavior();
+  }, [
```
bobular, line 789: There are a bunch of "indirect" dependencies listed here.

I would probably be better to either create `openPanelBehavior` with useCallBack with the dependency [`appState.activeMarkerConfigurationType`] (the useState-derived setters are stable and not needed), and then add `openPanelBehavior` as a dependency for the `openPanel` useCallback

Or... Just paste the code from `openPanelBehavior` into `openPanel`, double-checking the necessity of the dependencies as mentioned above. This would be fine because I don't see `openPanelBehavior` used elsewhere.

bobular (commented):
This looks really nice now!

There's just one behavioural problem, where if the side panel is already expanded and on, say, the Filter panel, then it won't switch to the marker config panel when the gear icon is clicked.

A smaller problem is that if the marker config panel was last open in "supporting plots" sub-panel mode, then the marker config panel is opened in this mode, not the "markers" sub-panel.

A more sophisticated/flexible approach to UI state saving is probably needed long-term (see #785). Instead of `appState.activeMarkerConfigurationType` and non-persisted `activeSideMenuId` we probably need something like `appState.sidePanelNavigationState` that stores the full menu state, in a breadcrumb array or similar.  appState.isSidePanelExpanded could continue to be toggle.

Short-term, perhaps we can make it work with a function `isMarkerConfigOpen(appState.isSidePanelExpanded, appState.activeMarkerConfigurationType, activeSideMenuId)` ?  (not sure if the second arg is needed)

We can leave the problem of the "supporting plots" sub-tab to a wider-reaching ticket, such as #785

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
moontrip: Hi @bobular! Thank you for your detailed reviews 👍 Actually I noticed that indirect dependencies this morning and as you suggested I think we can just use single openPanel useCallback. Updated accordingly.

Also, I just addressed the first item, "There's just one behavioural problem, where if the side panel is already expanded and on, say, the Filter panel, then it won't switch to the marker config panel when the gear icon is clicked." It is now switched to marker configuration with snackbar behavior. I think that the other UI/UXs tuffs can be handled in other ticket #785 as you suggested.

bobular: Hi - that's still not quite the correct behaviour, IMO.

I was thinking more like this

```
  const openPanel = useCallback(() => {
    const nextSideMenuId = 'single-variable-' +
           (appState.activeMarkerConfigurationType === 'barplot'
          ? 'bar'
          : appState.activeMarkerConfigurationType);
    
    if (appState.isSidePanelExpanded && activeSideMenuId === nextSideMenuId)
      enqueueSnackbar('Marker configuration panel is already open', {
        variant: 'info',
      });
    if (!appState.isSidePanelExpanded)
      setIsSidePanelExpanded(true);
    if (activeSideMenuId !== nextSideMenuId)
      setActiveSideMenuId(nextSideMenuId);
  }, [
    appState.isSidePanelExpanded,
    enqueueSnackbar,
    activeSideMenuId,
    setIsSidePanelExpanded,
    appState.activeMarkerConfigurationType,
  ]);
```

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
```diff
@@ -786,5 +786,5 @@ function useMarkerData(props: DataProps) {
         boundsZoomLevel,
       },
     };
-  }, [rawMarkersResult, legendDataResult.data, overlayConfig]);
+  }, [rawMarkersResult, legendDataResult.data, overlayConfig, boundsZoomLevel]);
```
bobular, line 789: Good catch!

bobular (approved):
Almost what I was imagining!

I'm going to pre-emptively approve. Could you double check the logic I proposed? It works for me. Thanks!

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
moontrip: Thanks @bobular! 👍 So, you don't want to show snackbar if the side panel is already open and does not point out marker configuration panel (Donuts, Bar plots, or Bubbles), right? This is updated~

moontrip: Thank you for your confirmation, @bobular  👍 I will merge this PR, finally 🎉

PR #801, opened 2024-01-24
Amend text inputs for string params

### LINKED ISSUE #612: text box (stringParam) in go term and other searches
### Updates from 2024-01-23 (present: @dmfalke, @aurreco-uga, @jernestmyers)

We decided to use the `multiLine` parameter property to decide between `<input type="text">` and `<textarea>`. The updated rules are:

1. TextArea if `multiLine` is `true`
2. Retain current input box size for numbers
3. Double input box size for strings 

We will also need to update [StringParamFormatter](https://github.com/VEuPathDB/WDK/blob/4b5f8c8205edd5534358e0d61b723c6829cd37f9/Service/src/main/java/org/gusdb/wdk/service/formatter/param/StringParamFormatter.java) to include the `multiLine` property. An issue will be opened shortly.

---
This todo was raised by steve while working on teh goterm query,  an update to the goterm SQL:
https://epvb.slack.com/archives/C03PTKWGUPM/p1675972858632629

If we only allow one ID or a wildcard expression in the textbox/textarea, we need to update the parameter help, ask Susanne. 

### Older Description

FUT Nov 7 2023:     https://epvb.slack.com/archives/GBDR4LDNC/p1699373317247839
model defines a length.  add regex: disallow newline
client cutoff: 100 (now 50) for the ecnumber textarea

client rules for string param:    
1. text area if length > 100
5. retain current text box size for numbers
6. otherwise double the text box size

jernestmyers:
Resolves #612

### DIFF
--- a/packages/libs/wdk-client/src/Views/Question/Params/StringParam.tsx
+++ b/packages/libs/wdk-client/src/Views/Question/Params/StringParam.tsx
@@ -25,15 +25,7 @@ function Component(props: Props<StringParam, undefined>) {
   const { parameter, value, onParamValueChange } = props;
   const cols = DEFAULT_COLS;
   const rows = calculateRows(parameter, cols);
-  return parameter.length <= 50 ? (
-    <TextBox
-      type="text"
-      value={value}
-      readOnly={parameter.isReadOnly}
-      onChange={onParamValueChange}
-      required={!parameter.allowEmptyValue}
-    />
-  ) : (
+  return parameter.isMultiLine ? (
     <TextArea
       cols={cols}
       rows={rows}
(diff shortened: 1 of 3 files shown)


packages/libs/wdk-client/src/Views/Question/Params/StringParam.tsx
```diff
@@ -25,22 +25,23 @@ function Component(props: Props<StringParam, undefined>) {
...
-      cols={cols}
-      rows={rows}
-      readOnly={parameter.isReadOnly}
+    <TextBox
+      type="text"
       value={value}
+      readOnly={parameter.isReadOnly}
       onChange={onParamValueChange}
       required={!parameter.allowEmptyValue}
+      size={40}
```
dmfalke, line 44: If `parameter.isNumber` is `true`, set this to `20`.

dmfalke (changes requested):
See the requested change below. It will necessitate a change in `QuestionDecoders.ts`, and requires completion of https://github.com/VEuPathDB/WDK/pull/78.

packages/libs/wdk-client/src/Views/Question/Params/StringParam.tsx
dmfalke: I realized that I never made it clear that `StringParam` can also be used for numbers, even though we do have a `NumberParam`. It's possible `isNumber` is not used anymore.

dmfalke (approved):

PR #805, opened 2024-01-25
remove absolute abundance collections via annotations

### LINKED ISSUE #804: Mbio: Re-remove absolute abundances from dropdowns
on qa:

Screen Shot 2024-01-25 at 6 54 01 AM

The correct behavior is to only see one of each taxonomic level. We're seeing both because somehow absolute abundances have slipped back into the dropdown!

To complete this issue, remove the absolute abundances from the dropdowns for all mbio apps.

asizemore:
Resolves #804 

During a previous PR I simplified the `isNotAbsoluteAbundance` logic but ended up introducing a bug. Turns out the display name does not include "absolute" ever! Instead of reverting to the previous logic, which still included looking up by display name, this PR uses the hard core collection annotation properties to check if a collection is for absolute abundance. Reminder, those properties are set [here](https://github.com/VEuPathDB/ApiCommonData/blob/master/Load/ontology/General/collections/collections.yaml)

I've tested with a metagenomic study, a 16S study, and the FARMM study that has extra funky assays. All look good now 👍 
Third time's a charm!

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/Utils.ts
+++ b/packages/libs/eda/src/lib/core/components/computations/Utils.ts
@@ -71,10 +71,16 @@ export function removeAbsoluteAbundanceVariableCollections(
 export function isNotAbsoluteAbundanceVariableCollection(
   variableCollection: CollectionVariableTreeNode
 ): boolean {
-  return variableCollection.displayName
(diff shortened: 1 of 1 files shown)


dmfalke (approved):
Looks great!

PR #812, opened 2024-01-29
Network partitions api update

d-callan:
depends on https://github.com/VEuPathDB/EdaDataService/pull/347

### DIFF
--- a/packages/libs/components/src/plots/BipartiteNetwork.tsx
+++ b/packages/libs/components/src/plots/BipartiteNetwork.tsx
@@ -49,8 +49,10 @@ export interface BipartiteNetworkProps {
 
 // Show a few gray nodes when there is no real data.
 const EmptyBipartiteNetworkData: BipartiteNetworkData = {
-  column1NodeIDs: ['0', '1', '2', '3', '4', '5'],
-  column2NodeIDs: ['6', '7', '8'],
+  partitions: [
+    { nodeIds: ['0', '1', '2', '3', '4', '5'] },
+    { nodeIds: ['6', '7', '8'] },
+  ],
   nodes: [...Array(9).keys()].map((item) => ({
     id: item.toString(),
     color: gray[100],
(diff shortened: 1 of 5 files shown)


packages/libs/components/src/plots/BipartiteNetwork.tsx
```diff
@@ -107,18 +109,14 @@ function BipartiteNetwork(
   // nodes based on their column, then will use their order in the column
```
asizemore, line 109: mentions of "column" should be updated with "partition", or at least we should explain somewhere how they relate

packages/libs/components/src/plots/BipartiteNetwork.tsx
```diff
@@ -107,18 +109,14 @@ function BipartiteNetwork(
...
   // (given by columnXNodeIDs) to finally assign the coordinates.
```
asizemore, line 110: I checked out some of the other comments in this file and lots make references to the columns. To prevent confusion, we should update the wording to explain that it's a bipartite network, so it has two partitions of nodes. By default we'll render the nodes in a two-column layout. More options for layouts are a later feature.

packages/libs/components/src/stories/plots/BipartiteNetwork.stories.tsx
```diff
@@ -193,7 +193,6 @@ function genBipartiteNetwork(
   return {
     nodes,
     links,
-    column1NodeIDs,
-    column2NodeIDs,
+    partitions: [{ nodeIds: column1NodeIDs }, { nodeIds: column2NodeIDs }],
```
asizemore, line 196: For sake of completeness, I think it'd be good to change "columnXblah" to "partitionXblah" throughout this file as well.

asizemore (commented):
Tested with storybook so far. Looking good!
I think we should completely (or as completely as reasonable) remove references to columns. I think all my comments are about this. But the point is i think having boht "partitions" and "columns" will be confusing, and i'd rather use "partitions" when we can and explain that we're just by defualt rendering the partitions in two columns. There's no reason we have to draw the network in two columns - that was a choice we made.

PR #813, opened 2024-01-30
TidyTree as a React component

bobular:
This isn't working yet but I think it's close.

Firstly I followed much of Dave's work here https://github.com/VEuPathDB/web-monorepo/pull/559
But I wanted it to be a component, not something only working in ortho-site.

So I started with a super-simple story, and I'm working backwards from there.

0. `cd packages/libs/components`
1. `yarn add d3v5@npm:d3@5` - adds version 5 with the alias `d3v5`
2. `yarn add tidytree@github:d-callan/TidyTree`
3. `yarn add --dev imports-loader@^1.1.1 exports-loader@^1.1.1` - Storybook uses webpack 4, [this page](https://github.com/webpack-contrib/exports-loader/releases) suggests v1.1.1 is the last suitable for webpack 4
4. make crude `tidytree.d.ts` for now
5. add Dave's shimming to `.storybook/main.js`
6. make story
7. copy test data to story directory

Still doesn't work but are we close?

How do we get the shimming to work when the component is used outside of Storybook?


bobular:
tidytree margin specs are not `[ top, right, bottom, left ]` as indicated in the code, and it's difficult to work out what is going on. For now we will work with zero margins and live with the leaf nodes being chopped in half and the left-most edges not being very clear either.

bobular:
> tidytree margin specs are not `[ top, right, bottom, left ]` as indicated in the code, and it's difficult to work out what is going on. For now we will work with zero margins and live with the leaf nodes being chopped in half and the left-most edges not being very clear either.

Fixed these in our fork :-) https://github.com/d-callan/TidyTree/pulls

Now we have the vertical alignment bit pretty much sorted. Can start building a tree-table component next.

First I think I will nail down this TidyTree.tsx component to only do horizontal dendrograms - or wrap a horizontal dendrogram component around it. The `rowHeight` prop would make no sense in a vertical tree, for example. Not sure how general purpose TidyTree will need to be...

packages/libs/components/src/stories/tidytree/TreeTable.stories.tsx
bobular, line 1: There must be a better way of creating an emotion style and using its className directly?

bobular: Solved!

bobular:
Would the props be better split into `treeProps` and `tableProps` rather than extending the tree props as done presently?

packages/libs/components/src/stories/tidytree/TreeTable.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  },
+  {
+    leafId: 'Mouse',
+    accession: 'mus0738',
+    length: 150,
+    domains: 'ABC DEF XYZ',
+  },
+];
+
+const tableColumns: MesaColumn<LeafRow>[] = [
```
d-callan, line 68: idk enough to know... is it easy to add a column of check boxes?

d-callan (approved):
mostly i just played w storybook, but i think its looking good! :smile_cat:  I notice there isnt a dendrogram story w the branches colored as well, but just looking at the component api quickly suggests it should be possible. :+1:

packages/libs/components/package.json
```diff
@@ -71,6 +73,7 @@
     "@babel/core": "^7.17.9",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.16.7",
     "@babel/preset-env": "^7.16.11",
+    "@emotion/css": "^11.11.2",
```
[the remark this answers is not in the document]
dmfalke: yeah, it probably does

packages/libs/components/src/stories/tidytree/TreeTable.stories.tsx
bobular: Thanks for having a look. There's a separate story in "HorizontalDendrogram" with some non-table checkboxes that colour the tree.

I should probably wire up the TreeTable to use table-based checkboxes. Would be a good thing for the story.

bobular:
> Would the props be better split into `treeProps` and `tableProps` rather than extending the tree props as done presently?

I did this while adding the checkbox selection, which seems to work fine.

image

bobular:
I factored out the shimming here https://github.com/VEuPathDB/web-monorepo/pull/813/commits/c62b6222de8b23f606c35807884e1fd0810eb596

Hopefully this will be visible to other packages as `const { addD3Shimming } = require('@veupathdb/components/webpack-shimming');` but we will see! I might have to move it into somewhere that gets packaged up?

PR #819, opened 2024-02-02
allow wgcna collections in correlation metadata app

asizemore:
This PR includes changes that let folks see the wgcna data working in the correlations apps.

To run, use jbrestel plasmo backend.

02/02/24 - can select the collections for the correlation app. Found issue on the backend https://github.com/VEuPathDB/microbiomeComputations/issues/81

02/09/24 - all resolved. Data looks good and works!

## To Do before asking for review
- [x] Change the hard-coded assay entity ids to a function that just finds the entities with collections. 

For testing, please check that there are two collections options in the Correlation (assay x metadata) app.

I've left the Functional Associations correlations app out for now. There are open questions about how to handle that app.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/Utils.ts
+++ b/packages/libs/eda/src/lib/core/components/computations/Utils.ts
@@ -75,12 +75,11 @@ export function isNotAbsoluteAbundanceVariableCollection(
   // 1. normalizationMethod = NULL
   // 2. isCompositional = true
   // 3. isProportion = false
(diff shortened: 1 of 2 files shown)


d-callan:
backend work relating to this is now merged

dmfalke:
The code changes look good. I'm about to test.

In the mean time, I'll ask: Is there a plan for finding assay entities in a different way? When should we expect the approach in the PR to no longer work?

PR #821, opened 2024-02-02
Wire marker selection to floaters

### LINKED ISSUE #432: Map - have floating plots respond to selected markers
See #431 

When no markers are selected, floating plots use the whole viewport

When markers are selected, floating plots use the viewport AND a filter on geohash_x using the values of the selected markers.

The floaters should inform the user (perhaps in the floater title) what is being plotted (all visible data or selected markers).

bobular:
Resolves #432 

Currently branched off #794 

First off, as suggested by Dave months ago, I have moved the event handling of "single click to unselect all" out of MapVEuMap and into SemanticMarkers.  This sets the scene for wiring up selected markers to appState per marker mode (which makes sense, because some marker modes might have no markers!)

<s>Testing the floaters is tricky - there's a performance issue with sample-level variables, and assays are affected by this https://github.com/VEuPathDB/EdaDataService/issues/351 However, for PoC you can make a floating barplot on "continent" and click around to see what happens quite quickly.</s> SOLVED :heavy_check_mark: 

- [x] check story still works
- [x] wire up donut mode
- [x] and bar and bubble
- [x] and wire to floaters
- [x] address double data-requests made by histogram
- [x] think more about UX and feedback for users (what if they select markers with no floater - nothing will really happen) - and then just implement a straw man and have a meeting
- [x] move snackbars to top
- [x] make multiple select require shift-click
- [x] add extra snackbar to explain shift-click for multiple select
- [x] fix flashing popups after selection
- [x] increase mouse wheel sensitivity
- [x] roll out to other marker modes and viz types

image

### DIFF
--- a/packages/libs/components/src/map/BoundsDriftMarker.tsx
+++ b/packages/libs/components/src/map/BoundsDriftMarker.tsx
@@ -290,9 +290,12 @@ export default function BoundsDriftMarker({
             setSelectedMarkers(
               selectedMarkers.filter((id: string) => id !== props.id)
             );
-          } else {
-            // select
+          } else if (e.originalEvent.shiftKey) {
+            // add to selection if SHIFT key pressed
             setSelectedMarkers([...(selectedMarkers ?? []), props.id]);
+          } else {
+            // replace selection
+            setSelectedMarkers([props.id]);
           }
         }
 
--- a/packages/libs/components/src/map/BoundsDriftMarker.tsx
+++ b/packages/libs/components/src/map/BoundsDriftMarker.tsx
@@ -1,5 +1,5 @@
 import { useMap, Popup } from 'react-leaflet';
-import { useRef, useEffect, useState } from 'react';
+import { useRef, useEffect, useCallback } from 'react';
 // use new ReactLeafletDriftMarker instead of DriftMarker
 import ReactLeafletDriftMarker from 'react-leaflet-drift-marker';
 import { MarkerProps, Bounds } from './Types';
@@ -15,7 +15,7 @@ export interface BoundsDriftMarkerProps extends MarkerProps {
   // selectedMarkers state
   selectedMarkers?: string[];
   // selectedMarkers setState
-  setSelectedMarkers?: React.Dispatch<React.SetStateAction<string[]>>;
+  setSelectedMarkers?: (selectedMarkers: string[] | undefined) => void;
 }
 
 /**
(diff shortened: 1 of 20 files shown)


packages/libs/components/src/map/BoundsDriftMarker.tsx
```diff
@@ -281,30 +281,30 @@ export default function BoundsDriftMarker({
...
-          setSelectedMarkers((prevSelectedMarkers: string[]) =>
-            prevSelectedMarkers.filter((id: string) => id !== props.id)
-          );
-        } else {
-          // select
-          setSelectedMarkers((prevSelectedMarkers: string[]) => [
-            ...prevSelectedMarkers,
-            props.id,
-          ]);
+  const handleClick = useCallback(
```
bobular, line 284: useCallback doesn't seem essential but seems sensible given it is debounced and this may avoid the debouncing being reset unnecessarily.

bobular:
How about just having a yellow border around the floating plot when it is showing data from selected markers?

I think we probably need a Map UX meeting this Thurs.

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -463,73 +458,70 @@ function HistogramViz(props: VisualizationProps<Options>) {
...
-        series: [
-          distributionResponseToDataSeries(
-            'Subset',
-            distribution,
-            red,
-            NumberVariable.is(xAxisVariable) ? 'number' : 'date'
-          ),
-        ],
-      };
+  const distributionDataPromise = usePromise(
```
bobular, line 461: removed the double-nested useCallback

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -560,13 +552,6 @@ function HistogramViz(props: VisualizationProps<Options>) {
       )
         return undefined;
 
-      // wait till distributionDataPromise is ready
```
bobular, line 563: There was no need to wait for this - it was indirectly causing a double data request due to `distributionDataPromise.*` being added as dependencies.

bobular:
Sorry, will roll out new logic to all modes/viz and un-draft tomorrow.

bobular:
All good for review now!

packages/libs/eda/src/lib/core/components/visualizations/implementations/BarplotVisualization.tsx
```diff
@@ -902,6 +903,7 @@ function BarplotViz(props: VisualizationProps<Options>) {
       .every((reqdVar) => !!(vizConfig as any)[reqdVar[0]]);
 
   const LayoutComponent = options?.layoutComponent ?? PlotLayout;
+  const plotSubtitle = options?.getPlotSubtitle?.();
```
bobular, line 906: see standaloneVisPlugins.ts

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
bobular: Though waiting for it probably avoided the runtime bug that I later [fixed](https://github.com/VEuPathDB/web-monorepo/pull/821/commits/87b7ae9f5875372b093ddd1e2c4cba8054110bbb).

moontrip (approved):
Overall, it worked fine and I couldn't find specific issues after careful tests. Also I didn't notice specific issues on codes. :+1: 

One note is that there still exists handling null of binslider/binspec on Histogram Viz. Will try to use similar approach used in https://github.com/VEuPathDB/web-monorepo/pull/826. But this does not block the approval of this PR as it will be handled in the other ticket/PR

Issue #817, opened 2024-02-01
Very long variable names make for a funky variable coverage table


asizemore:
from the ablevins backend. Ranked abundance app with box or scatter viz. Choose data = Mass spectrometry assay

<img width="900" alt="Screen Shot 2024-02-01 at 8 05 14 AM" src="https://github.com/VEuPathDB/web-monorepo/assets/11710234/ebc9f8c4-dca4-4828-8040-a3c86f5df8cf">

To prevent the text from running off the side of the table, i think we just wrap, wrap, wrap those variable display names. Open to other ideas!

PR #823, opened 2024-02-05
break word for long variable name for coverage table

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/817

The culprit in the ticket's screenshot was a very long variable name without specific separation in words, fructose/glucose/galactose:HIL-neg: Metabolite. So, if we want to keep all table contents inside the table, then we need to use a css, word-break. Screenshot is attached with this PR.

mass table fix01

(update) I didn't like the style for Data column so I made some changes. Here is new one:
mass table fix02


asizemore (approved):
Tested with BONUS study. Looks great! Thanks @moontrip !

moontrip:
Thank you for your review and confirmation, @asizemore 👍

PR #824, opened 2024-02-05
capitalize text, Ok

moontrip:
This will address https://github.com/VEuPathDB/web-components/issues/435. 
Super easy fix 😅


asizemore (approved):
Seems good to me! Was this the only example of "Ok" we had?

moontrip:
> Seems good to me! Was this the only example of "Ok" we had?

Thank you for your quick review 👍 Based on Bob's comment at https://github.com/VEuPathDB/web-components/issues/435#issue-1522514612, it seems to be the only one.

Issue #265, opened 2022-07-25
EDA  Ranked abundance: clarify plot subtitle text


asizemore:
Feedback from @dpbisme : The plot subtitle isn't very clear. Not sure what "variables" are.

Proposal: change to "Ranked abundance: X-axis variables with ..." for boxplot and "Ranked abundance: Overlay variables with... " for scatter

dmfalke:
The configuration component is passed the `computation` and `visualizationId`, so it should be able to look up the visualization and figure out its type: https://github.com/VEuPathDB/web-monorepo/blob/7f136ad8de0eb6315496f417a205a7b1a92c00ec/packages/libs/eda/src/lib/core/components/computations/Types.ts#L18-L23

asizemore:
Brilliant! Thanks @dmfalke !!

asizemore:
Update:
The number in the plot subtitle for ranked abundance box and scatter has been fixed. So the scope of this ticket is now to just change "Ranked abundance: Variables with..." to "Ranked abundance: X-axis variables with..." for the box plot and to "Ranked abundance: Overlay variables with..." for the ranked abundance scatterplot. Hoping for a quick and easy update!

PR #825, opened 2024-02-05
change plot subtitles for Ranked abundance

moontrip:
This will resolve https://github.com/VEuPathDB/web-monorepo/issues/265. 

It is obvious what to do, though I could not find where the plot subtitles are displayed 😅

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
@@ -58,8 +58,8 @@ export const plugin: ComputationPlugin = {
             <>
               <br />
               <span>
-                Ranked abundance: Variables with {config.rankingMethod} = 0
-                removed. Showing up to the top ten variables.
+                Ranked abundance: X-axis variables with {config.rankingMethod} =
+                0 removed. Showing up to the top ten variables.
               </span>
             </>
           );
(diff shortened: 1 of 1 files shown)


asizemore:
haha no worries! They are very subtle :) Here's one using your branch!

Screen Shot 2024-02-06 at 6 06 48 AM

packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
```diff
@@ -87,8 +87,8 @@ export const plugin: ComputationPlugin = {
             <>
               <br />
               <span>
-                Ranked abundance: Variables with {config.rankingMethod} = 0
-                removed. Showing up to the top eight variables.
+                Ranked abundance: Overlay variables with {config.rankingMethod}{' '}
```
asizemore, line 90: Is it possible to make the two subtitles more similar by having this part be `...with {config.rankingMethod} = 0 removed...`? Or alternatively can we change the subtitle at lines 61/62 to be `with {config.rankingMethod}{' '}= 0 removed`?

asizemore (commented):
One quick question! Otherwise i tested and it looks good!

packages/libs/eda/src/lib/core/components/computations/plugins/abundance.tsx
moontrip: @asizemore I found it interesting. I have tested and realized that actually that part, {' '}, is automatically added by pre-commit tool(s) (probably prettier). So I cannot change or add that part by myself: it is treated as empty commit. 

Thus we just need to leave it as is

asizemore: wild. Good to know!

PR #826, opened 2024-02-06
nicely handle empty data for visualizations in SAM

moontrip:
This will address https://github.com/VEuPathDB/web-eda/issues/1631.

Following on from some changes in back end responses (see #836) we are revisiting the behaviour of floating plots when there is nothing to plot. A key background point here is that the SAM floating plot endpoints (e.g. histogram, lineplot etc) don't return `completeCasesAllVars` and `completeCasesAxesVars` for performance reasons and because missing data is not handled the same as in EDA. These two values are used to calculate `outputSize` which is fed to `PluginError` to produce an appropriate warning banner when it is zero.

In addition we have relaxed `io-ts` validation to allow nulls for binSlider and binSpec when there is an empty response from the back end (but no 204 or 500). This happens when there is a subset, but the plot variables are absent from that subset.

### old PR notes

The error message(s) indicates type errors of backend response. I think that null returns of binSlider/binSpec in Lineplot Viz means empty data return, thus relevant types are loose to accommodate null values. After this change, the error message is the same with what we could see at Scatterplot Viz

lineplot error message02

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -613,6 +613,7 @@ function HistogramViz(props: VisualizationProps<Options>) {
         facetVariable?.vocabulary,
         facetVariable
       );
+
       return grayOutLastSeries(
         substituteUnselectedToken(
           reorderData(
@@ -650,6 +651,18 @@ function HistogramViz(props: VisualizationProps<Options>) {
       facetEntity,
     ])
   );
+
+  // controls need the bin info from just one facet (not an empty one)
+  const checkData = isFaceted(data.value)
+    ? data.value.facets.find(
+        ({ data }) => data != null && data.series.length > 0
+      )?.data
+    : data.value;
+
+  // we can't always rely on data.value?.completeCasesXXXVars (e.g. in SAM)
+  // in the outputSize determination above, so we make a simple check the data (or one non-empty facet)
+  const isEmptyData =
+    checkData?.series.find((series) => series.bins.length > 0) == null;
 
   // Note: Histogram distribution data contains statistical values such as summary.min/max,
   // however, it does not fully respect multiple filters.
@@ -1375,7 +1388,10 @@ function HistogramViz(props: VisualizationProps<Options>) {
         )}
       </div>
 
-      <PluginError error={data.error} outputSize={outputSize} />
+      <PluginError
+        error={data.error}
+        outputSize={isEmptyData ? 0 : outputSize}
+      />
 
       <div style={{ display: 'flex', flexDirection: 'column' }}>
         {!hideInputsAndControls && (
@@ -1411,9 +1427,6 @@ export function histogramResponseToData(
   overlayVariable?: Variable,
   facetVariable?: Variable
 ): HistogramDataWithCoverageStatistics {
-  if (!response.histogram.data.find((item) => item.value.length > 0))
-    throw Error(`Expected one or more data series, but got zero`);
-
   const facetGroupedResponseData = groupBy(response.histogram.data, (data) =>
     data.facetVariableDetails && data.facetVariableDetails.length === 1
       ? fixLabelForNumberVariables(
@@ -1482,9 +1495,11 @@ export function histogramResponseToData(
   });
 
   return {
-    // data
-    ...(size(processedData) === 1 &&
-    head(keys(processedData)) === '__NO_FACET__'
+    // to avoid erronous assignment to facetedData for empty data reponse
+    // need to check if size(processedData) === 0 condition has any side effect
+    ...((size(processedData) === 1 &&
+      head(keys(processedData)) === '__NO_FACET__') ||
+    size(processedData) === 0
       ? // unfaceted
         head(values(processedData))
       : // faceted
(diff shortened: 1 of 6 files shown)


bobular:
Maybe fix histogram too?

asizemore (approved):
Thanks for taking this @moontrip !

Works great! Tested with a different study.

I didn't quite understand one thing - why not change `BinSpec` and `BinSlider` to accommodate nulls?

moontrip:
> Thanks for taking this @moontrip !
> 
> Works great! Tested with a different study.
> 
> I didn't quite understand one thing - why not change `BinSpec` and `BinSlider` to accommodate nulls?

Thank you for your review and test, @asizemore 👍 
I didn't work for typing Lineplot Viz initially so I don't know why they didn't accommodate nulls. My pure guess is that at the beginning of the implementation of the Lineplot Viz, we wanted to check whether the backend returned the response correctly even if data existed.

moontrip:
@bobular: it would be great if you can review this PR. 
This PR will also address https://github.com/VEuPathDB/web-monorepo/issues/842. Refactored this PR to include Histogram Viz as well. Used the pre-existing message text at Histogram Viz.

- examples: Vectorbase Mega study

SAM histogram null04

SAM historgram null03

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -1411,7 +1411,17 @@ export function histogramResponseToData(
...
+  // dataExist before filter() will have an array of booleans, e.g.,  [true, false, true,...]
+  // if data does not exists, the array consists of all false values
+  // filtering out true cases, then no data  = dataExist.length === 0
+  const dataExist = response.histogram.data
+    .map((item) => {
+      return item.value.length > 0;
+    })
+    .filter((element: boolean) => element);
+
+  if (response.histogram.data.length === 0 || dataExist.length === 0)
```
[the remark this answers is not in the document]
moontrip: Thank you for your suggestion, @bobular 👍 Complied!

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -1411,7 +1411,7 @@ export function histogramResponseToData(
   overlayVariable?: Variable,
   facetVariable?: Variable
 ): HistogramDataWithCoverageStatistics {
-  if (response.histogram.data.length === 0)
+  if (!response.histogram.data.find((item) => item.value.length > 0))
     throw Error(`Expected one or more data series, but got zero`);
```
bobular, line 1415: This is an old error message that was never/rarely seen by users mainly because null values were causing `io-ts` to complain first.

I have removed this test entirely and found that nothing bad happens. However there is an empty plot with no explanation to the user. Which is bad. However, the `PluginError` component should handle this for us, but `outputSize` being passed to it is `undefined`.  We could either ensure that `outputSize` is zero before it is passed, or change the login in `PluginError` to this:

```diff
diff --git a/packages/libs/eda/src/lib/core/components/visualizations/PluginError.tsx b/packages/libs/eda/src/lib/core/components/visualizations/PluginError.tsx
index 51956105a3..6f8a1ea0b8 100644
--- a/packages/libs/eda/src/lib/core/components/visualizations/PluginError.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/PluginError.tsx
@@ -23,7 +23,7 @@ export default function PluginError({ error, customCases, outputSize }: Props) {
   const fallbackErrorMessage =
     error == null ? '' : error instanceof Error ? error.message : String(error);
 
-  const emptyCase = () => (outputSize === 0 ? emptyCaseMessage : undefined);
+  const emptyCase = () => (!outputSize ? emptyCaseMessage : undefined);
 
   const errorContent =
     (customCases ?? [])
```

I think I prefer the latter.

bobular (approved):
Please remove the test/throw for empty histogram data and modify `PluginError` as suggested.  Any questions, just shout!

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: I realized that the latter did not work. It is because SAM's histogram always return outputSize = undefined. Thus, if we want to use PluginError, then we need to add an inline condition before passing to PluginError

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -651,6 +652,18 @@ function HistogramViz(props: VisualizationProps<Options>) {
     ])
   );
 
+  // controls need the bin info from just one facet (not an empty one)
+  const checkData = isFaceted(data.value)
```
bobular, line 656: Need to get rid of `data0` below and replace it with `checkData`

moontrip:
@bobular I have implemented the logic we discussed at Slack. It took more time than I expected to address empty plot issue. In one case, there is empty data response, not null case. In that case, Histogram Viz treated it as facetedData, so I needed to add a condition to handle it as not. It worked fine for me but it may also cause an edge case. 

With all various use cases that we cannot test all of them, honestly I am not quite sure if this is better than previous logic. But anyway this seems to work fine for SAM's Histogram VIz.

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: oh I see. will do that

moontrip: done!

bobular:
Didn't do any special `isEmptyData` handling for the contingency table. After relaxing the io-ts constraints, it renders with zeroes and NaN% just fine, in my opinion

image

jernestmyers (approved):
I didn't test on a dev site, but the code looks okay to me. 💪

asizemore:
tested boxplots and lineplots on mbio. Looks great!!

asizemore (approved):
I had a few questions but nothing serious!

Issue #808, opened 2024-01-26
Volcano plot: handle no data gracefully


asizemore:
Found on qa. 

<img width="1479" alt="Screen Shot 2024-01-26 at 7 22 39 AM" src="https://github.com/VEuPathDB/web-monorepo/assets/11710234/7ac37203-7861-43c3-acef-70f8411f816c">

Instead of an empty plot, the volcano should say something nice like "no data returned". 

Note that this happens because no data was returned from the backend. There is a matching [ticket](https://github.com/VEuPathDB/microbiomeComputations/issues/80) around how to handle this case in microbiomeComputations as well.

asizemore:
We did something [similar](https://github.com/VEuPathDB/web-monorepo/pull/774) in the bipartite network/correlation case. The difference here is that changing those Threshold lines don't affect the data returned from the backend, so there's no need to make those inputs yellow.

PR #827, opened 2024-02-06
Add message for no data in volcano plot

chowington:
Closes https://github.com/VEuPathDB/web-monorepo/issues/808

Most of the diff is just indentation!

### DIFF
--- a/packages/libs/components/src/plots/VolcanoPlot.tsx
+++ b/packages/libs/components/src/plots/VolcanoPlot.tsx
@@ -1,7 +1,6 @@
 import {
   CSSProperties,
   forwardRef,
-  ReactNode,
   Ref,
   useCallback,
   useImperativeHandle,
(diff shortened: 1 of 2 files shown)


asizemore (approved):
lgtm!

dmfalke:
I may be wrong, but isn't this different than how we handle empty data in other vizs? Eg, https://github.com/VEuPathDB/web-monorepo/pull/826

asizemore:
Huh. You're totally right. For the network the user could do something about the no data, so it made sense to give them a message. But this volcano viz getting no data is just like any other plot getting no data. I had my mbio-app blinders on! Good call, @dmfalke . 
The other vizs just show the red banner saying something like "no plottable data". That makes sense here, too.

Sorry @chowington that's my fault!

asizemore (approved):
lovely! Works great! Thanks for handling the changeup @chowington !

PR #830, opened 2024-02-07
Fix bug with grid rows and breakpoints

jernestmyers:
Our breakpoints targeting the grid layout are off. On the live site, notice how changing between the workspace tabs for all datasets to new upload jumps the page:

https://github.com/VEuPathDB/web-monorepo/assets/69446567/8a14a6f6-a6ba-4718-954e-770d605356dd

I reverted an old change I made to `default-layout` that added the `!important` attribute. It was a hack that masked other problems. Somehow it seems Ortho is using just the default layout grid, even on the homepage, without issue.

As for genomics, I made specific changes to breakpoint logic that fixes the original issue that the `!important` attribute aimed to fix, which was the "ghost" effect of the hamburger menu rendering without the site header's background behind it, and it fixed the jumping effect that prompted this investigation. I also moved a breakpoint that is intended to target the homepage grid areas but was actually being applied sitewide.

### DIFF
--- a/packages/libs/web-common/src/styles/default-layout.scss
+++ b/packages/libs/web-common/src/styles/default-layout.scss
@@ -36,7 +36,7 @@ body.vpdb-Body {
     grid-template-rows: $expanded-header-height auto 1fr auto;
 
     @media screen and (max-width: $hamburger-width) {
-      grid-template-rows: auto auto 1fr auto !important;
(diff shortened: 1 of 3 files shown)


dmfalke (approved):
From what I can tell, everything looks good. Nice work!

Issue #829, opened 2024-02-07
bugs on EDA viz buttons


moontrip:
While working on other EDA-related tickets, I noticed that there were two bugs concerning behaviors of viz buttons. One is close (x) button at 'Select a Visualization', and the other is minimize (_) button at each Viz. It appears that they are related to routing.

I checked Live site (ClinEpi) but it worked okay while qa site also had the same issues.

PR #831, opened 2024-02-07
fix bugs on some EDA viz buttons

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/829.

Although this would work ok, I have a question, @dmfalke. For the second Lin/pathname at Minimize Visualization, I have no idea what the role of computationId is in there. I tried both ClinEpi and Microbiome site to check the value of computationId in there, but it is not that meaningful value for the behavior of the "Minimize Viz": either pass-through or a random word (mbio). So, I think that it can be simply used as the following. What do you think?

`pathname: '../visualizations',`


dmfalke (approved):
We used to include the `computationId` as a part of the url, so your changes look good. Thanks!

PR #834, opened 2024-02-07
SAM mega study details

### LINKED ISSUE #275: Map - save locations of draggable elements (floater and legend) in user analysis
Remember and persist the locations and size of floating visualizations.

### LINKED ISSUE #696: Map - sub-studies information for megastudies (issue WIP)
This will change the top-right part of the header

Mockups are [here](https://www.figma.com/file/gKYVrWu6qfDTUgYKQuvaIy/SAM?type=design&node-id=1926-42&mode=design&t=TZqeVOCAs4PT7R7D-0) (but screenshots below tell the whole story)

### Detecting megastudy vs regular study

The client asks if the root entity is a GeoEntity
* yes = regular study
* no = megastudy

This might not always hold true but it will avoid us having to hard-code in the study entity ID.

### Header for regular studies

We could either keep the current mini table (All, Filtered, View counts) or go for a new streamlined look:

image

This would be equivalent to the current "View" count.

The ℹ️ icon would provide the following nicely formatted tooltip text (between the horizontal rules):

---
1,234 <output entity name(s)> are currently visualized on the map using markers. These are the <output entity name(s)> that
- satisfy all your filters (you currently have X filters active)
- to do: also time slider filter
- have data for the <variable name> variable that is currently displayed on the markers
- have appropriate values for <variable name> if the marker has been custom-configured
---

The "appropriate values" text could be spelled out for the different marker modes more explicitly if we like.

### Header for megastudies

image

ℹ️ icon text

---
1,234 <output entity name(s)> are currently visualized on the map using markers. These are the <output entity name(s)> that
- satisfy all your filters (you currently have X filters active)
- to do: also time slider filter
- have data for the <variable name> variable that is currently displayed on the markers
- have appropriate values for <variable name> if the marker has been custom-configured

The visualized data comes from 56 <root entity name(s)> [button="Show list"]

---

The `Show list` button would bring up a DraggablePanel on top of the map containing a simple clickable list of studies (see below). 

### How does the client get the number/list of relevant studies?

I think this will need to be a new request to (`studies/.../tabular`?) that includes 
- regular filters
- viewport filter
- time slider filter
- little filters generated from current marker mode configuration
  - default configuration, this will be `[MIN to MAX]` for continuous vars, and `value in [ALL VALUES]` for categorical
  - custom configuration for categoricals, `value in [DENOMINATOR VALUES]`

For the fields in the output, we ask for the first variable of this entity (ordered by displayOrder if available). This is supposed to be Study name for the megastudy.

I'm not liking the `[ALL VALUES]` for variables like species... (which ironically doesn't need it because there are no missing values for this variable)

Do we just make this request and use the number of rows/records for the header number? That way we can re-use the same request for the floating table below:

**Alternative approach?** Use map-markers endpoint and ask for outputEntity=Study

### The new floating table

A simple list of the tabular response as described above. Should respond to marker-selections.

If we're showing study names, are we linking using that, or the entity's row ID (the one that downloads displays)?

Links should go to the WDK record page. How are we doing that?

Will we need some hardcoding anyway to make sure we only make links when we have megastudy study entities?

dmfalke:
Closes #696
Partially addresses #275

This PR adds a panel for sub-studies visible on the map. It also shows how we can retain panel position and dimensions, in a generic way (see #275). We can apply that to all draggable panels in a follow up.

TODO:

- [x] Refine verbiage
- [x] Add new header component to all map types
- [x] Only show new header component in mega studies. This may need to be hard coded (as a top-level prop) for now.
- [x] Make sure the correct filters are being passed.
- [x] Figure out app state backwards compatibility (`studyDetailsPanelConfig` is a mandatory prop at the moment)

**Screenshot of current progress**

image


packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  if (
+    rootEntityCount.data == null ||
+    visibleEntityCount == null ||
+    outputEntity == null
+  )
+    return null;
+
+  const totalCountFormatted = format(visibleEntityCount);
+  const totalRootEntityFromatted = format(rootEntityCount.data.count);
```
bobular, line 58: "Fromatted" 😆

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
```diff
@@ -0,0 +1,125 @@
...
+          padding: '1em',
+          fontSize: '1.2em',
+          display: 'flex',
+          gap: '1ex',
+          alignItems: 'center',
+        }}
+      >
+        {format(visibleEntityCount)} {outputEntity.displayNamePlural} from{' '}
+        {format(rootEntityCount.data.count)} {rootEntity.displayNamePlural}
+        <Tooltip title={tooltipContent} interactive style={{ width: 'auto' }}>
```
bobular, line 104: The tooltip is rendering half off-screen unless my browser (Firefox) window is >1800 px wide. This means the "show list" button isn't visible. I'm not sure why, sorry!

bobular (commented):
Just made some comments on some draggable and tooltip positioning issues and figured out the debouncer. Will look more closely into the study-related stuff later.

packages/libs/coreui/src/components/containers/DraggablePanel/index.tsx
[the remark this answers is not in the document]
dmfalke: I discovered this as well, on Friday. I have a fix that I will push later today

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
dmfalke: This is because the legend is off screen. It will be fixed soon

bobular: Oh, I see. The legend is extending some kind of bounding box behind the scenes?

dmfalke:
I've wire up bar and bubble markers, too. There is an inconsistency with bubble markers. The proportion is acting like a little filter, whereas selecting values in the donut and bar do not act like little filters.

Oh, I see... the "denominator" is acting like a little filter. Hopefully this isn't too confusing.

dmfalke:
I've hard coded the study entity and variable IDs for now. We can figure out how to make that configurable later.

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
```diff
@@ -42,12 +42,12 @@ export function MapTypeHeaderStudyDetails(props: Props) {
   const outputEntity = entities.find((entity) => entity.id === outputEntityId);
 
   const studyEntityCount = useEntityCount(
-    STUDIES_ENTITY_ID, // Study entity
+    studyEntity?.id, // Study entity
```
bobular, line 45: Why was this better than just using the constant ID?

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
```diff
@@ -42,12 +42,12 @@ export function MapTypeHeaderStudyDetails(props: Props) {
...
     filterForVisibleData // Should be subset, map type, timeline, and viewport filters
   );
 
   const outputEntityCount = useEntityCount(
-    outputEntityId,
+    outputEntity?.id,
```
bobular, line 50: And the same here?

bobular (commented):
Couple of questions for my own benefit before I review the other recent commits!

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
```diff
@@ -242,7 +242,7 @@ function BubbleMapLayer(props: MapTypeMapLayerProps) {
         selectedMarkers,
       });
     },
-    [props.configuration, updateConfiguration]
+    [handleSelectedMarkerSnackbars, props.configuration, updateConfiguration]
```
bobular, line 245: oops

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
dmfalke: The hook handles undefined, which prevents requests when the entity does not exist

bobular (commented):
I'm pretty sure it's possible to get rid of all mentions of `totalVisibleWithOverlayEntityCount` in source files under `packages/libs/eda/src/lib/map/` (there is some legacy EDA map viz code that still uses it)

The same goes for `totalVisibleEntityCount` (and incidentally, it looks like these two numbers have been the same for some time)

packages/libs/eda/src/lib/map/analysis/SubStudies.tsx
```diff
@@ -0,0 +1,143 @@
...
+              type: 'error',
+              message: String(result.error),
+            }}
+          />
+        ) : result.data == null || result.isFetching ? (
+          <Spinner />
+        ) : (
+          <div>
+            <p>
+              There {studyCountPhrase(result.data.length - 1)} for the{' '}
```
bobular, line 110: I take it `result` comes with a header row?

packages/libs/eda/src/lib/map/analysis/SubStudies.tsx
```diff
@@ -0,0 +1,143 @@
...
+              {hasSelectedMarkers ? 'selected' : 'visible'} markers on the map.
+            </p>
+            <ul>
+              {result.data.slice(1).map(([id, display]) => (
+                <li>
+                  <Link
+                    target="_blank"
+                    to={{
+                      pathname: `/record/dataset/${datasetIdByStudyId[id]}`,
```
bobular, line 119: Does this need some safety checking?

packages/libs/eda/src/lib/map/analysis/littleFilters.ts
```diff
@@ -9,9 +14,15 @@ export interface UseLittleFiltersProps {
...
-  ) => { entity: StudyEntity; variable: Variable } | undefined;
+}
+
+export interface UseLittleFiltersFuncProps extends UseLittleFiltersProps {
+  findEntityAndVariable: (variableDescriptor?: VariableDescriptor) =>
+    | {
+        entity: StudyEntity;
+        variable: Variable;
+      }
+    | undefined;
```
bobular, line 25: Nice

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
```diff
@@ -0,0 +1,178 @@
...
+        currently visualized on the map using markers. These are the{' '}
+        {outputEntity.displayNamePlural} that
+        <ul css={{ padding: '1em 0' }}>
+          <li>satisfy all your filters</li>
+          {includesTimeSliderFilter && (
+            <li>satisfy the time range you have selected</li>
+          )}
+          <li>
+            have data for the variable that is currently displayed on the
+            visible or slected markers
```
bobular, line 110: slected

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
```diff
@@ -0,0 +1,178 @@
...
+        )}
+        <Tooltip title={tooltipContent} interactive style={{ width: 'auto' }}>
+          <Info style={{ color: '#069', height: '.8em', width: '.8em' }} />
+        </Tooltip>
+      </div>
+    </>
+  );
+}
+
+function useEntityCount(entityId?: string, filters?: Filter[]) {
```
bobular, line 161: did you mean to put this somewhere central?

bobular (approved):
LGTM!

packages/libs/eda/src/lib/map/analysis/SubStudies.tsx
dmfalke: Yep! I suppose this could use a comment

packages/libs/eda/src/lib/map/analysis/SubStudies.tsx
dmfalke: Ah, in case `datasetIdByStudyId[id]` is not defined? Good idea. What should we do in those cases?

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
dmfalke: Eventually, I would like to replace the other hook with this.

PR #836, opened 2024-02-08
Handle 204s in DataClient and FetchClient

### LINKED ISSUE #571: Map - handle empty map-markers responses elegantly
If you zoom into an area of the map with no markers, nothing bad happens and  `"mapElements": []` is returned from `map-markers`. Currently the map remains empty and nothing else is communicated to the user.

There are some cases in the map where `map-markers` is returning a 500 with (and the error message `Requested data stream 'single_tabular_dataset' did not contain any data`). This seems to be where a combination of big filters and little filters (due to marker variable choice or configuration) leads to no data.

In the regular EDA visualizations this is handled in PluginError.tsx - the error message is parsed for `"did not contain any data"` and a pink banner is displayed with "The visualization cannot be made because the current subset is empty."

There's another case handled by PluginError.tsx for a 200 response when one of `response.histogram.config.completeCasesAllVars` or `response.histogram.config.completeCasesAxesVars` is zero (for histogram).  Then the message is `"The visualization cannot be made because there is no plottable data for selected variable(s) in the current subset."`

Back to the map, with the VB megastudy, you can recreate the problem like this:

filter: insecticide = alpha-cypermethrin
markers: donuts with "Plasmodium falciparum infection status" as the variable.

This results in no map markers (after the main map spinner is shown) and an ever-spinning legend.

(Aside: The map behaviour is a little bit strange (I'm on Dave's [new refactor branch](https://github.com/VEuPathDB/web-monorepo/pull/424)) - each time I bring my browser tab back into focus, more map-markers requests are made (map spinner again))

I think the desired behaviour would be to present a message to the user "There is no plottable data for <marker variable> with current filters." and make sure the legend is not still spinning.

The message could possibly even go in the legend?

### LINKED ISSUE #606: `FetchClient` - Add a way to handle specific response status codes
Currently, `FetchClient` treats responses as being in one of two states: success and error. There are cases where it is desirable to handle specific response statuses, such as `200` vs `204`.

Need to develop an interface that is easy to use, yet extensible.

### LINKED ISSUE #607: EDA Visualizations (and map markers) - handle 204 response
~This depends on #606, and is needed for https://github.com/VEuPathDB/EdaDataService/pull/328.~

We just need to add `io-ts`'s `undefined` type as a possible response for the map-markers viz, and then map it to an empty array.

This is needed for https://github.com/VEuPathDB/EdaDataService/pull/328.

jernestmyers:
Depends on https://github.com/VEuPathDB/EdaDataService/pull/328
Resolves #606  
Resolves #607
Resolves #571

In the map, a "no data" response will render a warning banner. To do so, I tweaked the `MapFloatingErrorDiv` to accept `children` as a means to customize its content. Screenshot below:
image

I've also added the `Banner` component with `type: warning` to the `PluginError` messaging as shown:
image

### DIFF
--- a/packages/libs/eda/src/lib/core/api/DataClient/index.ts
+++ b/packages/libs/eda/src/lib/core/api/DataClient/index.ts
@@ -62,7 +62,7 @@ export default class DataClient extends FetchClientWithCredentials {
         transformResponse(body) {
           if (body == null) {
             throw new NoDataError(
-              'The visualization cannot be made because the current subset is empty.',
+              'The visualization cannot be made because no data remains after filtering.',
               'No data',
               204,
               uuid()
--- a/packages/libs/eda/src/lib/core/api/DataClient/index.ts
+++ b/packages/libs/eda/src/lib/core/api/DataClient/index.ts
@@ -1,4 +1,5 @@
 import { TypeOf, Decoder } from 'io-ts';
+import { v4 as uuid } from 'uuid';
 
 import {
   createJsonRequest,
@@ -34,6 +35,7 @@ import {
   StandaloneMapBubblesLegendRequestParams,
   StandaloneMapBubblesLegendResponse,
 } from './types';
+import { NoDataError } from './NoDataError';
 
 export default class DataClient extends FetchClientWithCredentials {
   getApps(): Promise<TypeOf<typeof AppsResponse>> {
@@ -59,11 +61,13 @@ export default class DataClient extends FetchClientWithCredentials {
         body: params,
         transformResponse(body) {
           if (body == null) {
-            throw new Error(
-              'The visualization cannot be made because the current subset is empty.'
+            throw new NoDataError(
+              'The visualization cannot be made because the current subset is empty.',
+              'No data',
+              204,
+              uuid()
             );
           }
-          console.log(body, decoder);
           return ioTransformer(decoder)(body);
         },
       })
--- a/packages/libs/eda/src/lib/core/api/DataClient/index.ts
+++ b/packages/libs/eda/src/lib/core/api/DataClient/index.ts
@@ -57,7 +57,15 @@ export default class DataClient extends FetchClientWithCredentials {
         method: 'POST',
         path: `/apps/${computationName}/visualizations/${visualizationName}`,
         body: params,
-        transformResponse: ioTransformer(decoder),
+        transformResponse(body) {
+          if (body == null) {
+            throw new Error(
+              'The visualization cannot be made because the current subset is empty.'
+            );
+          }
+          console.log(body, decoder);
+          return ioTransformer(decoder)(body);
+        },
       })
     );
   }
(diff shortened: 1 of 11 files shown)


packages/libs/eda/src/lib/core/api/DataClient/index.ts
```diff
@@ -57,7 +59,17 @@ export default class DataClient extends FetchClientWithCredentials {
         method: 'POST',
         path: `/apps/${computationName}/visualizations/${visualizationName}`,
         body: params,
-        transformResponse: ioTransformer(decoder),
+        transformResponse(body) {
+          if (body == null) {
+            throw new NoDataError(
+              'The visualization cannot be made because the current subset is empty.',
```
bobular, line 65: Probably the first of the suggestions below.
In the map we don't want to say "subset" if poss.
There's still an issue that the user might think "but I haven't added any filters" - some of these 204 no datas can be caused by little filters.
I think we'll have to punt that one to the next release. I'm wondering if the filter chips area in the header should show read-only chips for the viewport, timeslider, marker-config and selected markers?

```suggestion
              'The visualization cannot be made because no data remains after filtering.',
```

```suggestion
              'The visualization cannot be made because all data has been filtered out.',
```

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
```diff
@@ -547,17 +549,36 @@ export const GLOBAL_VIEWPORT = {
  */
 
 const noDataPatterns = [
-  /did not contain any data/, // comes from map-markers endpoint
   /Could not generate continuous variable metadata/, // comes from continuous-variable-metadata endpoint
 ];
 
 export function isNoDataError(error: unknown) {
+  if (error instanceof Error && error.name === 'NoDataError') return true;
```
bobular, line 556:
```suggestion
  if (error instanceof NoDataError) return true;
```

This works for me (with the appropriate import). Did you try this?

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
```diff
@@ -547,17 +549,36 @@ export const GLOBAL_VIEWPORT = {
...
   return noDataPatterns.some((pattern) => String(error).match(pattern));
 }
 
-export const noDataErrorMessage = (
+const noDataErrorMessage = (
```
bobular, line 560: Could possibly rename this `noDataLegendMessage`? And move it down next to the function that uses it.

bobular (approved):
This is great. Just some verbiage suggestions (Dave wanted the verbiage tweaking too) and one or two inconsequential code suggestions.

packages/libs/eda/src/lib/map/analysis/mapTypes/shared.tsx
jernestmyers: No, but I thought it would work. Instead, [I mocked a different custom error subclass we use](https://github.com/VEuPathDB/web-monorepo/blob/2512f4540573ecb57cfc00d9e14242a729db2f45/packages/libs/wdk-client/src/Service/ServiceError.ts#L13-L15).

Issue #832, opened 2024-02-07
Correlation: Simplify and polish prefilter inputs


asizemore:
Currently the correlation apps have the following "Prefilters"
<img width="1046" alt="Screen Shot 2024-02-07 at 12 09 15 PM" src="https://github.com/VEuPathDB/web-monorepo/assets/11710234/75cf6c11-8354-4a65-b2dd-0f119905f109">

These prefilters were added as a test to see if Dan though users would like them. After some thought we decided that we only should keep the first one, currently called "Proportion Non-Zero". So, we're ready to polish up the one remaining prefilter input.

For this ticket, please

- [ ] Remove the Variance and Standard Deviation prefilter inputs. The default values can still go with the request to the backend, but the user shouldn't see the inputs on the viz page at all.
- [ ] Change the user-facing name of "Proportion Non-Zero" to "Presence"
- [ ] Add a caution note that setting the value too high may cause the computation to fail. The mockups below show the proposed placement and font (gray, regular weight), but please let me know if this causes an issue or if a better idea arises for its placement!!
- [ ] The new Presence input should have a default value of 0.5. The user shouldn't have to interact with the Presence input at all in order to get the Generate Results button to activate. Currently there is a bug where the user has to interact with one of the prefilters for the default values to get added to the computation config. I assume this is a good time to tackle that as well! Lmk if that causes an issue!

<img width="1391" alt="Screen Shot 2024-02-07 at 12 14 14 PM" src="https://github.com/VEuPathDB/web-monorepo/assets/11710234/e9477741-6c04-4675-ac94-a629c98baeba">

PR #850, opened 2024-02-14
modifying correlation inputs

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/832.

For the last item to fix a bug or change behavior, honestly, I could not fully understand how the computation worked, but I managed to make it fix anyway 😅 

Some screenshots after changing:

832 mbio correlation01

832 mbio correlation02

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/Plugins.scss
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/Plugins.scss
@@ -42,6 +42,11 @@
       }
     }
   }
+
+  &-DescriptionContainer {
+    font-weight: normal !important;
+    text-transform: none;
+  }
 }
 
 .ConfigDescriptionContainer {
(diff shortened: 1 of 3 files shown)


moontrip:
> Looking great! One small request - can we make the text "Keep ... in at least" and "% of samples" in normal casing (somehow there has been a `capitalize` text-transform applied) and in regular font weight please? In other words, for both apps, everything after "Prevalence" should be regular, non-bold weight and regular capitalization without transform.
> 
> Otherwise it looks and works great! I am so happy to have that input working correctly 😄

Thank you for your review and suggestion @asizemore 👍 Yes I also noticed the capital characters at each word, which is probably a CSS. I will change it and make screenshots to confirm :)

moontrip:
Done @asizemore Please check below screenshots:

832 mbio correlation03

832 mbio correlation04

asizemore:
gorgeous!

asizemore (approved):
lgtm!!

Issue #838, opened 2024-02-10
WDK:  results table: column help: enable html


aurreco-uga:
(1) make attribute help an element in the rng file, ( for model backward compatibility just name it 'help' and client will choose element over attribute if exists)

(2) update the tooltip in the client table (dave said mesa knows how to do it)

example of column that will benefit from this:
https://plasmodb.org/plasmo/app/workspace/strategies/import/fa20c76032cb87cc
(any column where we show terms that need definition)

PR #851, opened 2024-02-14
Allow html in result table column tooltips

jernestmyers:
Depends on https://github.com/VEuPathDB/WDK/pull/80

This is the client work for #838


dmfalke (approved):
LGTM!

PR #852, opened 2024-02-15
Re-enable "Full" mode for lineplot y-axis

### LINKED ISSUE #833: Bug: Line plot y-axis range should always include zero
It looks like we are no longer always including 0 as the default range for the y-axis? was this change intentional? Can we please bring it back to the way it was (ie, default to go through zero)? also, it seems like the radio button option for “full” (which i believe would include zero) is disabled. 

I plotted Time since LLIN distribution (x-axis) vs Malaria incidence rate per 1000 people (y-axis), & overlayed for Cluster study arm for unsubsetted LLINEUP2 data.

Note that none of these variables have annotated values for defaultDisplayRangeMin, defaultDisplayRangeMax, or	defaultBinWidth in the conversion file.

It looks like this:
image

It should look like this by default (except, the radio button for "Full" y-axis should be selected, instead of having to play with a "custom" range):
image

bobular:
Fixes #833

See https://github.com/VEuPathDB/web-monorepo/issues/833#issuecomment-1946211463 for diagnosis and cure.

image
image

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
@@ -885,9 +885,26 @@ function LineplotViz(props: VisualizationProps<Options>) {
   );
 
   // use a hook to handle default dependent axis range for Lineplot Viz Proportion
+  // but first, effectively disable the distributionDefaults.rangeMin/Max to force
+  // the axis range calculations to use the observed min/max (because these are aggregate
+  // (e.g. mean/median) values whose range is much reduced compared to their min/max)
+
+  const rangelessYAxisVariable =
+    Variable.is(yAxisVariable) &&
+    (yAxisVariable.type === 'number' || yAxisVariable.type === 'integer')
+      ? {
+          ...yAxisVariable,
+          distributionDefaults: {
+            ...yAxisVariable.distributionDefaults,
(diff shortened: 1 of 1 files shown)


bobular:
What this does mean is that unannotated variables won't look great!  See this body temperature plot from GEMS1

image

moontrip:
@bobular @danicahelb @nkittur-uga 

So, I tracked down why we did set Auto-zoom as a default, and found the reason. It was requested by @danicahelb long time ago and I did work for it 😅 I wonder whether we should change back to Full or not as we had a reason to show Auto-zoom as a default

https://github.com/VEuPathDB/web-eda/issues/1433

bobular:
Perhaps a compromise is to make Auto-zoom default but don't disable the Full option?

bobular:
Implemented suggestions made here: https://epvb.slack.com/archives/CBE520QUC/p1708113915402589?thread_ts=1708008612.121099&cid=CBE520QUC 
except that the desired behaviour happens in "Full" y-axis mode.

moontrip:
@bobular I didn't look at the channel carefully (so many discussions) but if we ignore metadata's rangeMin/Max for Lineplot Viz, then it would work.

dmfalke (approved):
Clever! Hopefully we won't have to touch this stuff again for a while.

bobular:
Thanks both! @moontrip @dmfalke

Issue #853, opened 2024-02-15
EDA Histogram filter - axis control reset buttons out of place


dmfalke:
The reset buttons for the axis controls for the histogram filter are overlapping with the section label. It looks like the `margin-left` value of the button needs to change to `-1em`.

![image](https://github.com/VEuPathDB/web-monorepo/assets/365139/3c47beb9-c873-4ae5-8ea0-22cd8337eeeb)

PR #856, opened 2024-02-15
reset button position for histogram filter

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/853. As Dave said, marginLeft: -1em looks similar to the Live site's position


dmfalke (approved):
LGTM!

PR #858, opened 2024-02-15
SAM study details ux followup

### LINKED ISSUE #854: SAM study details UX feedback
The following changes are needed, before build 67:

- Reduce default size of study panel
- Update header verbiage:
  - "Showing: 440,305 samples from 5 studies"
  - "Selected: 1,234 samples from 2 studies"
- Use CTRL-CLICK
- Remove opacity from marker on hover

dmfalke:
closes #854

This PR addresses UX feedback:
- Reduce default size of study panel
- Update header verbiage:
  - "Showing: 440,305 samples from 5 studies"
  - "Selected: 1,234 samples from 2 studies"
- Use CTRL-CLICK
- Remove opacity from marker on hover

Note that in addition to removing opacity from markers on hover, I also made the bounding rectangle appear above the marker. This makes that rectangle visible when smaller than the marker, and easier to see in general. The rectangle is also non-interactive, so it does not impede the ability to click on the marker, nor does it interfere with the popup.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/mapTypes/types.ts
+++ b/packages/libs/eda/src/lib/map/analysis/mapTypes/types.ts
@@ -43,8 +43,6 @@ export interface MapTypeMapLayerProps {
   // and sent to plugin components that don't need it - we should also address this
   hideVizInputsAndControls: boolean;
   setHideVizInputsAndControls: (hide: boolean) => void;
-  // selectedMarkers and its state function
-  selectedMarkers?: string[];
   setSelectedMarkers?: React.Dispatch<React.SetStateAction<string[]>>;
   setStudyDetailsPanelConfig: (config: PanelConfig) => void;
   setTimeSliderConfig?: (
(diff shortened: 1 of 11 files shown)


packages/libs/eda/src/lib/map/analysis/mapTypes/types.ts
```diff
@@ -43,8 +43,6 @@ export interface MapTypeMapLayerProps {
   // and sent to plugin components that don't need it - we should also address this
   hideVizInputsAndControls: boolean;
   setHideVizInputsAndControls: (hide: boolean) => void;
-  // selectedMarkers and its state function
-  selectedMarkers?: string[];
```
bobular, line 47: good catch!

bobular (approved):
This all looks great, thanks. The rectangle z-index stuff looks nice too.

packages/libs/eda/src/lib/map/analysis/mapTypes/types.ts
dmfalke: It was the source of much confusion!

PR #860, opened 2024-02-15
Feature 771 expandy section apps

### LINKED ISSUE #771: Mbio: add expandable info section to apps
Our apps run some serious computations that are worth explaining a bit to the user. Also we've had user feedback asking for more details on the functions we use for the computations, etc. So, we'd like to add an expandable section, just like with the mosaic plots, that offer more description about the computation.

Each app would use a specific text that describes the computation, and i think the expandable section could go right below th ecomputation inputs. So for example in the next line below Data and Method, but before step 2 in the screenshot. Totally happy to make mockups if that would help!

Screen Shot 2024-01-16 at 11 52 24 AM

This ticket will require work from a few folks. To start:
- [ ] @dpbisme and @asizemore need to write app-specific text
- [ ] a kind frontend dev could help with the implementation? @dmfalke just tagging you here because i believe you already talked to dan about this. But once we're ready we can get anyone with time to implement!

asizemore:
Partially resolves #771 

Adds expandable section info to the two correlation apps.

https://github.com/VEuPathDB/web-monorepo/assets/11710234/95b7a972-42da-421a-9a54-78df7c148869

Not much else to say!

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
@@ -242,6 +242,19 @@ export function CorrelationAssayAssayConfiguration(
           </li>
         </ul>
       </p>
+      <br></br>
+      <H6>More Questions?</H6>
+      <p>
+        Check out the{' '}
+        <a href="https://github.com/VEuPathDB/microbiomeComputations/blob/master/R/method-correlation.R">
+          correlation function
+        </a>{' '}
+        in our{' '}
+        <a href="https://github.com/VEuPathDB/microbiomeComputations/tree/master">
+          microbiomeComputations
+        </a>{' '}
+        R package.
+      </p>
     </div>
   );
 
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
@@ -203,8 +203,8 @@ export function CorrelationAssayAssayConfiguration(
       <p>
         <ul>
           <li>
-            <strong>Taxonomic Level.</strong> The taxonomic abundance data to
-            used in the calcualtion
+            <strong>Taxonomic Level.</strong> The taxonomic abundance data to be
+            used in the calculation.
           </li>
           <li>
             <strong>Functional Data.</strong> The pathway, metabolic, or gene
@@ -212,7 +212,7 @@ export function CorrelationAssayAssayConfiguration(
           </li>
           <li>
             <strong>Method.</strong> The type of correlation to compute. The
-            Pearson method looks for a linear trend in the data, while the
+            Pearson method looks for linear trends in the data, while the
             Spearman method looks for a monotonic relationship. For Spearman and
             Pearson correlation, we use the rcorr function from the Hmisc
             package.
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
@@ -28,6 +28,7 @@ import SingleSelect from '@veupathdb/coreui/lib/components/inputs/SingleSelect';
 import { IsEnabledInPickerParams } from '../../visualizations/VisualizationTypes';
 import { entityTreeToArray } from '../../../utils/study-metadata';
 import { NumberInput } from '@veupathdb/components/lib/components/widgets/NumberAndDateInputs';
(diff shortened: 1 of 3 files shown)


dmfalke (approved):
LGTM!

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
```diff
@@ -182,6 +183,81 @@ export function CorrelationAssayAssayConfiguration(
...
+      <ol>
+        <li>Abundance of taxa at a given taxonomic level</li>
+        <li>Abundance of functional data (e.g. pathways, genes)</li>
+      </ol>
+      <br></br>
+      <H6>Inputs:</H6>
+      <p>
+        <ul>
+          <li>
+            <strong>Taxonomic Level.</strong> The taxonomic abundance data to
```
d-callan, line 206: to be used?

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
```diff
@@ -182,6 +183,81 @@ export function CorrelationAssayAssayConfiguration(
...
+            used in the calcualtion
+          </li>
+          <li>
+            <strong>Functional Data.</strong> The pathway, metabolic, or gene
```
d-callan, line 210: Maybe we can be more specific here? Pathway what? Is it accurate/ clear to say genes here?

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
```diff
@@ -182,6 +183,81 @@ export function CorrelationAssayAssayConfiguration(
...
+          <li>
+            <strong>Method.</strong> The type of correlation to compute. The
+            Pearson method looks for a linear trend in the data, while the
+            Spearman method looks for a monotonic relationship. For Spearman and
+            Pearson correlation, we use the rcorr function from the Hmisc
+            package.
+          </li>
+          <li>
+            <strong>Prevalence Prefilter.</strong> Remove variables that do not
+            have a set percentage of non-zero abundance across samples. Removing
```
d-callan, line 222: Is the input a percentage or decimal?

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayMetadata.tsx
```diff
@@ -161,81 +162,169 @@ export function CorrelationAssayMetadataConfiguration(
...
+      </p>
+      <p>Here we look for correlation between:</p>
+      <ol>
+        <li>
+          Abundance of taxa at a given taxonomic level or functional data such
+          as pathway abundance
+        </li>
+        <li>
+          Continuous metadata variables that are compatable, i.e. on a 1-1 path
+          from the taxonomimc assay entity.
```
d-callan, line 183: I'm not sure this is clear but I'm also not sure how to improve it

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayMetadata.tsx
```diff
@@ -161,81 +162,169 @@ export function CorrelationAssayMetadataConfiguration(
...
+      <H6>Inputs:</H6>
+      <p>
+        <ul>
+          <li>
+            <strong>Data.</strong> The abundance data to be correlated against
+            the study's metadata variables.
+          </li>
+          <li>
+            <strong>Method.</strong> The type of correlation to compute. The
+            Pearson method looks for a linear trends in the data, while the
```
d-callan, line 196: Remove 'a' before linear trends

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
asizemore: It's a percentage now. It gets transformed to a proportion before sending to the backend.

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayAssay.tsx
asizemore: Haha I was intentionally trying to leave it vague because we have multiple types of pathway data. I think gene is similarly a good term that can cover lots of data types. I don't want to have to update this list each time we get a slightly different data type.

packages/libs/eda/src/lib/core/components/computations/plugins/correlationAssayMetadata.tsx
asizemore: i'll give it a shot

PR #861, opened 2024-02-16
Ignore searches that are not defined in the model

dmfalke:
Addresses cases where not all searches are defined in the model, as is the case in VectorBase.


jernestmyers (approved):
LGTM!

dmfalke:
Thank you!

PR #865, opened 2024-02-16
Overhaul of timeslider verbiage

### LINKED ISSUE #722: Map - timeslider tooltip verbiage should mention that it is visualising data for the whole world not just the viewport
Probably just by adding `(anywhere in the world)` to the first bullet item?

bobular:
Fixes #722 
Screenshot from 2024-02-16 17-01-30

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
@@ -204,8 +204,8 @@ export default function TimeSliderQuickFilter({
           </li>
           <li>satisfies any filters you have applied</li>
           <li>
-            has data for the variable currently displayed on markers, and with
-            appropriate values for custom-configured markers
+            has values for the variable currently displayed on markers,
+            including custom-configurations if applicable
           </li>
           <li style={{ listStyleType: 'none' }}>
             (* data that has no geolocation will also be shown on the timeline)
--- a/packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
@@ -205,12 +205,10 @@ export default function TimeSliderQuickFilter({
           <li>satisfies any filters you have applied</li>
           <li>
             has values for the variable currently displayed on markers,
-            including custom-configurations if applicable
-          </li>
-          <li style={{ listStyleType: 'none' }}>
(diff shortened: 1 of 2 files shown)


bobular:
Just saw it says "has data that... has data for".  Will amend!

packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
```diff
@@ -196,12 +196,25 @@ export default function TimeSliderQuickFilter({
...
-          <li>Permanent filters are applied if applicable</li>
+          <li>is located anywhere on Earth*, not just currently in view</li>
           <li>
-            You currently have {filters?.length} active permanent filter(s)
+            has a value for <b>{variableMetadata?.variable.displayName}</b>
           </li>
+          <li>satisfies any filters you have applied</li>
+          <li>
+            has values for the variable currently displayed on markers,
+            including custom-configurations if applicable
```
dmfalke, line 208: I think we can lose "if applicable"

packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
```diff
@@ -196,12 +196,25 @@ export default function TimeSliderQuickFilter({
...
+          </li>
+          <li style={{ listStyleType: 'none' }}>
```
dmfalke, line 210: Ideally this would not be a list item, since it isn't semantically speaking.

dmfalke (approved):
Left some suggestions, but fine as-is

packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
bobular: For screen-reader purposes?

dmfalke: Yep, that's the main reason

bobular: I wanted it to be the same indentation level but not have to eyeball the indentation level. But putting it back to the <p> level looks fine, especially with parentheses.

dmfalke: I was about to say that I also don't like that it's indented with the rest of the list. To me, it gets lost. I expect "footnotes" to appear not indented.

PR #867, opened 2024-02-16
Make "### Studies" a link and remove button from tooltip

dmfalke:
This was suggested at a recent map ux meeting. I think it makes sense.

image

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
@@ -125,9 +125,6 @@ export function MapTypeHeaderStudyDetails(props: Props) {
           {studyEntityCount.data.count === 1
             ? studyEntity.displayName
             : studyEntity.displayNamePlural}{' '}
-          <button type="button" onClick={() => onShowStudies(true)}>
-            Show list
-          </button>
         </p>
       )}
     </div>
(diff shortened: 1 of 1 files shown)


packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
```diff
@@ -151,13 +147,20 @@ export function MapTypeHeaderStudyDetails(props: Props) {
           : outputEntity.displayNamePlural}
         {onShowStudies && studyEntity && (
           <>
-            {' '}
-            from {format(studyEntityCount.data.count)}{' '}
-            {studyEntityCount.data.count === 1
-              ? studyEntity.displayName
-              : studyEntity.displayNamePlural}
+            &nbsp;from&nbsp;
```
bobular, line 150:
```suggestion
            &nbsp;from
```
My OCD brain can't cope with the extra space before the link. Could we get rid of it please?

bobular (approved):
ALGTM!

packages/libs/eda/src/lib/map/analysis/mapTypes/MapTypeHeaderStudyDetails.tsx
dmfalke: Without that, there would be no space between "from" and "### Studies"

bobular: There was for me when I tried it

dmfalke: I just tested again. Strange!

dmfalke: I'm seeing consistent behavior in all major browsers (on linux) :shrug:

PR #868, opened 2024-02-16
Intercept nulls in another place and refactor

### LINKED ISSUE #866: Map - Unexpected backend type(s) encountered for continuous marker variable (donut/bar markers only)
Megastudy

Choose a continuous marker variable, e.g. **prevalence of Kdr L1014F**

Select an empty region of the timeslider

image

bobular:
Fixes #866 

This fixes one of the runtime exceptions when the back end `eda-service/filter-aware-metadata/continuous-variable` returns null and a 204.  However, it really makes a difference where you make the timeslider filter.

Here is where it causes a 204 and is now fixed in this PR.  In a comment below I'll demonstrate a 500 that probably needs fixing on the back end.

image

### DIFF
--- a/packages/libs/eda/src/lib/core/api/DataClient/index.ts
+++ b/packages/libs/eda/src/lib/core/api/DataClient/index.ts
@@ -59,17 +59,7 @@ export default class DataClient extends FetchClientWithCredentials {
         method: 'POST',
         path: `/apps/${computationName}/visualizations/${visualizationName}`,
         body: params,
-        transformResponse(body) {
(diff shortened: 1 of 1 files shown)


bobular:
So the following scenario causes a 500 on the back end. I think the subset is non-empty (there are no filters) but there's no data for the marker variable in that time range. The variable is **prevalence of wild-type Kdr L1014** in the megastudy.

image

jernestmyers (approved):
LGTM!

PR #878, opened 2024-02-19
remove parentheticals from Steps 2-3 in computes

asizemore:
We've have a run around with the correlation app names. We finally agreed on calling them both "Correlation" and using parentheticals to distinguish between the two. That seemed really nice on the new visualization page and on the thumbnail page but today i realized the actual viz page looks like this with parentheses all over the place. Yuck!

Screen Shot 2024-02-19 at 3 55 50 PM

This PR removes parentheses and the text they contain from the Generate Results button and the titles of Steps 2 and 3.

Screen Shot 2024-02-19 at 3 56 43 PM

Would really love to get this in for b67 if anyone is around!

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/RunComputeButton.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/RunComputeButton.tsx
@@ -27,7 +27,11 @@ export function RunComputeButton(props: Props) {
     >
       <FilledButton
         themeRole="primary"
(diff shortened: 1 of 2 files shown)


dmfalke (approved):
It might be nice to make a utility function called something like `removeParenthetical` or something like that.

asizemore:
Thanks @dmfalke ! I'll go ahead and merge but make a new PR for the improvement. That way there's no more rush but at least it will get done :)

PR #879, opened 2024-02-19
add remove or replace parentheticals utils

asizemore:
Discussed in #878 , we repeated ourselves _three_ times!!! It's function time!

This PR adds `removeParentheticals` and `replaceParentheticals` to our util functions. The former just strips out the parentheticals and surrounding whitespace from a string. The latter replaces that parenthetical+white space with a new string. This latter one isn't used yet but it seemed an obvious extension so I thought I might as well go ahead and add it.


dmfalke (approved):
LGTM!

PR #886, opened 2024-02-23
Remove jquery DataTable

### LINKED ISSUE #835: Replace jQuery-powered DataTable with Mesa
A request was made to add a feature to `DataTable` (selection counts) that already exists in `Mesa`. This led to evaluating what we gain by having both `DataTable` and `Mesa` versus consolidating to one configurable table component.

From prelim investigation and discussion, `DataTable` appears to include only one feature that Mesa does not: expandable rows. Since `DataTable` is used in only one instance, record tables, it may make more sense to add expandable rows to `Mesa` and to remove `DataTable`. 

Bonus: `DataTable` is driven by jquery, and we would like to remove all jquery (eventually) from our codebase.

jernestmyers:
Resolves #835

This is a pretty intense PR, so apologies in advance.

These model changes are trivial and would remove some client code and/or fix at least one issue I've noticed:
1. Declare `thumbnail` and `clustalInput` attributes as unsortable
2. The key for `pubmed link` needs the space removed for CSS reasons
3. Ortho's `accession` attribute isn't typed as a `link`like it should be

These model changes are less trivial but would eliminate some client hacks/guesswork:
1. Declaring sort types reliably
2. Declaring column types reliably

Other tasks/questions that come to mind are as follows:
1. Go up the `RecordTable` component's tree to remove unused props/code
2. Do we need to add truncation to Mesa? Currently the `truncateTo` property is being ignored.


dmfalke:
I've played with the code on a local dev site, and I think it looks and behaves great. The only comment I have, at the moment, from a UI perspective, is that I think the expanded child rows should have a border, and expanded rows should be in bold. I think it makes it easier to parse what's going on. Thoughts?

**Live site**

Screenshot 2024-02-29 at 13-37-36 Gene PF3D7_0409500


**This branch**

Screenshot 2024-02-29 at 13-37-46 Gene PF3D7_0409500

dmfalke (commented):
Some more observations:

- It looks like we can't have both selectable and expandable rows. I don't think we have a use case for both right now, but seems plausible.
- If the first row is expanded, and then the sorting is updated, then new first row will be expanded. We should either use something other than the index to track expanded rows, or we should clear expansion when sorting.
- It might be nice if `Mesa` provided its own error boundary when rendering child rows. That way, if a developer forgets to add one to the childRow component, a render error won't destroy the whole table.

jernestmyers:
Some updates below regarding the feedback up to this point.

> It looks like we can't have both selectable and expandable rows. I don't think we have a use case for both right now, but seems plausible.

I made some changes to allow both features in a single table. As of now, the expand/collapse toggle renders in the first column and the select column renders in the second column. If and when we have a use case for this, we can explore the UX further.

> If the first row is expanded, and then the sorting is updated, then new first row will be expanded. We should either use something other than the index to track expanded rows, or we should clear expansion when sorting.

I fixed this issue by passing in a `getRowId` method that uses the `RecordTable`'s existing `getSortIndex` function. I think the identifiers stored in the `expandedRows` array are unique and stable now.

> It might be nice if Mesa provided its own error boundary when rendering child rows. That way, if a developer forgets to add one to the childRow component, a render error won't destroy the whole table.

In a slack thread, we decided to just beef up the documentation of the optional `childRow` prop that tells users to wrap the components in `wdk-client`'s `ErrorBoundary`. If we add an `ErrorBoundary` in the `coreui` library, we won't have the Redux dispatching that the `wdk-client` version includes.

dmfalke (approved):
Changes look good!

PR #887, opened 2024-02-26
Missing bearer token support

dmfalke:
This PR adds missing bearer token support. It also removes an outdated authentication method for local dev in multi-blast.

### DIFF
--- a/packages/libs/user-datasets/src/lib/Service/UserDatasetUploadWrappers.ts
+++ b/packages/libs/user-datasets/src/lib/Service/UserDatasetUploadWrappers.ts
@@ -23,24 +23,40 @@ type UserDatasetUploadServiceWrappers = ReturnType<
 >;
 
 /*
- * The authentication method uses an 'Auth-Key' header, not the cookie like in WDK
+ * The authentication method uses a header, not a cookie like in WDK
  */
-function fetchWithCredentials(
+async function fetchWithCredentials(
   serviceUrl: string,
   path: string,
   method: string,
   body: any,
   contentType?: string
 ) {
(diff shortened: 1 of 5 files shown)


packages/libs/user-datasets/src/lib/Service/UserDatasetUploadWrappers.ts
```diff
@@ -23,24 +23,40 @@ type UserDatasetUploadServiceWrappers = ReturnType<
...
 ) {
-  const wdkCheckAuth =
-    document.cookie.split('; ').find((x) => x.startsWith('wdk_check_auth=')) ||
-    '';
-  const authKey = wdkCheckAuth.replace('wdk_check_auth=', '');
-  const authO = {
-    'Auth-Key': authKey,
-  };
-  const contentTypeO: Record<string, string> =
+  const cookies = Object.fromEntries(
```
ryanrdoherty, line 35: Seems like some duplicated logic here from the other PR, but I assume reuse is not convenient or you will address later.

ryanrdoherty (approved):

packages/libs/user-datasets/src/lib/Service/UserDatasetUploadWrappers.ts
dmfalke: It is duplicated, but not worth the effort to reuse `FetchClientWithCredentials`, since this package will be replaced by vdi (see https://github.com/VEuPathDB/web-monorepo/tree/feature/vdi-client-work/packages/libs/user-datasets). That branch will need an update, but different from this one.

PR #890, opened 2024-02-27
Refactor computation method options to include displayName property

jernestmyers:
Resolves https://github.com/VEuPathDB/web-eda/issues/1579


asizemore (approved):
Changes look and work great! Now that we have such nice ability to set the display name, i checked in with Dan B about some of the names to ensure we're using the right ones. He had the following to say for beta div:

> JSD should be 'Jensen–Shannon divergence'
Bray should probably be 'Bray Curtis dissimilarity'
Jaccard should be 'Jaccard index'

Do you mind making the above switches for the display names please?

I'll go ahead and approve though!

PR #891, opened 2024-02-27
Display fixed dummy input in correlation config UI for UX reasons

### LINKED ISSUE #843: Correlation assay v metadata: Add fixed input for metadata variables
There is a bit of confusion around what exactly is being correlated in the correlation assay v metadata app. To clarify, let's add a "fixed" input for the metadata. So it would look like:

Screen Shot 2024-02-12 at 1 00 52 PM

Just the same styling we do for the fixed inputs in other locations, for example the alpha div boxplot y-axis variable, but up in the computation configuration. This would be completely stylistic - no information about this "fixed" variable needs to be sent to the backend or added to a computation configuration object or anything like that.

jernestmyers:
Resolves #843 

Screenshot of the change:
image

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/Plugins.scss
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/Plugins.scss
@@ -35,6 +35,12 @@
     span {
       font-weight: 500;
     }
+    .FixedInput {
+      text-transform: none;
+    }
+    .FixedInputValue {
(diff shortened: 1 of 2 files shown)


asizemore (approved):
Looks and works great!

Do you mind removing the colon after (fixed) please? Today I learned that in all the other visualizations, there are no colons after the input name. Who knew!?? I was so surprised!

PR #889, opened 2024-02-26
Clear errorMessage when a new ontology summary is loaded

dmfalke:
Fixes #888 

This PR clears any error message related to fetching an ontology summary.

PR #892, opened 2024-02-27
Clear errorMessage when a new ontology summary is loaded (#889)

dmfalke:
(cherry picked from commit 2c8f4402caf91723fb145a9373a423fb2c94fd69)


dmfalke:
Branched from the wrong tag (should have been v1.0.123).

Issue #863, opened 2024-02-16
Map - rectangular marquee marker-selection with Ctrl-drag on map


bobular:
Ctrl (or Cmd for Mac) click-drag of a rectangle should select the markers within it. The rectangle does not need to remain when the mouse button has been released.

Shift-drag-rectangle should continue to do the regular Leaflet zoom.

It's probably low cost once we figure out how to do it!

dmfalke:
How do we decide if a marker is included in the selection? Some possibilities I can think of are
1. The _marker_ is _completely_ inside the rectangle
2. The _marker_ is _partially_ inside the rectangle
3. The _geohash_ is _completely_ inside the rectangle
4. The _geohash_ is _partially_ inside the rectangle
5. The _bounding box_ is _completely_ inside the rectangle
6. The _bounding box_ is _partially_ inside the rectangle

moontrip:
I think that geohash-based one is not a good option as it is mostly invisible unless one does mouse over on a marker. Currently I am thinking of the center of a marker as a basis.

PR #897, opened 2024-02-28
Area selection by ctrl key with mouse for SAM

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/863. A couple of notes are provided in the following and a video capture is attached in the end:

- The main area-select is taken from 'leaflet-area-select' library. I couldn't figure out why the lib did not work with our repo, even if  it just worked fine with simple react env (after running create-react-app). Thus, I modified the lib and directly loaded it instead of installing (i.e., yarn add). 

- The 'leaflet-area-select' is quite old library and does not have a typescript definition. Since it is not trivial task to type a leaflet-plugin, I just used `// @ts-nocheck` to ignore typescript (AreaSelect.tsx)

- After thinking over and trying several cases, the best place to use marker selection logic I found is each marker's MapType (e.g., DonutMarkerMapType, etc.) and a useEffect is utilized to update selectedMarkers. However, since MapType involves conditions (checking errors for marker response, etc.), thus some duplicate codes were utilized inside the useEffect, (markers & selectedMarkers).

- Overall, it seems to work fine: a video shot is attached in the next

https://github.com/VEuPathDB/web-monorepo/assets/12802305/d2fb3dc1-39c8-4671-bb5d-9d204507efad

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
@@ -9,7 +9,7 @@ import {
 import { getValueToGradientColorMapper } from '@veupathdb/components/lib/types/plots/addOns';
 import { TabbedDisplayProps } from '@veupathdb/coreui/lib/components/grids/TabbedDisplay';
 import { capitalize, sumBy, omit } from 'lodash';
-import { useCallback, useMemo } from 'react';
+import { useCallback, useMemo, useEffect } from 'react';
 import {
   useFindEntityAndVariable,
   Filter,
@@ -202,6 +202,8 @@ function BubbleMapLayer(props: MapTypeMapLayerProps) {
     },
     updateConfiguration,
     geoConfigs,
+    // pass coordinates of selected area
+    boxCoord,
   } = props;
 
   const configuration = props.configuration as BubbleMarkerConfiguration;
@@ -244,6 +246,42 @@ function BubbleMapLayer(props: MapTypeMapLayerProps) {
     },
     [handleSelectedMarkerSnackbars, props.configuration, updateConfiguration]
   );
+
+  // set useEffect for area selection to change selectedMarkers via setSelectedmarkers
+  // define useEffect here to avoid conditional call
+  // thus, this contains duplicate codes, e.g., markers and selectedMarkers
+  useEffect(() => {
+    if (!markersData.error && !markersData.isFetching) {
+      // convert marker data into markers
+      const markers = markersData.data?.markersData?.map((markerProps) => (
+        <BubbleMarker {...markerProps} />
+      ));
+      // define selectedMarkers
+      const selectedMarkers = markerConfigurations.find(
+        (markerConfiguration) =>
+          markerConfiguration.type === activeMarkerConfigurationType
+      )?.selectedMarkers;
+
+      // update selectedMarkers
+      if (boxCoord != null && markers != null) {
+        const boxCoordMarkers = markers
+          .map((marker) => {
+            // check if the center of a marker is within selected area
+            return marker.props.position.lat >= boxCoord.southWest.lat &&
+              marker.props.position.lat <= boxCoord.northEast.lat &&
+              marker.props.position.lng >= boxCoord.southWest.lng &&
+              marker.props.position.lng <= boxCoord.northEast.lng
+              ? marker.props.id
+              : '';
+          })
+          .filter((item) => item !== '');
+
+        // then, update selectedMarkers
+        setSelectedMarkers([...(selectedMarkers ?? []), ...boxCoordMarkers]);
+      }
+    }
+    // additional dependency may cause infinite loop
+  }, [boxCoord]);
 
   if (markersData.error && !markersData.isFetching)
     return getErrorOverlayComponent(markersData.error);
--- a/packages/libs/components/src/map/AreaSelect.tsx
+++ b/packages/libs/components/src/map/AreaSelect.tsx
@@ -1,0 +1,56 @@
+/*
+ * Select area by ctrl + mouse
+ */
+
+// several type errors because leaflet-area-select does not have type definition
+// so ignore it for now
+// @ts-nocheck
+import { useEffect } from 'react';
+import { useMap } from 'react-leaflet';
+import L from 'leaflet';
+// load modified leaflet-area-select directly
+import './AreaSelectHack.js';
+import { Bounds as BoundsProp } from './Types';
+
+interface AreaSelectProps {
+  setBoxCoord: (value: React.SetStateAction<BoundsProp | undefined>) => void;
+}
+
+export default function AreaSelect({ setBoxCoord }: AreaSelectProps) {
+  const map = useMap();
+
+  useEffect(() => {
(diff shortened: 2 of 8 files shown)


packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx
```diff
@@ -245,6 +247,42 @@ function BubbleMapLayer(props: MapTypeMapLayerProps) {
     [handleSelectedMarkerSnackbars, props.configuration, updateConfiguration]
   );
 
+  // set useEffect for area selection to change selectedMarkers via setSelectedmarkers
+  // define useEffect here to avoid conditional call
+  // thus, this contains duplicate codes, e.g., markers and selectedMarkers
+  useEffect(() => {
+    if (!markersData.error && !markersData.isFetching) {
+      // convert marker data into markers
+      const markers = markersData.data?.markersData?.map((markerProps) => (
```
bobular, line 256: Had a quick look (not tested yet!) and `markersData` seems missing from the dependencies and there's no need to convert `markersData` into `markers`. You can do the bounds check below with `markerData` instead of `markers` (so it's `data.position.lat` instead of `marker.props.position.lat`.

moontrip: @bobular Thank you for your feedbacks 👍 As for markersData-related one, yes it can be simplified so I made a new commit to address them. For the dependency, I checked that markersData-related one causes infinite loop so I didn't add it to the dependency array.

bobular:
Hi @moontrip - it looks like Mac command key is going to be a problem for us

https://github.com/w8r/leaflet-area-select/issues/15

moontrip:
> Hi @moontrip - it looks like Mac command key is going to be a problem for us
> 
> [w8r/leaflet-area-select#15](https://github.com/w8r/leaflet-area-select/issues/15)

@bobular Thank you for your information 👍 It is quite interesting. I didn't know that as a Windows/Ubuntu user 😅 I will try to find a way, probably next week when I am working for VEuPathDB.

bobular (changes requested):
This is a great start.

A few things are of concern though.

1. The installation issues of `leaflet-area-select` - I haven't investigated but combined with
2. The lack of Apple `Cmd` key support I think that
3. We should consider better maintained alternatives such as https://github.com/Leaflet/Leaflet.draw or https://www.npmjs.com/package/@geoman-io/leaflet-geoman-free 
4. Leaving out essential hook dependencies to avoid loops 
5. I think we can avoid tracking state of `boxCoord` at MapAnalysis.tsx and repetition of code within the map types and instead deploy a custom hook in each MapType that returns a handler for the selected bounds, and do something like this in DonutMarkerMapType.MapLayerComponent:

```
  const handleSelectedArea = useAreaSelectedMarkers({
    selectedMarkers,
    setSelectedMarkers,
    markerDataResponse,
  });

  return (
    <>
      {markerDataResponse.isFetching && <Spinner />}
      {markers && (
        <AreaSelect onSelectedArea={handleSelectedArea} />
        <SemanticMarkers
          markers={markers}
          // no changes to any props here
        />
      )}
    </>
  );
```

We probably want to check `markerDataResponse.pending` inside the hook and have the handler do nothing if the markers are pending.

bobular:
FYI: Stories are fixed with https://github.com/VEuPathDB/web-monorepo/pull/901 (merged)

bobular:
I think for now, all the story needs to do is call the callback which prints the bounds to the console. No need to actually do any marker-selection logic.

moontrip:
@bobular Thank you for your comments. I didn't have much time to work for VEuPathDB this week as I needed to do other project. I will check your feedbacks carefully, but I could manage to enable Mac's Command key for current area selection. Although I am using Windows/Ubuntu, I could test it works as Window key is essentially the same with Mac's Command key in the key mapping: `metaKey`.

bobular:
Hi @moontrip - I wasn't able to use the windows key on Ubuntu Linux unfortunately. Not without remapping keys in X11 anyway, which I'm not going to do at this point.

Assuming your fix is good, then let's continue with the "hacked (not) import" - I had a closer look and realise it's not interfering with any other Leaflet internals. No need to look at the other libraries (though we might need them one day for polygon/lasso support - though I think I remember you already started looking at that  some time ago?)

If you could take the logic out of MapAnalysis and make a hook/handler as suggested, that would be great.

I also wonder if ctrl-rectangle select should *add* to the marker selection?

moontrip:
Hi @bobular Thank you for your comments 👍 You can find my reply (if necessary) in the following starting from **DK:** with bold style.

> Hi @moontrip - I wasn't able to use the windows key on Ubuntu Linux unfortunately. Not without remapping keys in X11 anyway, which I'm not going to do at this point.

**DK:** Actually, I am also using Ubuntu in the VM. Unfortunately, Ubuntu assigned the windows key as a Super key for their purpose, so certainly it cannot be tested under Ubuntu. What I did for test was that I used `ngrok`, which, in short, delivers traffics from local services to public domain, to open SAM at Windows (my main OS) web browser. Then, I could test windows key in there.

> 
> Assuming your fix is good, then let's continue with the "hacked (not) import" - I had a closer look and realise it's not interfering with any other Leaflet internals. No need to look at the other libraries (though we might need them one day for polygon/lasso support - though I think I remember you already started looking at that some time ago?)
> 

**DK:** Sounds great. Yes IIRC I did look at react-leaflet-draw a couple of years ago. Not quite sure how much it was upgraded since then. 

> If you could take the logic out of MapAnalysis and make a hook/handler as suggested, that would be great.

**DK:** I will test it next Monday when I will work for VEuPathDB

> 
> I also wonder if ctrl-rectangle select should _add_ to the marker selection?

**DK:** I'm sorry that I cannot fully understand your question. Can you please elaborate it more?

bobular:
> > I also wonder if ctrl-rectangle select should _add_ to the marker selection?
> 
> **DK:** I'm sorry that I cannot fully understand your question. Can you please elaborate it more?

At the moment, the rectangle selection is replacing the current selected markers. I think it should add the markers in the rectangle to whatever previous selected markers existed (if any).

moontrip:
@bobular Ah I got your point 👍 I am not quite sure if it is doable due to a couple of reasons, e.g., mouse click event, etc., but I will check 

> > > I also wonder if ctrl-rectangle select should _add_ to the marker selection?
> > 
> > 
> > **DK:** I'm sorry that I cannot fully understand your question. Can you please elaborate it more?
> 
> At the moment, the rectangle selection is replacing the current selected markers. I think it should add the markers in the rectangle to whatever previous selected markers existed (if any).

moontrip:
@bobular couple of updates are made:

a) Area selection can add markers with pre-existing selected markers
b) Also changed change logic to avoid duplicate markers
c) Made a custom hook to simplify codes: typing and adjusting for different marker responses
d) tested to move area selection to AreaSelect component, but it did not work correctly. So, I will leave it as a future technical debt as it seems not to be trivial

bobular:
Hi @moontrip  Looks like these changes got merged when I merged the child PR. Nice.

moontrip:
> Hi @moontrip Looks like these changes got merged when I merged the child PR. Nice.

Yes because you branched out from this PR. Thank you for enhancing this work 👍 👍

Issue #899, opened 2024-02-29
Parent Issue: Remove all use of jquery


jernestmyers:
This issue provides a checklist of all instances of jquery in use in our client code. Items or an item can be made into an issue as we work toward excising all jquery code.

**Configs**
- [ ]  `packages/configs` has some template files related to jquery
- [ ] each package has a `config-overrides.js` file that specs jquery

**wdk-client**
- [ ] `Overlays/Tooltip`: possibly obviated by https://github.com/VEuPathDB/CoreUI/issues/75
- [ ] `Overlays/Popup`: used only in Dialog, but Dialog is used widely; two possible options as follows:
  - option 1: replace jquery code in Popup w/ vanilla JS and/or react refs
  - option 2: replace Dialog with CoreUI’s DraggablePanel and delete Popup (higher cost)
- [ ] `DataTable`: only used in `RecordTable`; [work is underway to delete this and use Mesa tables in the `RecordTable`](https://github.com/VEuPathDB/web-monorepo/pull/886)
- [ ] `TabbableContainer`: used in `Popup`, `AnswerFilterSelector`, `RecordFilter`, `OverThumbnails`; should be simple enough to replace jquery code w/ vanilla JS and/or react refs
- [ ] `AnswerFilterSelector`: used [in one trivial line](https://github.com/VEuPathDB/web-monorepo/blob/a69a18ee18044395544f40c7e4b50a65d18c0d57/packages/libs/wdk-client/src/Views/Answer/AnswerFilterSelector.jsx#L37)
- [ ] `HistogramAnalysis` and `AttributeFilter/Histogram`: these two components use the `jquery/flot` library to render histograms; can we not refactor to use our `Histogram` component in `packages/components`?

**genomics-site**
- [ ] `RemoteContent` and `VerticalScrollHelper`: neither seem to be in use and can be deleted
- [ ] `PathwayRecordClasses.PathwayRecordClass`: used in the `cytoscape` code but should be able to use JS instead
- [ ] `Gbrowse`: replace with vanilla JS and/or react refs

**ortho-site**
_Note: Not clear to me yet how any of these are used._
- [ ] `apycom-menu.js`
- [ ] `jquery.tooltip`
- [ ] `jquery.timers`

PR #903, opened 2024-03-01
Remove simple jquery instances

jernestmyers:
Works toward tasks in #899, specifically:

- TabbableContainer: used in Popup, AnswerFilterSelector, RecordFilter, OverThumbnails; should be simple enough to replace jquery code w/ vanilla JS and/or react refs
- AnswerFilterSelector: used [in one trivial line](https://github.com/VEuPathDB/web-monorepo/blob/a69a18ee18044395544f40c7e4b50a65d18c0d57/packages/libs/wdk-client/src/Views/Answer/AnswerFilterSelector.jsx#L37)
- RemoteContent and VerticalScrollHelper: neither seem to be in use and can be deleted


dmfalke (approved):
LGTM!

PR #904, opened 2024-03-01
OrthoMCL - Custom tree-table for OrthoGroup page

bobular:
Testing:

There are only a few tree files available on the back end, so only a few orthogroups work.  Here's are the valid OGs at the moment:

* OG6_102443
* OG6_103280
* OG6_105395
* OG6_105810
* OG6_126264
* OG6_126317
* OG6_126334
* OG6_126431

Run `ortho-site` locally and access the path `/app/record/group/OG6_102443`
Back end config for `.env` is

```
BASE_PROXY_URL=https://qa.orthomcl.org/
LEGACY_WEB_APP_URL=${BASE_PROXY_URL}/orthomcl
PROJECT_ID=OrthoMCL
```

```
yarn && yarn nx start @veupathdb/ortho-site
```

To do:

- [x] restore clustal functionality
- [x] pfam domain architecture in main table
- [x] pfam legend
- [x] pfam legend checkboxes apply to main table
- [x] search box
- [x] take care of width issues/horizontal scroll/remove some columns/sticky checkbox column? (partially solved by using the horizontal scroll of the whole page)
- [ ] sortable columns? (hide tree if sorted) (wont do?)
- [x] maybe do core/peripheral filters?
- [ ] also add EC "legend filter"? - some questions around partial ECs
- [x] do general WDK page/record reorganisation
- [x] fix inline click with tooltips? (currently clicking changes the row height - perhaps just hide the tree if any rows are expanded?)
- [x] reset button? (this has never been requested by users/domain experts)
- [x] full_id needs display name and go on left
- [x] un-collapse the main section

From the https://github.com/VEuPathDB/web-monorepo/pull/1140 branch 
- [ ] Remove "Phyletic distribution" section
- [ ] Move "Protein" section to top
- [x] Add ability to highlight table rows from species filter (not done directly, but row highlighting via checkboxes is now a thing)

UX suggestions

- [x] organism/taxon legend(s)? (filter done)
- [ ] zoom out to see whole tree? or maybe a modal?
- [x] tree bolder (thicker lines)
- [ ] add explanation of tree
- [ ] tree branch metrics? (future)
- [x] pfam cartoons should scale to protein length
- [x] move filters to the left as per https://epvb.slack.com/archives/CHA5FRKQF/p1721315477658799?thread_ts=1721310024.223819&cid=CHA5FRKQF
- [ ] (SB) Move protein length column earlier - the columns might go:  domain, accession, length, description, Organism 
- [ ] DEAL WITH BIG GROUPS (speed and usability) - collapsable nodes - some by default?


packages/libs/components/src/components/tidytree/TreeTable.tsx
```diff
@@ -74,18 +91,24 @@ export default function TreeTable<RowType>(props: TreeTableProps<RowType>) {
...
           styles={globalStyle`
 	  .DataTable {
 	    margin-bottom: 0px !important;
+            width: 80%;
 	  }
 	`}
         />
         <Mesa state={tableState} />
-      </>
+      </div>
```
dmfalke, line 111: Sorry, I'm just noticing this. I'm not sure we should use `Global` here, as the style will be reflected on all subsequent pages.

Instead, could you use a `css` prop on the parent `div` component (line 96), and target `.DataTable` in there?

```tsx
      <div
        css={{
          flexGrow: 1,
          width: 1 /* arbitrary non-zero width seems necessary for flex */,
          '.DataTable': {
            marginBottom: '0px !important',
            width: '80%'
          }
        }}
      >
        <Mesa state={tableState} />
    </div>

```

packages/sites/ortho-site/webapp/wdkCustomization/js/client/utils/tree.ts
dmfalke, line 1: Could we change these identifiers such that they include the "group" prefix? E.g., `GroupTreeResponse`, `groupTreeResponseDecoder`, etc?

dmfalke (commented):
This is looking really good. I made some comments below. Let me know what you think!

packages/libs/components/src/components/tidytree/TreeTable.tsx
bobular: Oh yeah, that's much simpler (and no side effects) - done in https://github.com/VEuPathDB/web-monorepo/pull/904/commits/5c896a0152eacb19703484cce39ce670734fc8c8

packages/sites/ortho-site/webapp/wdkCustomization/js/client/utils/tree.ts
bobular: Sure! https://github.com/VEuPathDB/web-monorepo/pull/904/commits/fd446f87e06b32b937a17ca637acf2ff4fd1180f

dmfalke:
I figured out why the section is always collapsed by default:
https://github.com/VEuPathDB/web-monorepo/blob/26807b89d8f8a131984a63f67da68a70e83309f2/packages/sites/ortho-site/webapp/wdkCustomization/js/client/store-modules/RecordStoreModule.ts#L19-L29

I don't know why this was done, but I suppose we can remove this entire case.

bobular:
Brilliant - thanks - yes that will help a lot! @dmfalke

bobular:
There's nothing left after removing that `case`.

Should we track down where the `reduce` is called, or just eviscerate this function and make it no op?

dmfalke:
> There's nothing left after removing that `case`.
> 
> Should we track down where the `reduce` is called, or just eviscerate this function and make it no op?

I think we can just remove the `reduce` function.


Not sure why I didn't see this comment sooner... sorry for the delay!

dmfalke:
It's imported here:

https://github.com/VEuPathDB/web-monorepo/blob/26807b89d8f8a131984a63f67da68a70e83309f2/packages/sites/ortho-site/webapp/wdkCustomization/js/client/wrapStoreModules.tsx

I'm pretty sure you can just delete the reduce and don't need to make any other changes.

bobular:
Ah, thanks, I was missing something fundamental that I get now.

bobular:
It looks like we can't avoid the "We're sorry, something went wrong." modal being rendered if the wdkService back end call returns a 404. That hook (useWdkService/useOrthoService) doesn't handle errors the same way as the EDA's `usePromise`.

However, we can avoid asking for a tree when there are too few sequences, thus avoiding some user-facing errors. Will push a commit soon to address this and make sure the regular table still renders.

```
  const numSequences = props.value.length;
  const treeResponse = useOrthoService(
    numSequences > 2
      ? (orthoService) => orthoService.getGroupTree(groupName)
      : () => Promise.resolve(undefined),
    [groupName, numSequences]
  );
```

dmfalke:
@bobular what is the state of the backend for this? Do I need to use the `diamond` branch and the rm database John mentioned?

packages/libs/coreui/src/components/Mesa/Ui/TableSearch.jsx
```diff
@@ -18,7 +18,7 @@ class TableSearch extends React.PureComponent {
   clearSearchQuery() {
     const query = null;
```
dmfalke, line 19: This can be deleted

packages/sites/ortho-site/webapp/wdkCustomization/js/client/services.tsx
```diff
@@ -26,6 +31,12 @@ const orthoServiceWrappers = {
       method: 'get',
       path: `/group/${groupName}/layout`,
     }),
+  getGroupTree: (wdkService: WdkService) => (groupName: string) =>
+    wdkService.sendRequest(groupTreeResponseDecoder, {
+      useCache: true,
```
dmfalke, line 36: It's worth noting that this property causes the response to be stored in indexeddb, which is governed by the browsers quota policy for domains. If the database exceeds the quota, attempts to cache responses will begin to fail (don't worry, this is handled for a seamless user experience). In all likelihood, important/critical responses will already be cached, so it should not impact website load times.

In the future, we might consider using LRU (least recently used) caches for things like this, where the cache container has an identifier, and we retain up to a certain number of entries, ejecting those used least recently.

For now, this is probably fine.

packages/sites/ortho-site/package.json
```diff
@@ -89,5 +89,8 @@
   "files": [
     "dist",
     "webapp"
-  ]
+  ],
+  "dependencies": {
+    "patristic": "^0.6.0"
```
dmfalke, line 94: I think this should be declared in packages/libs/components/package.json.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/types/patristic.d.ts
dmfalke, line 1: I think this should be moved to packages/libs/components

dmfalke (commented):
Overall, this looks great. I added some comments. Let me know what you think!

dmfalke:
Another thing... it would be nice for the row counts to indicate when the table is filtered, as in this screenshot:

image

packages/sites/ortho-site/package.json
bobular: There's
```
import { parseNewick } from 'patristic';
```
in Sequences.tsx though.

bobular: We don't use patristic directly in `components`. If `tidytree` needs it then that's its business!

packages/sites/ortho-site/webapp/wdkCustomization/js/client/services.tsx
bobular: Interesting. Are the cache entries keyed by the full request payload (like react-query)?

If so, this would seem to be a general WDK problem, not specific to fetching the tree (is it massively bigger than the responses from the other endpoints?)

packages/sites/ortho-site/package.json
dmfalke: Aha, thanks!

packages/sites/ortho-site/webapp/wdkCustomization/js/client/services.tsx
dmfalke: Entries are keyed by path and, if provided, cacheId. See [here](https://github.com/VEuPathDB/web-monorepo/blob/40eacc150830a81f32626d3e33893baa76c9fdd6/packages/libs/wdk-client/src/Service/ServiceBase.ts#L171). I don't know how big these newick strings can get, so it's hard to say how much of an issue it would be.

Again, I think this is probably fine, but I wanted to mention the current limitations.

bobular:
Hi all.  I've made some expensive synchronous operations into asynchronous effects and added a switch to show only core proteins when #seqs > 2000

OG7_0000009 with 10k sequences is a good test case.

The remaining performance issues seem to be 
a) filtering the tree based on search (which is currently synchronous) - takes a couple of seconds for a 10k group.
b) rendering

My FF still complains sometimes for the full (core+peripheral) and this seems to be a mix of Pfam architecture rendering and tree rendering. Only disabling both seems to prevent any slow-down warnings. We probably need to think a bit more about large groups and what sensible cutoffs are.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -448,3 +456,21 @@ function createSafeSearchRegExp(input: string): RegExp {
     return new RegExp(escapedInput, 'i');
   }
 }
+
+function summarizeIDMismatch(A: string[], B: string[]) {
```
dmfalke, line 492: Maybe name this something like `logIdMismatchSummary`

bobular:
Latest commit adds this. No dashed outline when there are PFam domains.
image

bobular:
Thanks @rdemko2332 - I'll get this wired up ASAP :smiley:

bobular:
Seems to work fine after merging `main` (including the React upgrade)

dmfalke:
I am changing this to a draft, until we are ready to merge into main. @bobular feel free to change back :)

bobular (commented):
Just one possible oversight - see the comment.

Works great - so much better than before!

packages/libs/wdk-client/src/Components/SearchBox/RealTimeSearchBox.tsx
```diff
@@ -43,138 +42,76 @@ type Props = {
...
-            onKeyDown={this.handleKeyDown}
-            placeholder={placeholderText}
-            value={searchTerm}
-          />
-          <i className={`fa fa-${iconName} ${searchIconClassName}`} />
-          <button
-            className={cancelBtnClassName}
-            style={
-              cancelBtnRightMargin ? { right: cancelBtnRightMargin } : undefined
+  }, [delayMs, searchTerm, onSearchTermChange]);
```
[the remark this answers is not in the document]
bobular: > Not using `delayMs`?

We should fix and test this before merging, @dmfalke

dmfalke:
I am planning to merge this PR. We can make a new issue for any remaining tasks we intend to implement.

PR #908, opened 2024-03-05
Persist UI state for map's draggable panels

### LINKED ISSUE #275: Map - save locations of draggable elements (floater and legend) in user analysis

jernestmyers:
Resolves #275 

**Notes on legend implementation:**
- I persisted legend state per marker type primarily to accommodate the bubble's two legends

**Notes on viz panel implementation**:
- I opted to store the viz panel's position and dimension state per marker type, not per visualization. My first plan was to store a panel state for each viz in the event we ever allow more than one viz panel to be open, but I ran into a syncing issue that didn't seem trivial. Thus, it made sense to me to implement a lower cost solution rather than pursue a UX we may not ever need nor want.
- I've done some work to including @dmfalke's `isVisible` property that he used for the substudy menu, though it's not actually getting used at this point; it's unclear to me if it's needed.


packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/DonutMarkerMapType.tsx
```diff
@@ -453,9 +485,75 @@ function MapOverlayComponent(props: MapTypeMapLayerProps) {
...
+  const updateLegendPosition = useDebouncedCallback(
+    (position: PanelConfig['position']) => {
+      setLegendPanelConfig({
+        ...legendPanelConfig,
+        pie: {
+          variable: position,
+        },
+      });
+    },
+    250
```
bobular, line 497: Is the debouncing really needed? I found that the callback really is only called on drag completion. And we don't debounce the viz position and size below.

packages/libs/eda/src/lib/map/analysis/appState.ts
```diff
@@ -342,5 +385,7 @@ export function useAppState(
     setViewport: useSetter('viewport'),
     setTimeSliderConfig: useSetter('timeSliderConfig', true),
     setStudyDetailsPanelConfig: useSetter('studyDetailsPanelConfig'),
+    setLegendPanelConfig: useSetter('legendPanelConfig'),
+    setVisualizationPanelConfigs: useSetter('visualizationPanelConfigs'),
```
bobular, line 389: I'm hoping we don't need to add these marker-type-specific setters.
We could imagine a marker type mode that doesn't have a legend, for example. Or no visualization panels. (I am channelling Dave here... :rofl: )

I'm pretty sure we can make the updates to the legend and viz panel configs using the regular `updateConfiguration` inside the relevant XXXMarkerMapTypes. (e.g. in the `updateLegendPosition`, `updateVisualizationPosition` and `updateVisualizationDimensions` callbacks)

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/DonutMarkerMapType.tsx
```diff
@@ -453,9 +485,75 @@ function MapOverlayComponent(props: MapTypeMapLayerProps) {
...
+    setActiveVisualizationId(undefined);
+    setVisualizationPanelConfigs({
+      ...visualizationPanelConfigs,
+      pie: {
+        position:
+          visualizationPanelConfigs?.pie?.position ??
+          DEFAULT_DRAGGABLE_VIZ_POSITION,
+        dimensions:
+          visualizationPanelConfigs?.pie?.dimensions ??
+          DEFAULT_DRAGGABLE_VIZ_DIMENSIONS,
```
bobular, line 544: Can't we just spread the old values?

```suggestion
        ...visualizationPanelConfigs?.pie,
```

bobular (commented):
This is working great, after I debugged an issue not at all related to your work (see https://github.com/VEuPathDB/web-monorepo/issues/913) but happens quite a lot after moving panels around (because of the PATCH updates)

I've made quite a large refactoring suggestion, sorry...

Happy to talk it through on Slack (the review comments might not come in a logical order)

bobular (commented):
Looks great. Thanks for the various tidy ups (e.g. `props.configuration as XYZMarkerConfiguration` --> `configuration`)

I'm pretty sure `isVisible` isn't used for the draggable visualization. The on/off control is 100% due to the defined/undefined nature of `activeVisualizationId`. Do we want to address that as a followup tech debt ticket, or strip it out here?

I had one UX problem with the floating visualization if I moved it bottom-left. When creating a second, new visualization, it appears in the "saved" position bottom-left. That's not necessarily a bad thing, but it's going to be hidden under the menu and the user might think nothing has happened.

I'm not sure how easy it is to save the panel state in the actual visualisation config. Probably not easy, because those configs are "pure EDA".

The options I think we have are
a) always open a brand-new visualization in the middle of the screen
b) always close/hide the left hand panel when opening a brand-new visualization

I think I prefer b) - though some people may object. On not-so-huge screens, the new yet-to-be-configured visualization is slightly underneath the left hand panel and the user needs to hide it or move the viz anyway.

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
```diff
@@ -38,6 +40,8 @@ export interface BarPlotMarkerConfiguration
   binningMethod: BinningMethod;
   selectedValues: SelectedValues;
   selectedCountsOption: SelectedCountsOption;
+  legendPanelConfig: PanelPositionConfig;
+  visualizationPanelConfig: PanelConfig;
```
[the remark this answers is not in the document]
dmfalke: Can you elaborate? I'm not seeing any duplication.

dmfalke:
> I've done some work to including @dmfalke's `isVisible` property that he used for the substudy menu, though it's not actually getting used at this point; it's unclear to me if it's needed.

Maybe `isVisible` should not be part of `PanelConfig`. We would have to add a new `isStudyDetailsPanelVisible` prop, but that's not too bad, imho.

packages/libs/eda/src/lib/map/analysis/MarkerConfiguration/BarPlotMarkerConfigurationMenu.tsx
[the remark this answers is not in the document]
dmfalke: I understand, now. FWIW, I plan on removing these types from `appState` in my collections branch: https://github.com/VEuPathDB/web-monorepo/blob/5fb0f76e97313abc04c8ebd4cf35d00a49bdab30/packages/libs/eda/src/lib/map/analysis/appState.ts#L41-L52. Those type defs don't really belong there, in our "MapType" world.

jernestmyers:
> > I've done some work to including @dmfalke's `isVisible` property that he used for the substudy menu, though it's not actually getting used at this point; it's unclear to me if it's needed.
> 
> Maybe `isVisible` should not be part of `PanelConfig`. We would have to add a new `isStudyDetailsPanelVisible` prop, but that's not too bad, imho.

I vote for punting this work to a future ticket.

jernestmyers:
> The options I think we have are
a) always open a brand-new visualization in the middle of the screen
b) always close/hide the left hand panel when opening a brand-new visualization
I think I prefer b) - though some people may object. On not-so-huge screens, the new yet-to-be-configured visualization is slightly underneath the left hand panel and the user needs to hide it or move the viz anyway.

I implemented option b. Dave and I both prefer this option as well.

bobular (approved):
This is great.  Added https://github.com/VEuPathDB/web-monorepo/issues/935 as follow-up (killing two birds with one stone, hopefully)

PR #932, opened 2024-03-11
75 coreui tooltip component

chowington:


chowington:
Almost ready for undrafting. The only thing I can't yet figure out is why the MesaTooltips in the Mesa header cells won't show. Any thoughts?

packages/libs/wdk-client/src/Components/Overlays/Tooltip.tsx
```diff
@@ -23,151 +16,58 @@ let defaultOptions = {
...
-export const EagerlyLoadedTooltip = wrappable(Tooltip);
-
-const withJquery = lazy<Props>(async () => {
-  // @ts-ignore
-  await import('!!script-loader!../../../vendored/jquery.qtip.min.js');
-});
+const WDKClientTooltip = ({
+  content,
+  children,
+  classes = 'qtip-wdk',
```
jernestmyers, line 52: Looks like we need to update the CSS applied to this class in `wdk-Common.scss`
image

packages/libs/coreui/src/components/Mesa/Ui/HeadingCell.jsx
```diff
@@ -186,14 +186,14 @@ class HeadingCell extends React.PureComponent {
 
     if (!column.helpText && !column.htmlHelp) return null;
     return (
-      <Tooltip
+      <MesaTooltip
```
jernestmyers, line 189: I swapped out this instance of `MesaTooltip` with `HelpTrigger` as shown below, and things seem to work. May need more testing, though.

```js
<HelpTrigger>
  {column.htmlHelp ?? column.helpText}
</HelpTrigger>
```

packages/libs/wdk-client/src/Components/Overlays/Tooltip.tsx
jernestmyers: After some digging, I found that if I pass in `classes = ''` to `WDKClientTooltip` in this instance- thereby bypassing the default `qtip-wdk` class assignment- then the icon renders as expected _**but**_ the tooltip doesn't render. I'm not able to figure out a quick solution to this problem.

dmfalke: I would even go so far to say that maybe we don't need this component at all. Thoughts?

dmfalke (commented):
I've started to review this. See my preliminary comments about `WDKClientTooltip` below

packages/libs/wdk-client/src/Components/Overlays/Tooltip.tsx
chowington: @dmfalke I agree

packages/libs/wdk-client/src/Components/AttributeFilter/Histogram.jsx
```diff
@@ -24,7 +24,7 @@ import {
 } from '../../Components/AttributeFilter/AttributeFilterUtils';
 import Icon from '../../Components/Icon/IconAlt';
 import { CollapsibleSection, IconAlt } from '../../Components';
-import Tooltip from '../../Components/Overlays/Tooltip';
+import Tooltip from '@veupathdb/coreui';
```
jernestmyers, line 27: This errors out because it's not a default import, so it needs to be wrapped in `{ }`

packages/libs/wdk-client/src/Views/Answer/AnswerFilter.jsx
```diff
@@ -2,7 +2,7 @@ import { debounce } from 'lodash';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { HelpTrigger } from '@veupathdb/coreui/lib/components/Mesa';
-import Tooltip from '../../Components/Overlays/Tooltip';
+import Tooltip from '@veupathdb/coreui';
```
jernestmyers, line 5: Same issue as `Histogram`, this needs to be wrapped in `{ }`

packages/libs/wdk-client/src/Components/AttributeFilter/FieldList.jsx
```diff
@@ -187,7 +187,7 @@ function FieldNode({ node, isActive, searchTerm, handleFieldSelect }) {
   }, [isActive, nodeRef.current, searchTerm]);
 
   return (
-    <Tooltip content={node.field.description} hideDelay={0}>
+    <Tooltip title={node.field.description} leaveDelay={0}>
```
jernestmyers, line 190: I forgot that this particular tooltip will render a small, empty yellow box when the title is null-ish. I wonder if we should open a follow-up ticket for the new `Tooltip` component to render null if its title prop is falsy?

packages/libs/user-datasets/src/lib/Components/UserDatasetStatus.tsx
```diff
@@ -56,7 +53,7 @@ export default function UserDatasetStatus(props: Props) {
     : 'exclamation-circle';
   const children = <Icon className="StatusIcon" fa={faIcon} />;
   const visibleContent = props.useTooltip ? (
-    <Tooltip content={content}>{children}</Tooltip>
+    <Tooltip title={content}>{children}</Tooltip>
```
jernestmyers, line 56: This tooltip is still not firing off in the `Status` column of the `My Data Sets` table. I think this component just needs refactoring. I think if we render the Tooltip above the Link component, it'll work just fine. If you don't mind, could you open an issue for this and I'll do the follow up work in the VDI client branch after merging this work into it? It'll create a big enough merge conflict in that component such that it makes just as much sense for me to handle it!

jernestmyers (changes requested):
Just a few more notes, some of which can be follow-up work. This is a big enough overhaul that we may also want to tip off some stakeholders to be on the lookout for any broken tooltips, similar [to my warning re: the jquery DataTable retirement](https://epvb.slack.com/archives/CPFT1FMFS/p1710337819434349).

packages/libs/wdk-client/src/Components/AttributeFilter/FieldList.jsx
chowington: Done, let me know what you think of the solution!

packages/libs/coreui/src/components/info/Tooltip.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  return !(
+    title === '' ||
+    _.isEqual(title, {}) ||
+    _.isEqual(title, []) ||
+    typeof title === 'boolean'
+  ) ? (
+    <MUITooltip {...props} />
+  ) : null;
+};
```
jernestmyers, line 19: So in [the checkbox tree of this UI](https://plasmodb.org/a/app/record/gene/PF3D7_0104400#SNPsAlignment), I don't think this logic is quite working yet. The title is being passed into the tooltip as `undefined`, which evaluates as true in the ternary, and results in an empty yellow box. When I tweak the `FieldNode` prop to render an empty string as a fallback to undefined, then evaluating to false in the ternary results in none of the content being rendered. BUT, I noticed in my testing that if the MUITooltip itself receives an empty string, it will render the desired content: the content itself with no tooltip. As a result, here's what I did that seems to achieve the desired result:

```js
const UnstyledTooltip = (props: ComponentProps<typeof MUITooltip>) => {
  const { title, ...remainingProps } = props;

  const finalProps = !(
    title === '' ||
    // I added this null-ish check to account for null or undefined
    title == null ||
    _.isEqual(title, {}) ||
    _.isEqual(title, []) ||
    typeof title === 'boolean'
  ) ? props : { title: '', ...remainingProps}
  
   return <MUITooltip {...finalProps} />
};
```

jernestmyers (approved):
All other testing looks good to me, so I'm going to go ahead and approve so that after you address the falsy title issue you can go ahead and merge without further delay!

packages/libs/user-datasets/src/lib/Components/UserDatasetStatus.tsx
chowington: Here's the new ticket! https://github.com/VEuPathDB/EdaNewIssues/issues/776

Issue #921, opened 2024-03-08
RecordTable: follow up work from jquery DataTable -> Mesa conversion


jernestmyers:
This issue is to track follow up work that results from https://github.com/VEuPathDB/web-monorepo/pull/886

**Model changes** (some of which can remove/simplify client code):
- [ ] Declare thumbnail and clustalInput attributes as unsortable
- [ ] The key for pubmed link needs the space removed for CSS reasons
- [ ] Ortho's accession attribute isn't typed as a linklike it should be

**UI**
- [ ] Add row counts
- [ ] Add selected rows count

PR #933, opened 2024-03-12
Data table followup work

jernestmyers:
Works to resolve issues found in https://github.com/VEuPathDB/web-monorepo/issues/921

Specifically, this branch does the following:
- fix Bob's discovery re: "eventHandlers undefined" issue in his orthotree branch
- fix ortho's accession column issues w/ links (previously thought this would be handled in the model by adding `type="link"` somewhere, but is client-side instead)
- fix issue with the `wrappedChildRow` logic breaking w/ classes only in production bundle (removed faulty function check)
- pass props to Mesa to enable row counts
- define column width for childRow toggle columns


dmfalke (approved):
LGTM!

Issue #199, opened 2023-05-10
Update layers icon (and bring back the button)


asizemore:
Discussed in styling 05/10/23

We like the two layers best! The simplest option is to use MUI's [Layer](https://mui.com/material-ui/material-icons/?query=layer&selected=Layers). The _best_ option is to use a rounded version of this because we've made a styling decision that we'll have rounded corners. If MUI doesn't have a rounded corners version, it's pretty trivial to make. Happy to do that if it's helpful.

We also need to actually show a layers icon/selector! It got lost along the way, somehow.

### Update for 2024-03

Need to use a default layer that shows the placenames in English - not the local language as we do at present.

Don't sweat too much about the icon. The Leaflet default is OK I think? Though rounded corners may be the dealbreaker...?

PR #934, opened 2024-03-13
adding layers for SAM

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/199. Screenshots are:

map layer

map layer01


bobular (approved):
This is great, and we got the English names back for free!

PR #937, opened 2024-03-13
Refactor area select to avoid two `useEffect`s

bobular:
This is branched out from https://github.com/VEuPathDB/web-monorepo/pull/897.

Addressing a few issues:

1. moving all logic out of `MapAnalysis`
2. improving the `useEffect` in `AreaSelect`
    - adding functional cleanup
    - including all dependencies
    - remove 0.25 padding (this wasn't previously active because the `setValidate()` was being cancelled straight after it was being set)
    - removing a rogue `this._map`
3. changing the hook so it returns a callback that updates the selectedMarkers
4. a bit of renaming and removing unneeded props

- [ ] add rectangle advice to snackbar


packages/libs/components/src/map/AreaSelect.tsx
```diff
@@ -10,47 +10,43 @@ import { useMap } from 'react-leaflet';
...
-    const bounds = map.getBounds().pad(-0.25);
-    // check restricted area on start and move
+    const bounds = map.getBounds(); //.pad(-0.25); // this was limiting area selection to only the center of the map!
     map.selectArea.setValidate((layerPoint) => {
-      return bounds.contains(this._map.layerPointToLatLng(layerPoint));
+      return bounds.contains(map.layerPointToLatLng(layerPoint));
     });
 
-    // switch it off
-    map.selectArea.setValidate();
```
bobular, line 52: This was effectively a no-op!

moontrip:
@bobular I have tested and reviewed your work and I realized what I did wrong in my previous attempts 😅 Anyway, mostly it worked fine but I found two issues. 

a) this does not work for bubble marker and I know what the reason is. It is because unlike other two marker types (Donut and Bar/Chart) its `markerProps` is not `markerDataResponse.markerProps`, but `markersData.data.markersData.markerProps`. This was the reason why I passed both `markerDataResponse` and `markersData.data.markersData` in my custom hook, `useAreaSelect`: https://github.com/VEuPathDB/web-monorepo/blob/15c0e0a65e381ba96fcfa9f64ad9c8427ab9ec01/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BubbleMarkerMapType.tsx#L252 So, you would probably need to do similar work for your `useOnAreaSelected` hook's arguments.

b) I noticed that there was intermittent issue for using area selection by Ctrl+Click, e.g., multiple area selections are made quickly etc.: Ctrl key was often cancelled that led to typical panning, not area selection. Just out of curiosity, I did test with my branch (https://github.com/VEuPathDB/web-monorepo/tree/863-area-select) but at least this issue did not happen. Thus, I do not have an answer for this issue.

bobular:
Thanks for the careful testing @moontrip :smiley: 

I removed the "don't select markers while `isFetching`" behaviour because
a) it simplifies our code and makes the donut/bar/bubble handling easier
b) we already allow ctrl-click selection while the spinner is still spinning and I think actually this is a good thing, so let's allow rectangle selection too.

I also got the occasional pan instead of rectangle (and very occasionally also a zoom behaviour) when trying to do things quickly. I also get a small blue square on the viewport (permanent - doesn't pan or zoom with the map) when this happens.

I'm pretty sure this happens when a rerender of `<AreaSelect>` is caused by its callback prop `onAreaSelected` changing. This changes when the dependencies within `useOnAreaSelected` change, of course.

I've just pushed https://github.com/VEuPathDB/web-monorepo/pull/937/commits/5e48f9ad6770337aca1f0d7cd5b0ce410ea04a32 to see if that helps but I doubt it will.

Also I notice a bug where if you ctrl-click select **three** markers quickly, the second one deselects when the third selects. I think this is something to do with queued state updates or something like that. Or maybe to do with "selected marker pruning" effects in `SemanticMarkers`. We can investigate further...

bobular:
Actually that might have helped a bit?

The select-three-markers-quickly bug is still there though.

moontrip:
> Actually that might have helped a bit?
> 
> The select-three-markers-quickly bug is still there though. (this is now a [new ticket](https://github.com/VEuPathDB/web-monorepo/issues/939) - it's nothing to do with this PR)

Hi @bobular Sorry that I couldn't check this yesterday as I was sick. Anyway, I checked the aforementioned inconsistency more carefully, and what I found is that the presence of snackbar hook (`handleSelectedMarkerSnackbars()`) inside the `setSelectedMarker()` (e.g., for `DonutMarkerMapType`, https://github.com/VEuPathDB/web-monorepo/blob/5e48f9ad6770337aca1f0d7cd5b0ce410ea04a32/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/DonutMarkerMapType.tsx#L342) seemed to cause the issue. If commenting it out, then it works okay like my PR. Thus, I guess that there exist some small delays when handling both selectedMarkers and snackbar, especially `setSelectedMarker()` is passed into `AreaSelect` component as a callback function. Moving that snackbar hook into the bottom of `updateConfiguration()` works better, but it is not always though. Hope this would be helpful.

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/useOnAreaSelected.ts
```diff
@@ -6,7 +6,12 @@ import { Bounds, MarkerProps } from '@veupathdb/components/lib/map/Types';
 interface EssentialMarkerDataResponseProps {
   error: unknown;
   isFetching: boolean;
-  markerProps?: MarkerProps[]; // base type is all we need here
+  // donut and bar
+  markerProps?: MarkerProps[];
+  // bubble is different!
+  data?: {
+    markersData?: MarkerProps[];
+  };
```
dmfalke: Can someone give me a recap of what's going on here? I find it very suspicious that we would need to do something like this. Why don't we just pass `markersData` in the same way from all places? Forgive me for not reading the code more carefully, this just stuck out to me.

dmfalke (commented):
I would suggest that `AreaSelect` be rendered in `SemanticMarkers`, rather then as a sibling. I think doing so would obviate the need for `useOnAreaSelect`, since that logic can be implemented in `SemanticMarkers`, which already has a handle on everything it needs. Thoughts @bobular @moontrip ?

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/useOnAreaSelected.ts
moontrip: @dmfalke my comment, a) at https://github.com/VEuPathDB/web-monorepo/pull/937#issuecomment-1996506016 would be helpful. :)

bobular:
Hi @moontrip - feel free to merge this to your PR if it helps.

@dmfalke - ah, yes, that looks like it would work. It would be conditionally rendered if `setSelectedMarkers` was provided. Thanks!

packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/useOnAreaSelected.ts
bobular: > Can someone give me a recap of what's going on here? I find it very suspicious that we would need to do something like this. Why don't we just pass `markersData` in the same way from all places? Forgive me for not reading the code more carefully, this just stuck out to me.

I think you were looking at an old version?

bobular:
Hi @moontrip - I've implemented Dave's suggestion and it all looks good.

Please could you review?  I think it's good enough for the public. Very very occasionally the ctrl-drag does a pan on the map (presumably because the `AreaSelect` component is in an intermediate state between renders?) but I can't find a way to reproduce it reliably.

I am tempted to re-target this to `main` and then merge in the latest. Or we can merge this to `863-area-select` and update that branch. What do you think?

moontrip:
> Hi @moontrip - I've implemented Dave's suggestion and it all looks good.
> 
> Please could you review? I think it's good enough for the public. Very very occasionally the ctrl-drag does a pan on the map (presumably because the `AreaSelect` component is in an intermediate state between renders?) but I can't find a way to reproduce it reliably.
> 
> I am tempted to re-target this to `main` and then merge in the latest. Or we can merge this to `863-area-select` and update that branch. What do you think?

@bobular I knew that you would be tempted to do this! 😄 👍 Anyway I will take a review this morning and I think that we can use this PR as a primary one to be merged into the main branch. We can just leave a link to point out my original PR (863-area-select) and close it: I will do that after review.

moontrip:
@bobular It looks good and such an inconsistent behavior observed previously is less frequent I think. Perhaps this simplification was helpful? Anyway, as I mentioned this PR can be merged into main instead of my previous PR. If you do that, then I will close my PR after this is merged into main. 👍

moontrip (approved):

bobular:
> @bobular It looks good and such an inconsistent behavior observed previously is less frequent I think. Perhaps this simplification was helpful? Anyway, as I mentioned this PR can be merged into main instead of my previous PR. If you do that, then I will close my PR after this is merged into main. 👍

Thank you DK - I will re-target, make updates from main, test and merge.

PR #941, opened 2024-03-15
Limit site search titles to one line and add title attribute

### LINKED ISSUE #919: site search result: when hit title is too long  (play with some limit) truncate and offer content in popup, as in gene page:prod descr
omar specifies in https://epvb.slack.com/archives/CPFT1FMFS/p1709825335527829?thread_ts=1709767969.978859&cid=CPFT1FMFS
just the top (hit title?) as in gene page.

Screen Shot 2024-03-06 at 2 52 03 PM

jernestmyers:
Resolves #919 

image

### DIFF
--- a/packages/libs/web-common/src/components/SiteSearch/SiteSearch.scss
+++ b/packages/libs/web-common/src/components/SiteSearch/SiteSearch.scss
@@ -454,6 +454,9 @@
   &--Result {
     &Link {
       font-size: 1.145em;
+      white-space: nowrap;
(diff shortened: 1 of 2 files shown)


jernestmyers:
> Not sure, but it could be useful to use a React tooltip component so that users can copy the full text. Barring that, I think this looks great.

Yeah, I had some concerns about performance but also figured the information is still attainable in the `Fields matched` section. 🤷

Issue #800, opened 2024-01-24
genomics: better message when accessing the record page of an old ID


aurreco-uga:
example:
https://toxodb.org/toxo/app/record/gene/TGME49_293370

when a gene maps into more than one gene we shows Page not found.
We should suggest that they try the site search (or better? run it for them :)

could be useful for 
https://redmine.apidb.org/issues/53447

dmfalke:
fwiw, i prefer not running it for them.

jernestmyers:
Here's a mockup of the UI to kickstart UX discussions:
![image](https://github.com/VEuPathDB/web-monorepo/assets/69446567/5a1061f5-6e29-4d11-83d7-824a61e8acfa)

aurreco-uga:
@jernestmyers we need to remove the auth-tkt key from the search link...  maybe it is not an issue in prod?  not sure

aurreco-uga:
This resurfaces when dealing with ortho7 and how old legacy genomic sites will be showing links to OG6 groups..
we decided  that in most cases there is a mapping 1:1 to OG7 and we want to send them to the OG7 page. (where maybe we will add link sto the rtho  legacy from OG6 groups).
Under these circumstances the current message is not good and steve dave and I updated it to: https://github.com/VEuPathDB/web-monorepo/commit/13ceaa2c3c58339b25ecd0570847c26b03c147b9

aurreco-uga:
This is the final verbiage committed in commit above

![Image](https://github.com/user-attachments/assets/9516b2a9-ebac-4487-bee3-b27841c9aa5b)

aurreco-uga:
implemented

<img width="1154" height="619" alt="Image" src="https://github.com/user-attachments/assets/8a02ac14-8b70-4471-8790-713909d5f9d3" />

PR #944, opened 2024-03-18
Improve UI for no record found

jernestmyers:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/800


aurreco-uga (approved):

aurreco-uga:
jeremy i iknow i just approved but i think it would be even better to add a link to run the site search with the non found ID?

jernestmyers:
> jeremy i iknow i just approved but i think it would be even better to add a link to run the site search with the non found ID?

I recall @dmfalke not liking that idea. Perhaps it needs a UX discussion first?

aurreco-uga:
just in case i wasnt clear:  im not suggesting to run the search but to link to it within the verbiage.. 
"Please use our site search feature to... for **ID_xxxxxx**"
(the ID in bold would be a link to get to a site search result  /app/search?q=ID_xxxxxxx/)

PR #945, opened 2024-03-18
Refactor Dialog and Popup to remove jquery

jernestmyers:
Works to resolve the second `wdk-client` task from https://github.com/VEuPathDB/web-monorepo/issues/899

### DIFF
--- a/packages/libs/wdk-client/src/Components/Overlays/Dialog.css
+++ b/packages/libs/wdk-client/src/Components/Overlays/Dialog.css
@@ -51,6 +51,9 @@ body.wdk-ModalOpen {
   border-top-right-radius: 4px;
   background: #eee;
   padding: 0.6em 0.8em;
+}
+
+.wdk-DialogHeader.draggable {
   cursor: move;
 }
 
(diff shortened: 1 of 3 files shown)


dmfalke (commented):
I think this looks great. However, I think we need to make the component draggable by default. There are some places where we definitely want the dialog to be draggable, even though we don't specify it. For example, the step details dialog in the strategy panel.

If you have the time and energy, you could find all of the places where we use the Dialog component and we can determine if each one should be draggable. Otherwise, let's just make it the default behavior.

jernestmyers:
> I think this looks great. However, I think we need to make the component draggable by default. There are some places where we definitely want the dialog to be draggable, even though we don't specify it. For example, the step details dialog in the strategy panel.
> 
> If you have the time and energy, you could find all of the places where we use the Dialog component and we can determine if each one should be draggable. Otherwise, let's just make it the default behavior.

I opted for the simpler solution of defaulting to true. That's essentially how it was working before, though now I suppose we'll needlessly be passing `draggable=true` to some components.

Issue #809, opened 2024-01-27
gene page, tables with checkbox selection: add dynamic selected count


aurreco-uga:
DR requests for the gene page ortholog table where we select rows for running an MSA
https://redmine.apidb.org/issues/53037
I see this as a fgeneric feature for any table that allows selection, as in favorites
<img width="1165" alt="Screen Shot 2024-01-27 at 10 00 06 AM" src="https://github.com/VEuPathDB/web-monorepo/assets/16354313/f0126a2b-f949-48b2-a13c-fc10ff670917">

PR #947, opened 2024-03-19
Refactor orthologs form to show selection counts

jernestmyers:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/809

As far as I know, the orthologs table in the gene record page is the only record table with selectable rows. Wiring this up for Mesa's selection counts required some refactoring. I don't love the slight jump in the table when the first selection is made or when selections are cleared, but I ran out of ideas on how to fix it (short of doing some reworking in the Mesa table itself).

Also note that a model update to `clustalInput` is in order to avoid the onerous condition that filters out the column. When this branch is ready to merge, I'll update the model and remove that logic, but it's useful in reviewing this PR.


packages/libs/coreui/src/components/Mesa/Ui/SelectionCell.jsx
```diff
@@ -43,9 +47,29 @@ class SelectionCell extends React.PureComponent {
...
-      <th className="SelectionCell" onClick={handler}>
+      <th
+        className={
+          'SelectionCell' + (selectColumnHeadingDetails ? '__withDetails' : '')
+        }
+        onClick={handler}
+      >
         {inert ? null : (
-          <Checkbox checked={checked} indeterminate={isIndeterminate} />
+          <AnchoredTooltip
```
jernestmyers, line 57: I'm not sure we really want/need this tooltip, but I added it to hedge against complaints re: losing the "Check All" and "Clear All" buttons. To be clear, this would be rendering a tooltip in all Mesa instances with row selections.

dmfalke (commented):
Code looks good, but I haven't tested yet.

I'm finding myself wishing we could use Mesa row selection in an "uncontrolled" mode, where we don't have to retain selection state, inputs have a configurable name attribute, and the various event handlers are not required. Then, we could render the table inside a form element and form submission would just work.

dmfalke (approved):

PR #948, opened 2024-03-20
617 snackbar for zooming and panning with selected markers

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/pull/937. 

A video capture is attached in the following:

Uploading Snackbar for zooming and panning.mp4…

### DIFF
--- a/packages/libs/components/src/map/MapVEuMap.tsx
+++ b/packages/libs/components/src/map/MapVEuMap.tsx
@@ -441,7 +441,7 @@ function CustomZoomControl(props: CustomZoomControlProps) {
         <div className="leaflet-control-zoom leaflet-bar leaflet-control">
           <a
             className={'leaflet-control-zoom-in' + disableMaxZoomButton}
-            href="#"
+            href="/#"
             title="Zoom in"
             role="button"
             aria-label="Zoom in"
@@ -451,7 +451,7 @@ function CustomZoomControl(props: CustomZoomControlProps) {
           </a>
           <a
             className={'leaflet-control-zoom-out' + disableMinZoomButton}
-            href="#"
+            href="/#"
             title="Zoom out"
             role="button"
             aria-label="Zoom out"
@@ -461,7 +461,7 @@ function CustomZoomControl(props: CustomZoomControlProps) {
           </a>
           <a
             className="leaflet-control-zoom-out"
-            href="#"
+            href="/#"
             title="zoom to data"
             role="button"
             aria-label="zoom to data"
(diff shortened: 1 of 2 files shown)


packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -441,7 +441,7 @@ function CustomZoomControl(props: CustomZoomControlProps) {
         <div className="leaflet-control-zoom leaflet-bar leaflet-control">
           <a
             className={'leaflet-control-zoom-in' + disableMaxZoomButton}
-            href="#"
+            href="/#"
```
bobular, line 444: Why was this needed?

packages/libs/components/src/map/SemanticMarkers.tsx
```diff
@@ -48,11 +49,33 @@ export default function SemanticMarkers({
...
+  const [isZoom, setIsZoom] = useState(false);
+  const [isDragend, setIsDragend] = useState(false);
+
   useMapEvents({
+    // cancel marker selection with a single click on the map
     click: () => {
       if (setSelectedMarkers != null) setSelectedMarkers(undefined);
     },
+    // check zooming: zoomend somehow does not work consistently here
+    // but zoomend just worked fine at MapVEuMap
```
bobular, line 64: I think we have a more general issue with trying to handle the same event in different places, see #952 

If we can't reliably handle the same events in different components then I suppose we need a different approach here. One that doesn't rely on Leaflet events.

How do we distinguish between no selected marker IDs are present in `consolidatedMarkers` because
1. we panned them all off the screen
2. we changed the zoom level
?

One approach would be to count the number of characters in `selectedMarkers[0]` and compare it with the number of characters in the first `consolidatedMarkers`'s ID.  The number of characters is equivalent to the zoom level.

packages/libs/components/src/map/MapVEuMap.tsx
moontrip: @bobular It is simply because eslint complained it 😓

packages/libs/components/src/map/SemanticMarkers.tsx
moontrip: @bobular Those events (zoom & dragend) are consistent for me, perhaps because they were not used at MapVEuMap's map events. Anyway, your suggestion would also work if we will only distinguish those two cases (zooming and panning) 👍

moontrip: @bobular Yes, Sir! 😄 Usually I am working for VEuPathDB on Monday and Tuesday so I didn't have a plan to do that this week anyway 😉

Issue #920, opened 2024-03-08
visx stories are broken


asizemore:
Opened storybook and noticed the Network and Volcano stories are all broken with the same error:

<img width="1095" alt="Screen Shot 2024-03-08 at 7 22 05 AM" src="https://github.com/VEuPathDB/web-monorepo/assets/11710234/a7cb836b-ca81-4737-80b5-0f43ac6deddd">

I tried the simple solution of adding `React` but no dice :(

If anyone wants a (hopefully) quick bug to fix, please go for it!

PR #953, opened 2024-03-22
change order of babel presets

dmfalke:
This PR fixes an issue with storybook stories in `@veupathdb/components`.

Several map stories, and all network stories, have been failing with a `React is not defined` error. I believe it is related to adding the `@emotion/babel-preset-css-prop` to the end of the list of babel presets. The `@babel/preset-react` automatically adds `import React from 'react'` to modules. If `@emotion/babel-preset-css-prop` introduces a reference to `React` to a module, this would explain the error, since `@babel/preset-react` has already run. 

Moving it to the beginning fixes the issue.

### DIFF
--- a/packages/libs/components/.storybook/main.js
+++ b/packages/libs/components/.storybook/main.js
@@ -16,7 +16,7 @@ module.exports = {
   babel: async (options) => {
     return {
       ...options,
-      presets: [...options.presets, '@emotion/babel-preset-css-prop'],
+      presets: ['@emotion/babel-preset-css-prop', ...options.presets],
       // See https://stackoverflow.com/questions/70406632/typescript-parameter-properties-not-working-with-storybook-rollup-in-developm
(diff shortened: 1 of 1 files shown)


moontrip (approved):
This works out! So, https://github.com/VEuPathDB/web-monorepo/issues/920 can also be resolved owing to this. I made a reference to the ticket so that this can close the ticket. 👍

asizemore:
Perhaps something for us to discuss later - i found [storyshots](https://github.com/storybookjs/storybook/tree/master/addons/storyshots/storyshots-core) which lets you add automatic testing to see if each story loads. Requires jest. Seems like less work than setting up lots of intricate jest testing and at least gains us some heads up if a new things breaks a story.

bobular (approved):
Amazing detective work and thanks for the clear explanation :smiley:

PR #957, opened 2024-03-22
Bipartite network click behavior

### LINKED ISSUE #923: Bipartite Network: Add first clicking abilities to node
The goal is to add a few click abilities to the node.
1. Linking out. If the node is an eigengene (gene list representative), we want to get from the node to the wdk gene list results page. If the node is an enzyme we want to get to kegg. Etc. 
2. Focus. The node we clicked on gets a thicker border and the label gets bold. This node's edges remain the same color but all other edges are colored Gray300.

The backend team will provide us with a base url via a [collections attribute](https://github.com/VEuPathDB/ApiCommonData/blob/master/Load/ontology/General/collections/collections.yaml), and we can use some part of the node id or display name (details on their way) to complete the link. 

The frontend has a few pieces to complete. Feel free to take each as a separate issue or do a few or redo the breakdown - whatever makes most sense

- [ ] For the `NodeWithLabel` component, create a little node menu that appears on click (ready to work on)
- [ ] When the node is clicked, gray out all edges except those coming from the clicked node. Add a gray "Unfocused" item in the Link Color legend. Neighbor nodes (nodes linked to the clicked node) get emphasized. (ready to work on)
- [ ] Add a "Link Out" button to this little node menu
- [ ] Incorporate the actual link into the button

I'll update this ticket as we get more information, but we could at least do the first two checkboxes for now!

Mockups live in Mbio Correlation app -> Bipartite Network 2.0. Some screenshots below! Please ignore adding to a scatterplot. That will be a separate issue.

Screen Shot 2024-03-08 at 1 34 41 PM
Screen Shot 2024-03-08 at 1 35 02 PM
Screen Shot 2024-03-08 at 1 35 16 PM

### LINKED ISSUE #951: Bipartite Network: Alphabetically order nodes
Right now the node order looks all over the place. They're not ordered by degree or alphabetically. They're probably ordered by our variableId, but that's not so helpful :) 

We'd like to order the nodes A -> Z in the columns by the variable's display name. This will improve the usability of the network by making it closer to a table so that users can find what they're looking for more quickly.

dmfalke:
closes #923
closes #951 

TODOs
---

- [x] Add click handling
- [x] Implement selectedNodes logic
- [x] Implement activeNode logic
- [x] Style "NodeActions" menu

Notes
---
I opted to kick the proverbial DRY can down the road:
- Rules about node selection (at most one per partition) are implemented directly in `BipartiteNetwork`, rather then in the visualization.
- `NodeActions` are implemented directly in `BipartiteNetwork`, rather then generalized in `NodeWithLabel`.

It's not entirely clear how this should all be generalized and abstracted. For example, we might want our own `Graph` component that encapsulates the logic and styling related to node click actions. If we know it will be useful in the foreseeable future, then I can abstract this more.


dmfalke:
Here is a screenshot of the latest:

image

Some questions I have:
- Is the styling of the menu too subtle? Eg, should the background be something other than white?
- Should the edges of selected nodes be highlighted?

dmfalke:
Here is a screenshot of the latest:

image

Some questions I have:
- Is the styling of the menu too subtle? Eg, should the background be something other than white?
- Should the edges of selected nodes be highlighted?

asizemore:
Hello! This is looking awesome! 
I've only done a quick look so far but can be back later today to go more in depth. At least I can respond to your questions!

> For example, we might want our own Graph component that encapsulates the logic and styling related to node click actions. If we know it will be useful in the foreseeable future, then I can abstract this more.

I can say a bit about the nearish term future (2024). I expect there will be one more network viz that is a much prettier version of what is shown in the Plots/Network stories. We'll end up with some clicking something-or-other on the nodes, likely at least a linking out function like we have here in the bipartite network. That's expected during the summer probably. So if you're having fun, feel free to abstract, if you need to move on, don't worry about it :)


> Is the styling of the menu too subtle?

Probably a little too subtle, but I like where it's going. I think this is worth a post in the styling group channel


> Should the edges of selected nodes be highlighted?

Yes please! We think the best way to do this is actually by graying out any edges that is not coming out of a selected node. Something like 

Screen Shot 2024-03-25 at 6 47 28 AM



More soon! But hope this at least answers some questions!

dmfalke:
Very helpful, thanks!

Regarding the click menu, I was thinking I should use the same styling as `SingleSelect`:

image

It's going to take some refactoring, to be able to use the "SingleSelectMenu" (naming is hard), without the button. I'm hoping that styling will be adequate in the context of the network.

asizemore:
Nice that's a good idea. Can menu items have icons in the SingleSelectMenu?

dmfalke:
> Can menu items have icons in the SingleSelectMenu?

Yes: https://github.com/VEuPathDB/web-monorepo/blob/a856efcc07dc1d006eee0fe10cd19e232b807fab/packages/libs/coreui/src/components/inputs/SingleSelect.tsx#L9

dmfalke:
This branch is now in shape for a demo. The code is a bit of a mess, and I will clean things up if this is approved.

Particularly, I would like to:

- [ ] Move the actions dropdown to be a part of `NodeWithLabel`
- [ ] Render the "gene list" option as a link, rather than a button
- [ ] Clean up styling/css


Here is a screencast showcasing the latest:

[Screencast from 2024-03-29 08-27-36.webm](https://github.com/VEuPathDB/web-monorepo/assets/365139/c8d520a8-83ae-4b0d-acc6-a28b08d822e5)

dmfalke:
Another screenshot showcasing the changes on plasmo (note the target site (qa.hostdb) at the end of the video is down).

[Screencast from 2024-03-29 08-34-30.webm](https://github.com/VEuPathDB/web-monorepo/assets/365139/5e903c28-2558-42d0-9d08-e401a0456ee4)

aurreco-uga:
some comments they might be premature.. ignore if so.. 
1- would be nice to have some indication that you should mouseover the node to get some options..  
2- the popup could say at top "_(click to select)_" ?
3- selecting the node: i would have expected to stay when clicking a second node.. (if this is for scatterplot), maybe we plan to enable ctrl-click

dmfalke:
> Particularly, I would like to:
> 
> * [ ] Move the actions dropdown to be a part of NodeWithLabel
> * [ ] Render the "gene list" option as a link, rather than a button
> * [ ] DRY up node actions for wgcna datasets
> * [ ] Clean up styling/css

Given that we are still in "design" mode for interactivity, I am not going to spend time cleaning things up. I don't think it's worth the effort, since much of this might be replaced by a different UI.

dmfalke:
I added a banner about interactivity:

image

I'm open to suggestions.

packages/libs/components/src/plots/BipartiteNetwork.tsx
```diff
@@ -26,6 +33,12 @@ export interface BipartiteNetworkSVGStyles {
   columnPadding?: number; // space between the left of the svg and the left column, also the right of the svg and the right column.
 }
 
+export interface NodeAction {
```
asizemore, line 36: I wonder if this would be better in `Network.tsx`? Or we can leave that for when we consolidate the components, too!

packages/libs/components/src/plots/BipartiteNetwork.tsx
```diff
@@ -47,6 +60,8 @@ export interface BipartiteNetworkProps {
   labelTruncationLength?: number;
   /** Additional error messaging to show when the network is empty */
   emptyNetworkContent?: ReactNode;
+  /** Entries for the actions that appear in the menu when you click a node */
+  getNodeActions?: (nodeId: string) => NodeAction[];
```
asizemore, line 64: This is totally a nit but could we change the name slightly to make it more clear that these are the action options in the menu? Before I got here, i expected this to have how nodes get clicked and hovered and all that jazz as "actions". Maybe "getNodeMenuActions"? Just a suggestion!

packages/libs/components/src/plots/BipartiteNetwork.tsx
```diff
@@ -112,65 +131,165 @@ function BipartiteNetwork(
...
+
   return (
     <>
       <div
         className={containerClass}
         style={{ width: '100%', ...containerStyles, position: 'relative' }}
       >
-        <div ref={plotRef} style={{ width: '100%', height: '100%' }}>
+        <div style={{ display: 'inline-flex' }}>
+          <Banner
```
asizemore, line 238: You beat me to the Banner haha! Two thoughts:
1. I think the banner might look nicer at the viz level so that it can float above the legends. We can add a little info icon, too. Alternatively i think centering it in the plot could help as well.
2. The dropdown is only available for eigengene nodes in plasmo/host. The other nodes won't have a dropdown for now. Maybe this should be handled in my PR, but would be easier to control the banner verbiage from the viz i'd think

asizemore: Oh wait i see there is `getNodeActions &&` on line 245... that should do it i'd think, but i'm still seeing the additional text in mbio. I need to look at this more

asizemore (approved):
In general this looks really good. I think ignore my comments about the banner changing based on `getNodeActions` because in my PR that'll be all fixed. I just originally didn't see where everything was coming from.

One thing i want to play with is if we can get the highlighted node's edges to render on top of all the others. Not crucial for this PR since it's adding so much value already, but something we should figure out

asizemore:
Just to follow up on my comment about the edges, it's possible it may be as "simple" as sorting the links so that the ones coming from the highlighted nodes are at the bottom. One could add the sort to the construction of `linksWithCoordinates` and even modify the gray color in that array as well. Food for thought

dmfalke:
> One thing i want to play with is if we can get the highlighted node's edges to render on top of all the others. Not crucial for this PR since it's adding so much value already, but something we should figure out

I thought about this, too. I think we have to reorder the links such that the highlighted ones are at the end of the array.

packages/libs/components/src/plots/BipartiteNetwork.tsx
dmfalke: Yeah, I'm inclined to wait for consolidation. I was going to move things around now, but decided it was better to wait for a clear path forward.

packages/libs/components/src/plots/BipartiteNetwork.tsx
dmfalke: Sure, that sounds good to me!

packages/libs/components/src/plots/BipartiteNetwork.tsx
dmfalke: I think I agree that this should go in the viz

dmfalke:
> > One thing i want to play with is if we can get the highlighted node's edges to render on top of all the others. Not crucial for this PR since it's adding so much value already, but something we should figure out
> 
> I thought about this, too. I think we have to reorder the links such that the highlighted ones are at the end of the array.

I went ahead and did this.

packages/libs/components/src/plots/BipartiteNetwork.tsx
dmfalke: done

PR #958, opened 2024-03-23
refs #53931

aurreco-uga:
dave, would be better to change tabs without reloading the page.. 
also will need to add this in the new vdi branch.

### DIFF
--- a/packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
+++ b/packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
@@ -372,6 +372,11 @@ function UploadForm({
       {errorMessages.length > 0 && <ErrorMessage errors={errorMessages} />}
       <div>
         <h2>{datasetUploadType.uploadTitle}</h2>
+        <div>
+          Before uploading your dataset, please ensure your data is formatted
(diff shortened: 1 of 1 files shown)


packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -372,6 +372,11 @@ function UploadForm({
       {errorMessages.length > 0 && <ErrorMessage errors={errorMessages} />}
       <div>
         <h2>{datasetUploadType.uploadTitle}</h2>
+        <div>
+          Before uploading your dataset, please ensure your data is formatted
+          according to the instructions listed in the{' '}
+          <a href="../datasets/help">"Help" tab</a>.
```
dmfalke, line 378: You can use

```ts
import { Link } from 'react-router-dom';

// ...

<Link to={{ path: '../datasets/help' }}>"Help" tab</Link>
```

dmfalke (commented):
> dave, would be better to change tabs without reloading the page..

See my comment below

jernestmyers:
Updated to render a warning banner with the message:

image

dmfalke (approved):

PR #963, opened 2024-03-25
testing alternative approach

moontrip:
This was branched out from https://github.com/VEuPathDB/web-monorepo/pull/948 to test Bob's suggestion, https://github.com/VEuPathDB/web-monorepo/pull/948#discussion_r1533827356. It seems to work fine as well, but leave it as draft for now.


moontrip:
@bobular now I made this as regular PR since as you commented before, https://github.com/VEuPathDB/web-monorepo/pull/948#discussion_r1533827356, it is now resolved (being merged) so we don't use useMapEvent inside SemanticMarkers. So this will possibly replace https://github.com/VEuPathDB/web-monorepo/pull/948 as an alternative. Though, as I mentioned before, https://github.com/VEuPathDB/web-monorepo/pull/948 is still valid. 

If you prefer not to have useMapEvent then this would be an alternative solution as we discussed before

packages/libs/components/src/map/SemanticMarkers.tsx
```diff
@@ -205,10 +208,34 @@ export default function SemanticMarkers({
...
+          consolidatedMarkers.length === 0
+            ? 0
+            : consolidatedMarkers[0].props.id.length;
+
+        const selectedMarkersElementLength =
+          selectedMarkers.length === 0 ? 0 : selectedMarkers[0].length;
+
+        const message =
+          consolidatedMarkersElementLength !== selectedMarkersElementLength
+            ? 'marker selection has been cancelled because aggregation level has changed due to zooming'
```
bobular, line 222: Consistent capitalization to match previous marker-related snackbars
```suggestion
            ? 'Marker selection has been cancelled because aggregation level has changed due to zooming'
```

packages/libs/components/src/map/SemanticMarkers.tsx
```diff
@@ -205,10 +208,34 @@ export default function SemanticMarkers({
...
+            : 'selected markers that went off-screen have been deselected';
```
bobular, line 223:
```suggestion
            : 'Selected markers that went off-screen have been deselected';
```

packages/libs/components/src/map/MapVEuMap.tsx
```diff
@@ -443,7 +443,7 @@ function CustomZoomControl(props: CustomZoomControlProps) {
         <div className="leaflet-control-zoom leaflet-bar leaflet-control">
           <a
             className={'leaflet-control-zoom-in' + disableMaxZoomButton}
-            href="#"
+            href="/#"
```
bobular, line 446: Could you explain these changes please?

I have played around and for some reason any changes I make don't affect the zoom controls on screen. Very odd.

bobular (commented):
I think the simple approach here is working fine.

I'd approve but I'd like to clear up the `href` issue first.

packages/libs/components/src/map/SemanticMarkers.tsx
moontrip: @bobular Yep I will change it. Thanks 👍

packages/libs/components/src/map/MapVEuMap.tsx
moontrip: @bobular As I mentioned before, it was simply to fix ESLint complaints, not a change to fix something.

packages/libs/components/src/map/SemanticMarkers.tsx
moontrip: @bobular Done for both

bobular:
Ah, thanks and sorry if I asked this twice!

If it isn't functionally important then I approve (which I don't think I
can do by email)

bobular (approved):
Thumbs up

moontrip:
Thanks @bobular for your quick reviews 👍

Issue #961, opened 2024-03-25
EDA/SAM histogram viz - binWidth should be empty by default


dmfalke:
Histogram visualizations are currently using `xAxisVariable.distributionDefaults.binWidth` for the default binWidth, but this is not correct. It should use `undefined` by default.

See https://github.com/VEuPathDB/web-monorepo/issues/959 for details.

d-callan:
might be worth looking at line plot/ timelines as well, to be sure they arent doing similar.

bobular:
I totally agree with letting the back end determine the binWidth etc for visualizations when the user hasn't set a binWidth explicitly. However, I found a "back door" that allows the client to send inappropriate binWidths to the back end.

VB megastudy using @moontrip's PR #964 
Zoomed out
+ floating histogram of "count of traps" - all is good.
set bin width to 10 manually - all is still good
zoom in to (or select by clicking) the central Madagascar 2428 marker:
![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/1d2bdfce-042b-4d3d-b80b-a495d536094f)

The histogram request still contains the `binWidth: 10` and causes some havoc on the back end. Here are the rserve logs: @d-callan 

```
2024-03-26 21:10:35.806114 Determined bin width slider min, max and step values.
Error in veupathUtils::cut_width(xVP, binWidth, boundary = min(xVP)) : 
  Less than two breaks in utils-cut.R
In addition: Warning messages:
1: replacing previous import ‘veupathUtils::toJSON’ by ‘jsonlite::toJSON’ when loading ‘plot.data’ 
2: In veupathUtils::nonZeroRound(binSliderMax, avgDigits) :
  Input is already zero and cannot be rounded to a non-zero number.
3: In veupathUtils::nonZeroRound(binSliderMin, avgDigits) :
  Input is already zero and cannot be rounded to a non-zero number.
4: In veupathUtils::nonZeroRound(((binSliderMax - binSliderMin)/1000),  :
  Input is already zero and cannot be rounded to a non-zero number.
```

This can be reproduced in regular EDA too. (Make histogram, change bin width, then filter on country=Madagascar ; though you can only reach the regular EDA for the megastudy from the dev "react server" I think)

I'm having a think about what the client could do to mitigate this. If possible!

d-callan:
[...]

PR #964, opened 2024-03-26
set default binWidth as undefined for HistogramViz and Lineplot Viz

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/961. Implementation didn't take too much time but tests did 😅. Changed both HistogramViz and Lineplot Viz: time series worked just fine after changing Lineplot Viz. 

One note is that backend did return incorrect binWidthRange for the example case in the ticket (please see the screenshot below), { min: 0.1, max: 0 }. @d-callan? 

histogram_binWidthRange01

Here are some screenshots with this PR:
histogram_binWidthRange00

hisgotram_binWidthRange02

lineplot_binWidthRange00

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -1548,9 +1548,6 @@ function getRequestParams(
     xAxisVariable,
   } = config;
 
-  //DKDK
-  console.log('binWidth, binWidthTimeUnit =', binWidth, binWidthTimeUnit);
-
   const binSpec: Pick<HistogramRequestParams['config'], 'binSpec'> = binWidth
     ? {
         binSpec: {
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
@@ -1540,12 +1540,7 @@ function getRequestParams(
   ) => HistogramRequestParams
 ): HistogramRequestParams {
   const {
-    binWidth = NumberVariable.is(variable) || DateVariable.is(variable)
-      ? config.independentAxisValueSpec === 'Full'
-        ? variable.distributionDefaults.binWidthOverride ??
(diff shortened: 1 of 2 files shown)


moontrip:
> That looks like a plot.data bug. Mind making an issue there w your screenshots?

@d-callan Sure! Thank you for your quick assessment 👍

moontrip:
@d-callan Made one at plot.data for you, https://github.com/VEuPathDB/plot.data/issues/255 :)

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -1540,12 +1540,7 @@ function getRequestParams(
...
-    binWidth = NumberVariable.is(variable) || DateVariable.is(variable)
-      ? config.independentAxisValueSpec === 'Full'
-        ? variable.distributionDefaults.binWidthOverride ??
-          variable.distributionDefaults.binWidth
-        : undefined
-      : undefined,
+    binWidth = undefined,
     binWidthTimeUnit = variable?.type === 'date'
       ? variable.distributionDefaults.binUnits
       : undefined,
```
bobular, line 1546: The `= undefined` is great for readability in the diff, but not necessary for functionality.

Furthermore, we don't need to set `binWidthTimeUnit` from the `variable.distributionDefaults` because we will never pass it to the back end in `binSpec` if `binWidth` is falsy.

```suggestion
    binWidth,
    binWidthTimeUnit,
```

The same goes for Lineplot

bobular (changes requested):

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: @bobular Yes you are correct. I was thinking of defining it explicitly, rather than implicitly to make it clearer, but we can simply not define it as you suggested. I was also gauging whether binWidthTimeUnit was set to be undefined as well, but I just left it as was 😅 Anyway I will update them.

bobular (approved):
Looks great. Glad it was an easy fix!

PR #965, opened 2024-03-26
fixing broken visx stories - Network & Volcano

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/920

It turned out that as Ann thought, it was correct that `import React from 'react';` was needed. I tracked down all error messages carefully, and added the part for all relevant files.

If I understand correctly, Dave has set up auto inclusion of those react import for web-monorepo (babel and/or webpack config), but I think that storybook did not have such configuration, which requires to have the react import in that case.

I checked all sub-plots and all seem to work correctly without complaints :)


dmfalke:
@moontrip I made a [PR](https://github.com/VEuPathDB/web-monorepo/pull/953) to address this the other day. I must have missed this issue... sorry!! My fix is to change the order of babel loaders.

PR #975, opened 2024-03-29
Reset table paging onSort and render row counts in Browse & Subset table

### LINKED ISSUE #839: UX: eda/genomics : when browsing a categorical variable with many values, sorting should reset paging to first page
mar 282024 sam encounters another issue with this https://epvb.slack.com/archives/CMQ8K9XRS/p1711643330714609

----------

playing with the popset-eda demo, i selected a study (see snapshot), and i get 55 samples.
https://qa.restricted.clinepidb.org/analysis/UwKE5TW

I wanted to find out what organisms in these 55 samples so i sorted to bring the non zero counts on view..  but all was greyed out 0. 
The confusion was that I was stuck on page 6 unknowingly... 
**I think  that when sorting on a column we should reset the page to 1.**

Screen Shot 2024-02-09 at 12 52 13 PM

jernestmyers:
Resolves #839 

~~Of particular note in this PR is the use of a `setTimeout` hack to ensure both state setters update the `uiState`. This hack was implemented directly in the mesa headers but also "necessary" in the membership table's filter. Of course, I'm open to better implementations given the stinky code smell going on with the hack!~~

Been updated to reset paging in the consumers of Mesa, not in Mesa itself. It's quite likely there are other tables that need to be updated, but this PR addresses the tables of concern for now.


dmfalke:
Hi Jeremy. I'm not crazy about using `setTimeout` in this way. I need to take a closer look, before I can make any suggestions. I will follow up later today or tomorrow.

dmfalke:
I think an alternative approach would be to modify https://github.com/VEuPathDB/web-monorepo/blob/82d6bd65d5ba2edfe3c9b288d4dde5c5851f704b/packages/libs/wdk-client/src/Views/Question/Params/FilterParamNew/FilterParamNew.tsx#L107, and set the current page there.

The downside is that this does not address all cases where we use paging in `Mesa`, so we could end up with inconsistent behavior (which we probably already have). But I do think this is better than the `setTimeout` solution, since it causes multiple requests in the strategy panel (see two for `standard`, below):

image

dmfalke (approved):
looks good!

bobular:
Also can confirm that this also addresses the desired "keep toggled at top" behaviour (to also switch to page 1)

bobular:
What's the concern about inconsistent behaviour? If we do the sorting externally to the Mesa component?

jernestmyers:
> What's the concern about inconsistent behaviour? If we do the sorting externally to the Mesa component?

My impression is just that the consumer of Mesa will have to remember to implement the sorting such that it resets the paging, which may introduce future inconsistencies since it's not just built into the sorting behavior of Mesa. And at present, I didn't spend much time trying to track down every Mesa instance in order to add this logic so it's quite likely there will be paged tables that won't reset to page 1 with a sort change.

PR #983, opened 2024-04-02
Add better typescript safety to legacy redirect state

jernestmyers:
Follow up work in response to https://github.com/VEuPathDB/web-monorepo/pull/936#discussion_r1532711008


packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -123,8 +121,11 @@ export function MapAnalysis(props: Props) {
   );
   const geoConfigs = useGeoConfig(useStudyEntities());
   const location = useLocation();
-  const locationState = location.state as LegacyRedirectState;
-  const [showRedirectModal, setShowRedirectModal] = useState(!!locationState);
+  const locationState = location.state;
+  const [showRedirectModal, setShowRedirectModal] = useState(
+    LegacyRedirectState.is(locationState) &&
+      locationState?.showLegacyMapRedirectModal
```
bobular, line 127: Doing this (or `Boolean(...)`) would keep this state variable 'pure' Boolean which would remove the need to cast it below.

```suggestion
      !!locationState?.showLegacyMapRedirectModal
```

bobular (approved):
This looks great. Just a tiny style suggestion that you may ignore!

Issue #969, opened 2024-03-27
Histogram and Lineplot visualizations: provide a reasonable viewport if sending a non-empty binSpec


bobular:
There's still an edge case bug described in https://github.com/VEuPathDB/web-monorepo/issues/961#issuecomment-2021509519 

And a possible solution is outlined below:

> In a case where the subset changes, what do we currently do about axes? Does the visible range persist or change w the subset? Does that change for different modes (full v auto zoom)? Id personally be ok, if the user didn't set the bin width and got one for free to do the same for a new subset. But if we choose to not do that, then at min being consistent w the axes is important. The backend won't complain about these types of bin widths if you also provide it a reasonable viewport.

_Originally posted by @d-callan in https://github.com/VEuPathDB/web-monorepo/issues/961#issuecomment-2021607027_
            

#### How would this be implemented?

Somehow we need to send the min and max of `defaultIndependentRange` to the histogram request's `viewport` (same goes for lineplot).  This could be done by
1. moving the definition of `onBinWidthChange` down a bit so that it can set the config's `independentAxisRange` to the values of `defaultIndependentRange` so it will automatically get sent as part of the request. This may also require shifting the user into "Custom" x-axis range mode, as that is effectively what we're doing (you can fix the 500 by going custom mode and setting a sensible range)
2. passing `defaultIndependentRange` into the get data promise and into getRequestParams and using it as a second-priority value for `viewport` within (when a `binWidth` is provided).

If 2. isn't too gnarly I think that would be the better UX.

PR #986, opened 2024-04-03
set default viewport for Histogram Viz

moontrip:
This will probably address https://github.com/VEuPathDB/web-monorepo/issues/969. Please check if its behavior is what you had in mind.

It is worth noting that this could be done for Histogram Viz because the defaultIndependentRange is computed from other API call (distribution), not from data. However, it is not the case for Lineplot Viz, as its defaultIndependentRange is based on data. Accordingly, default viewport for Lineplot Viz cannot be made before obtaining data from the backend.


bobular (commented):
I've tested my suggestion and I think this is probably the prefered solution. My reasoning is that we only get 500s with user-provided binWidth but no viewport (e.g. not using custom x-axis range). I think we should fix the 500s without changing existing behaviour, and always providing the viewport does change the backend-calculated binWidth returned.

I looked into the Lineplot and we can reproduce the 500 with the same procedure:

* GEMS1
* no filters
* lineplot: x=Sleeping rooms in dwelling count, y=whatever you like
* change binWidth to 10
* go to subsetting and filter on Sleeping rooms in dwelling count 10 to 10 (I know nobody would do this, but it simulates the single-valued variable, like count of traps in popbio)
* go back to visualisation and observe the 500

In dev tools you can edit the request and add `"viewport": { "xMin": 0, "xMax": 100}` and the 500 is gone.

We are not sending viewport at all to lineplot - which seems like an oversight?

I note your comments, @moontrip, about not having the data in-hand to send the viewport, I will take a closer look to see if we can solve this simply.

bobular:
I think something like this in LineplotVisualization's `getRequestParams` would work?

```
  const viewport = binWidth ?
    vizConfig.independentAxisRange != null ?
    { xMin: String(vizConfig.independentAxisRange.min),
      xMax: String(vizConfig.independentAxisRange.max)
    } :
    (xAxisVariableMetadata.type === 'number' || xAxisVariableMetadata.type === 'date') && xAxisVariableMetadata.distributionDefaults != null ?
    { xMin: String(xAxisVariableMetadata.distributionDefaults.rangeMin), // do we need to consider
      xMax: String(xAxisVariableMetadata.distributionDefaults.rangeMax), // displayRangeMin/Max too?
    } : undefined
    : undefined;
  
  return (
    customMakeRequestParams?.({
      studyId,
      filters,
      vizConfig,
      outputEntityId: outputEntity?.id,
      valueSpec,
      binSpec,
      errorBars,
    }) ?? {
      studyId,
      filters,
      config: {
        outputEntityId: outputEntity?.id,
        valueSpec,
        xAxisVariable: xAxisVariable!, // these will never be undefined because
        yAxisVariable: yAxisVariable!, // data requests are only made when they have been chosen by user
        ...binSpec,
	viewport,
        overlayVariable: overlayVariable,
        facetVariable: facetVariable ? [facetVariable] : [],
        showMissingness: showMissingness ? 'TRUE' : 'FALSE',
        errorBars,
        ...(valueSpec === 'proportion'
          ? {
              yAxisNumeratorValues: numeratorValues,
              yAxisDenominatorValues: denominatorValues,
            }
          : {}),
      },
    }
  );
}
```

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -1569,6 +1571,12 @@ function getRequestParams(
           xMin: config?.independentAxisRange?.min,
           xMax: config?.independentAxisRange?.max,
         }
+      : defaultIndependentRange?.min != null &&
```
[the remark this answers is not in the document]
moontrip: @bobular thank you for your comment. I added the additional condition

moontrip:
> I think something like this in LineplotVisualization's `getRequestParams` would work?
> 
> ```
>   const viewport = binWidth ?
>     vizConfig.independentAxisRange != null ?
>     { xMin: String(vizConfig.independentAxisRange.min),
>       xMax: String(vizConfig.independentAxisRange.max)
>     } :
>     (xAxisVariableMetadata.type === 'number' || xAxisVariableMetadata.type === 'date') && xAxisVariableMetadata.distributionDefaults != null ?
>     { xMin: String(xAxisVariableMetadata.distributionDefaults.rangeMin), // do we need to consider
>       xMax: String(xAxisVariableMetadata.distributionDefaults.rangeMax), // displayRangeMin/Max too?
>     } : undefined
>     : undefined;
>   
>   return (
>     customMakeRequestParams?.({
>       studyId,
>       filters,
>       vizConfig,
>       outputEntityId: outputEntity?.id,
>       valueSpec,
>       binSpec,
>       errorBars,
>     }) ?? {
>       studyId,
>       filters,
>       config: {
>         outputEntityId: outputEntity?.id,
>         valueSpec,
>         xAxisVariable: xAxisVariable!, // these will never be undefined because
>         yAxisVariable: yAxisVariable!, // data requests are only made when they have been chosen by user
>         ...binSpec,
> 	viewport,
>         overlayVariable: overlayVariable,
>         facetVariable: facetVariable ? [facetVariable] : [],
>         showMissingness: showMissingness ? 'TRUE' : 'FALSE',
>         errorBars,
>         ...(valueSpec === 'proportion'
>           ? {
>               yAxisNumeratorValues: numeratorValues,
>               yAxisDenominatorValues: denominatorValues,
>             }
>           : {}),
>       },
>     }
>   );
> }
> ```

@bobular Thank you for your suggestion. I will check it out

moontrip:
@bobular I have tested and your suggestion above for Lineplot viz works. I had thought we should have used data-based range instead of metadata (annotation) based range to pass a range into viewport, which was the reason why I didn't try to make a default viewport - I saw your previous comment somewhere that metadata's range is often incorrect with sub-settings but I might be wrong. Anyway if you think metadata-based range is good enough (at least for the initial range) then it would be fine.

I have made a commit with Lineplot Viz's viewport as you suggested, https://github.com/VEuPathDB/web-monorepo/pull/986/commits/7b8abd6f39dc2dcdd72d6801626ac1f48ff1fe53

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2130,6 +2130,24 @@ function getRequestParams(
       ? 'TRUE'
       : 'FALSE';
 
+  const viewport = binWidth
```
bobular, line 2133:
```suggestion
  // If a binWidth is passed to the back end, we should send a viewport too.
  // This helps prevent a 500 in edge-case scenarios where the variable is single-valued.
  // It doesn't really matter what the viewport is, but we send either the custom range
  // or the variable's automatically annotated range.
  // The back end requires strings for some reason.
  const viewport = binWidth
```

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2130,6 +2130,24 @@ function getRequestParams(
...
+    ? vizConfig.independentAxisRange != null
+      ? {
+          xMin: String(vizConfig.independentAxisRange.min),
+          xMax: String(vizConfig.independentAxisRange.max),
+        }
+      : (xAxisVariableMetadata.type === 'integer' ||
+          xAxisVariableMetadata.type === 'number' ||
+          xAxisVariableMetadata.type === 'date') &&
+        xAxisVariableMetadata.distributionDefaults != null
+      ? // do we need to consider displayRangeMin/Max too?
```
bobular, line 2143: let's remove this comment

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
```diff
@@ -1569,6 +1571,13 @@ function getRequestParams(
           xMin: config?.independentAxisRange?.min,
           xMax: config?.independentAxisRange?.max,
         }
+      : binWidth &&
```
bobular, line 1574:
```suggestion
      // send back end a viewport to prevent edge-case 500s with single-valued variables when a binWidth is provided
      : binWidth &&
```

bobular (approved):
Looks great. Just added some comments to document what's going on.

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
moontrip: @bobular Thanks. I added those paragraph 👍

packages/libs/eda/src/lib/core/components/visualizations/implementations/HistogramVisualization.tsx
moontrip: Done!

PR #990, opened 2024-04-04
Condition data access verbiage in download tab wrt projectDisplayName && isUserStudy

### LINKED ISSUE #927: EDA UD: Download: improve access verbiage
Screenshot 2024-03-08 at 12 38 48 PM

### LINKED ISSUE #989: VDI UDs: Update download tab text for user datasets
The download tab for user datasets says:

Screen Shot 2023-12-05 at 6 49 07 AM

Can we either
1. remove the Data Accessibility section for user datasets
2. update the text so that it says something along the lines of "This is your data. Only you and those with whom you've shared this data can download it."

jernestmyers:
Resolves #989
Resolves #927

Note that I've amended the logic to only display this verbiage in ClinEpiDB for non-user studies. Otherwise, this verbiage won't display in MicrobiomeDB or genomics sites irrespective of its `isUserStudy` status.

I also took a chance to clean up some of the styling. I didn't like how I handled the spacing via margins so set the container to be flex and used gap instead.

MicrobiomeDB:
image

User study in ClinEpiDB:
image

Curated study in ClinEpiDB:
image

### DIFF
--- a/packages/libs/eda/src/lib/workspace/DownloadTab/index.tsx
+++ b/packages/libs/eda/src/lib/workspace/DownloadTab/index.tsx
@@ -32,6 +32,8 @@ import { showLoginForm } from '@veupathdb/wdk-client/lib/Actions/UserSessionActi
 import { useHistory } from 'react-router';
 import { parsePath } from 'history';
 
+const EDA_USER_DATASET_PREFIX = 'EDAUD';
+
 type DownloadsTabProps = {
   downloadClient: DownloadClient;
   analysisState: AnalysisState;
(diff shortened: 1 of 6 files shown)


packages/libs/eda/src/lib/workspace/DownloadTab/index.tsx
```diff
@@ -32,6 +32,8 @@ import { showLoginForm } from '@veupathdb/wdk-client/lib/Actions/UserSessionActi
 import { useHistory } from 'react-router';
 import { parsePath } from 'history';
 
+const EDA_USER_DATASET_PREFIX = 'EDAUD';
```
asizemore, line 35: Mind adding a comment here about where the prefix comes from? I'm guessing some serviec in the backend defines it? Or where should we look in case this changes?

asizemore (approved):
Code looks good and i like the screenshots!
I haven't had a chance to test yet, i'm having trouble getting everything to run (a me thing, not your code at all), but i think it's worth an approval even without the test!

packages/libs/eda/src/lib/workspace/DownloadTab/index.tsx
jernestmyers: I've actually removed this code since the service is now sending us an `isUserStudy` attribute via https://github.com/VEuPathDB/EdaSubsettingService/pull/139

Issue #952, opened 2024-03-21
Map - click to deselect and/or close left hand panel is unreliable


bobular:
This could be related to #902 (we may decide on different UX)

It seems that the use of two different `useMapEvents` to attach different handlers to the `click` event is causing unpredictable behaviour. 

The handlers are
* `MapVEuMap` - close left hand menu panel 
* `SemanticMarkers` - deselect all markers

The menu panel closing is particularly unreliable. If we comment out the `SemanticMarkers` `useMapEvents` then it works perfectly.

There's nothing obvious on the web about attaching multiple handlers to the same event via different calls to the hook, but ChatGPT says it's likely only resolvable by consolidating the event handling somehow.

Issue #902, opened 2024-03-01
Map - revisit/refine auto-closing behaviour of the left hand panel/menu


bobular:
It would be great if we can distinguish between user actions that should close the menu (click, pan, zoom) and automatic actions (zoom during fly-to) that shouldn't.

An annoying behaviour at the moment is that upon opening a new analysis, the user might open the menu to make a filter, but in the background, markers are being retrieved and the fly-to that happens when they arrive will close the menu.

Related to potential overhaul here: 
https://github.com/VEuPathDB/web-monorepo/issues/628

Also note that a single click on the map is no longer closing the panel.

Sub tasks

- [ ] restore old intended behaviour: single click on map -> close panel
- [ ] prevent fly-to from closing the panel (unless the user clicked on the fly-to button perhaps?)
- [ ] panel should not close if user resizes their browser window

dmfalke:
My 2 cents: we should close the menu based on "event" (aka, via callbacks passed to various components), rather than based on changes to the app state.

PR #992, opened 2024-04-09
adjusting map click events

moontrip:
This will possibly address #902 and #952. 

As reported in #952, map event at `SemanticMarkers` did not work correctly. Thus, it is changed to handle map events at `MapVEuMap`. To achieve this, a combined handler (a callback function) at `MapAnalysis` that will function to deselect selectedMarkers and close side panel is attached to map click event. Throughout many tests, I couldn't see any specific issue.

Concerning sub tasks at #902:

- [x] restore old intended behaviour: single click on map -> close panel
  - this now works
- [x] prevent fly-to from closing the panel (unless the user clicked on the fly-to button perhaps?)
  - the reason why fly-to closed side panel was we implemented to close side panel when zoom event occurs. Thus, I removed the event so that side panel remains open when fly-to happens. Since single click on map can easily close side panel, I think this would be fine.
- [x] panel should not close if user resizes their browser window 
  - this was because we also set to close panel when map panning occurs. Thus, I removed that event callback which resolves this.

A simple video capture is attached in the following:

https://github.com/VEuPathDB/web-monorepo/assets/12802305/d4d6a29c-5d92-40f3-97e4-0a8f7c45c25c


packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
```diff
@@ -772,11 +774,20 @@ function MapAnalysisImpl(props: ImplProps) {
...
+      ...(activeMarkerConfiguration as PieMarkerConfiguration),
+      selectedMarkers: undefined,
+    });
+    // close side panel
+    setIsSidePanelExpanded(false);
+  }, [
+    updateMarkerConfigurations,
+    activeMarkerConfiguration,
+    setIsSidePanelExpanded,
+  ]);
```
bobular, line 790: This seems to remove the arbitrary choice of `PieMarkerConfiguration` & function renaming removes need for comments. Don't forget to remove unused imports.

```suggestion
  const deselectMarkersAndClosePanel = useCallback(() => {
    updateMarkerConfigurations({
      ...(activeMarkerConfiguration as MarkerConfiguration),
      selectedMarkers: undefined,
    });
    setIsSidePanelExpanded(false);
  }, [
    updateMarkerConfigurations,
    activeMarkerConfiguration,
    setIsSidePanelExpanded,
  ]);
```

bobular (approved):
This is working really nicely.  Great that the flyTo and window resizing no longer close the panel! I think users will learn to do a single (now reliable!) click, rather than pan or zoom to close the panel. In fact, users might want to pan the map while keeping the panel open so they can see certain markers.

I made a code suggestion - please double check and incorporate so we can clean it up a little bit more.

packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
moontrip: Thank you for your valuable comment @bobular 👍 I missed MarkerConfiguration 😅 Will comply with your comments and then merge

Issue #895, opened 2024-02-28
Map - revisit leaflet zoom level to geohash level mapping


bobular:
Now we have made zooming less granular (#763) there are zoom levels which have way too many markers, e.g. below:

![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/f295cfcb-22a1-4b5d-9ed5-4e7e93101997)

We should adjust the thresholds to make sure this doesn't happen.

Not sure if the `Math.round()` is necessary.

This needs to be tested with various screen sizes (and pixel density if appropriate?)

PR #993, opened 2024-04-09
Improve zoom to geohash level mapping

chowington:
Closes https://github.com/VEuPathDB/web-monorepo/issues/895


bobular (approved):
This is great. While zooming out, just as the markers start getting crowded, the geohash level changes. I can't see how different screen sizes would make a difference to this, so let's go with this :-)

PR #997, opened 2024-04-11
Update param UI to reflect visibleHelpPosition set to 'right'

jernestmyers:
Depends on the second commit in https://github.com/VEuPathDB/WDK/pull/86 that adds `visibleHelpPosition` as an optional attribute to params.

A `visibleHelpPosition: 'right'` would display as shown:
image


dmfalke (approved):
LGTM!

PR #1001, opened 2024-04-12
Prevent undefined ortho table props error

jernestmyers:
QA noticed a bug in ortho tables when `value.length` is 0. Instead of passing in `orthoTableProps` in the `if` statement of `GeneRecordClasses.GeneRecordClasses`- which would subsequently not get consumed anyway by `RecordTable`- I decided to amend `isOrthologTable` to be `isOrthologTableWithData`. Now, in the `RecordTable` we won't try to access `this.props.orthologProps` if we detect no table values.


dmfalke (approved):
Nice!

PR #1014, opened 2024-04-16
Use real links

### LINKED ISSUE #1007: eda genomics: wgcna network : link to gene list: new tab if different website (maybe even if same site?)
one can't use the right click to open in a new tab..

Screenshot 2024-04-15 at 1 28 01 PM

dmfalke:
Using real links allows users to open the link in a new tab with ctrl-click or right-click, etc.

closes #1007


asizemore (approved):
Looks great! I must admit i miss the look of the menu, but since we're possibly moving away from the menu in favor of a panel, we shouldn't worry about the look :)

PR #1015, opened 2024-04-16
Handle user dataset upload errors

### LINKED ISSUE #1011: VDI - non-success response for creating a dataset is not handled well
The client is not handling non-success responses for creating a new dataset. The browser ends up on a route like https://qa.plasmodb.org/plasmo.b68/app/workspace/datasets/undefined. We should handle the non-success response and show an error message on the upload form.

jernestmyers:
Resolves #1011

image

### DIFF
--- a/packages/libs/user-datasets/src/lib/Service/api.ts
+++ b/packages/libs/user-datasets/src/lib/Service/api.ts
@@ -102,13 +102,10 @@ export class UserDatasetApi extends FetchClientWithCredentials {
         }
       }
       if (xhr.readyState === XMLHttpRequest.DONE && xhr.status >= 400) {
-        try {
-          throw new Error(xhr.response);
-        } catch (error) {
-          dispatchUploadProgress && dispatchUploadProgress(null);
-          dispatchBadUpload && dispatchBadUpload(String(error));
-          console.log(error);
-        }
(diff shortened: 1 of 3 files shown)


packages/libs/user-datasets/src/lib/Service/api.ts
```diff
@@ -100,6 +101,15 @@ export class UserDatasetApi extends FetchClientWithCredentials {
...
         }
       }
+      if (xhr.readyState === XMLHttpRequest.DONE && xhr.status >= 400) {
+        try {
+          throw new Error(xhr.response);
+        } catch (error) {
+          dispatchUploadProgress && dispatchUploadProgress(null);
+          dispatchBadUpload && dispatchBadUpload(String(error));
+          console.log(error);
+        }
```
dmfalke, line 111: I don't think you need the try/catch here. You should be able to just do

```suggestion
          const error = new Error(xhr.response);
          dispatchUploadProgress && dispatchUploadProgress(null);
          dispatchBadUpload && dispatchBadUpload(String(error));
          console.log(error);
```

dmfalke (commented):
Just skimmed the code so far. Looks good. I made one suggestion. I'll test later.

dmfalke (approved):
Looks great!

PR #1017, opened 2024-04-17
Add groupBySelected filter to orthologs record table

jernestmyers:
Resolves https://redmine.apidb.org/issues/54091


dmfalke (approved):
I think this looks nice!

PR #1018, opened 2024-04-17
Markers automatically use the most appropriate geoEntity when there are multiple to choose from

### LINKED ISSUE #1024: Make studies with more than one geo-located entity work on the SAM
This only concerns **Transmission Zero Field Trials**, on qa.restricted.clinepi. It has four entities with lat/long/geohashes. A root entity (which was previously the only entity being 'mapped') and its three children.

Before this PR, everything was shown at the center of the geohash_4 grid.

Now, if you pick variables from any of the non-root entities, the markers will be shown at the position of the appropriate entity, not the artificial root entity.

bobular:
Fixes #1024 

This only affects the Transmission Zero Field Trials prototype dataset at the moment (qa restricted clinepi) because this is the only study where more than one entity has geo variables.

Luckily a lot of the code was already passing around `geoConfigs` rather than `geoConfigs[0]` so there wasn't a huge amount of upheaval.

I have implemented this as if there was a drop-down for the user to choose the geoEntity they wanted, but instead have chosen the least ancestral entity for them automatically. So this is why this "choice" is stored in appState. We can add UI later if needed.

- [x] Donut mode
- [x] Bar mode
- [x] Bubble mode

### DIFF
--- a/packages/libs/eda/src/lib/core/utils/geoVariables.ts
+++ b/packages/libs/eda/src/lib/core/utils/geoVariables.ts
@@ -1,5 +1,6 @@
 import { GeoConfig } from '../types/geoConfig';
 import { StudyEntity } from '../types/study';
+import { findLast } from 'lodash';
 
 /**
  * Given an entity note in the study tree, check to see if it has the following variables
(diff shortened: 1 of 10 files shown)


dmfalke:
I'm having trouble remembering how this will manifest in the UI. Where will I see a change in behavior?

bobular:
> I'm having trouble remembering how this will manifest in the UI. Where will I see a change in behavior?

The only study with multiple geo-entities (geoConfigs) is Transmission Zero Field Trials, so for every other study this PR is a 'no op' - all the logic returns `geoConfigs[0]`

If you compare qa.restricted.clinepi with this PR you'll see that when you pick an adult entomology sample variable like "Morphological species ID" the locations are different. In the main branch, only the root entity locations are used for markers. In the PR, the closest entity to the chosen variable is used.

dmfalke:
Ah, okay! So it only impacts location? Anyway, this looks great to me so far. I will test w/ the mega study, just to be sure.

bobular:
I realised I didn't do the other marker modes yet. Doing now!

dmfalke:
> I realised I didn't do the other marker modes yet. Doing now!

:smile: 

On that note, I'm primarily looking for fatal errors. I think we need to rely on data experts for a more robust qa on this update.

bobular:
I am the data expert on this study...

dmfalke:
> I am the data expert on this study...

I know :smile:

dmfalke (approved):
Changes look good. Nothing appears broken. :rocket:

bobular:
Added an issue to help with QA. Merging now. Thank you!

PR #1019, opened 2024-04-18
Temporarily spoof the dataElementDependencyOrder for histogram

bobular:
This just a test - up for discussion and maybe to help with a demo this+next week.

I've changed the `dataElementDependencyOrder` from
```
{
	"dataElementDependencyOrder": [
		[
			"xAxisVariable"
		],
		[
			"overlayVariable"
		]
	]
}
```
to
```
{
	"dataElementDependencyOrder": [
		[
			"xAxisVariable",
			"overlayVariable"
		]
	]
}
```

So that we can stratify a time-based histogram by species (see below in the Transmission Zero Field Trials prototype). When switching between stratify and no-stratify, the output entity (see the plot title) changes appropriately. So far I can't see any downsides...

image

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/MapAnalysis.tsx
@@ -286,7 +286,23 @@ function MapAnalysisImpl(props: ImplProps) {
   const appsPromiseState = usePromise(
     useCallback(async () => {
       const { apps } = await dataClient.getApps();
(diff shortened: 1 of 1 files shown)


bobular:
I've found one problem.

You can get a 400 by changing the marker variable while the plot is in "stratify mode".
The reason is that `useFindOutputEntity` isn't using `providedOverlayVariableDescriptor`. It should be fixable.

PR #1032, opened 2024-04-19
Revert "Only show Management Dashboard button for user datasets"

dmfalke:
Reverts VEuPathDB/web-monorepo#1003

This logic is not correct. If anything, we only want to show the button if a study has any access restrictions.

### DIFF
--- a/packages/libs/eda/src/lib/workspace/EDAWorkspaceHeading.tsx
+++ b/packages/libs/eda/src/lib/workspace/EDAWorkspaceHeading.tsx
@@ -77,8 +77,7 @@ export function EDAWorkspaceHeading({
           </H3>
         )}
         <div>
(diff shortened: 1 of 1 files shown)


dmfalke:
Force-merging so this can get to qa asap

aurreco-uga:
frankly i would just  fix the bug for now.. so it only shows when needed.. 
danica describes a situation when it will be needed in public studies (in qa sites providers wont be able to see other providers studies even if public) 
 but that might not be the case for a while and i rather eliminate clutter..  and will be clue on how things are working.

dmfalke:
> in qa sites providers wont be able to see other providers studies even if public

I don't really understand this. Who exactly would it be "public" for?

Anyhow, I will work on a fix for not showing the button for public studies soon.

Issue #1031, opened 2024-04-19
Map - little blue squares persist after panning


bobular:
This isn't just me - seeing a lot of these in screenshots from qa posted on Slack today

![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/7bdf3c19-404a-41da-9d4d-f205f2c345d5)

(they are blue squares on my screen - the screenshot has antialiased them or something, but you can still see them)

PR #1041, opened 2024-04-22
fix blue dot issue for SAM

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/1031.

Extensive tests were done via different web browsers, and finally it could be replicated: though it hardly (and randomly) happens. After tracking down the result, it is concluded that a function at the area selection module causes the issue. The area selection behavior itself partly relies on leaflet zoom box and the current logic is conflicted with map panning: in other words, when panning (clicking map and drag), the area selection initiates selection, which creates dot-size selection, and then it is cancelled immediately due to drag. It is not always and rare but it often leaves a dot-size selection in the map.

To solve this, a condition is set to check whether Control or Meta key is pressed so that area selection is not initiated when map panning.

### DIFF
--- a/packages/libs/components/src/map/AreaSelectHack.js
+++ b/packages/libs/components/src/map/AreaSelectHack.js
@@ -234,33 +234,38 @@ L.Map.SelectArea = L.Map.BoxZoom.extend({
    * @override
    */
   _onMouseMove: function (e) {
-    if (!this._moved) {
-      this._box = L.DomUtil.create('div', 'leaflet-zoom-box', this._pane);
-      L.DomUtil.setPosition(this._box, this._startLayerPoint);
-      this._map.fire(L.Map.SelectArea.AREA_SELECT_START);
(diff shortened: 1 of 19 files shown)


bobular:
Hi @moontrip I think this might have stopped rectangle selection of markers?

bobular (approved):
This looks like a solid solution.

If you're feeling creative - perhaps we should style the selection rectangle with yellow somehow? (So it looks different from the shift-zoom rectangle.) Feature creep, I know!

moontrip:
> This looks like a solid solution.
> 
> If you're feeling creative - perhaps we should style the selection rectangle with yellow somehow? (So it looks different from the shift-zoom rectangle.) Feature creep, I know!

@bobular Thank you for your suggestion. 👍 I made one for you 😉 https://github.com/VEuPathDB/web-monorepo/pull/1046

PR #1046, opened 2024-04-24
styling area selection box

moontrip:
This was branched out from https://github.com/VEuPathDB/web-monorepo/pull/1041. 

Yellow dot-line are less visible than blue dot-line, perhaps due to the background color of gray (with 50% opacity), but it may be okay? Tried green one, just out of curiosity, and it may have better visibility than yellow (screenshot): but this PR is for yellow. Perhaps you can try to change color at the CSS's border. 

If it is good enough, then I will merge it into the parent PR and then merge the parent PR which was already approved. 

area selection styling01

area selection styling03

### DIFF
--- a/packages/libs/components/src/map/styles/map-styles.css
+++ b/packages/libs/components/src/map/styles/map-styles.css
@@ -131,7 +131,6 @@
   height: 0;
   box-sizing: border-box;
   z-index: 800;
-  /* border: 2px dotted #38f; */
   border: 2px dotted rgba(255, 255, 0, 1);
   background: rgba(255, 255, 255, 0.5);
 }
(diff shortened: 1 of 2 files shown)


packages/libs/components/src/map/styles/map-styles.css
```diff
@@ -121,3 +121,16 @@
...
+/* area selection for selecting markers */
+.leaflet-area-selection-box {
+  position: absolute;
+  left: 0;
+  top: 0;
+  width: 0;
+  height: 0;
+  box-sizing: border-box;
+  z-index: 800;
+  border: 2px dotted rgba(255, 255, 0, 1);
```
bobular, line 134: I think it looks pretty cool with the same styling as the marker highlighting

```suggestion
  border: 6px solid rgba(255, 255, 0, 1);
```

bobular (approved):
This is great, thanks for allowing some mission creep!

What do you think about my suggestion? I've got nothing against the thin dotted line but prefer a bolder approach.

packages/libs/components/src/map/styles/map-styles.css
bobular: image

moontrip: @bobular Thank you for your suggestion 👍 Then I will comply with your suggestion and merge this into original PR

PR #1047, opened 2024-04-25
draw highlighted edges atop gray edges

asizemore:
I noticed that the gray were sitting above the higlighted edges. Turns out it was a very simple fix!

Voila!

https://github.com/VEuPathDB/web-monorepo/assets/11710234/87f4115d-dee6-4675-be92-ced8fee69115


dmfalke (approved):
Nice catch! I must have changed something in an earlier PR, after I had tested this. I swear it was working once!

PR #1053, opened 2024-04-26
Make low-cardinality variable legend order static

### LINKED ISSUE #279: Map - low-cardinality variables should not be sorted by frequency

### LINKED ISSUE #1035: BUG: standalone map legend colors should NOT change as the subset changes
on qa.restricted.clinepidb.org, using the PRISM Resistance surveillance study, the legend colors change as the filtered data changes.

If I filter "District" = Kanungu & Kabale and use donut markers to look at Plasmodium genotype / k13 Ala675Val, the legend looks like this:

image

If I filter "District" = Lamwo and use donut markers to look at Plasmodium genotype / k13 Ala675Val, the legend looks like this:

image

bobular:
Fixes #279 
Fixes #1035 

Low-cardinality variables are now presented in vocabulary order in the legend and this can only be affected by filters directly on the variable.

High-cardinality variables will continue to be sorted in descending frequency order and this order will be affected by ALL filters.


dmfalke (approved):
This looks good to me

PR #1057, opened 2024-04-26
Open treeBox branches to make visible default selection

jernestmyers:
Resolves https://redmine.apidb.org/issues/53705


packages/libs/coreui/src/components/inputs/checkboxes/CheckboxTree/CheckboxTree.tsx
```diff
@@ -155,7 +155,7 @@ export type CheckboxTreeProps<T> = {
   renderNode?: (node: T, path?: number[]) => React.ReactNode;
 
   /** List of expanded nodes as represented by their ids, default to null; if null, expandedList will be generated by the expandable tree. */
```
dmfalke, line 157: Oh my!

dmfalke (approved):
Nice catch!

PR #1060, opened 2024-04-29
Don't send little filters to `useDistributionOverlayConfig` and `useLegendData`

### LINKED ISSUE #896: Map - continuous markers are not matching the legend when timeslider is in use
To reproduce:
Megastudy

Bar markers - variable = collection start date
Binning method: **quantile** (makes it easier to debug, but the bug seems to be unrelated to the binning method)

image

Note that the rightmost two bins should be 2017 ish to 2022

If we select that region in the timeslider we should only get barcharts that look like this `________||` but we don't. The wrong "overlayConfig" is being sent to the back end map-markers request it seems.  The legend remaining constant is the desired behaviour.

image

bobular:
Fixes #896 

The problem for both donut/bar and bubble markers was that `useMarkerData` was calling either `useDistributionOverlayConfig` or `useLegendData` internally using the `filters` that was passed to it. Problem was that we were sending filters plus "little filters for marker data" (aka viewport and timeslider).  This meant that the `overlayConfig` (or `legendData`) needed for the marker requests (and/or processing) was not the same as has been used for the legend (which didn't use any little filters).

As explained in #896 this was most obvious when showing the timeslider variable on the markers. The legend and markers were telling completely different stories.

Now the code is slightly more verbose because `useDistributionOverlayConfig` or `useLegendData` are called outside `useMarkerData` and the result is passed into it.

The alternative was to change the `useMarkerData` props so there was an additional `allFilters` prop (or similar). I thought this approach was cleaner in the long run.

### DIFF
--- a/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BarMarkerMapType.tsx
+++ b/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/BarMarkerMapType.tsx
@@ -161,9 +161,17 @@ function ConfigPanelComponent(props: MapTypeConfigPanelProps) {
     enabled: configuration.selectedCountsOption === 'visible',
   });
 
-  const previewMarkerData = useMarkerData({
+  const overlayConfigQueryResult = useDistributionOverlayConfig({
     studyId,
     filters,
+    binningMethod,
+    overlayVariableDescriptor: selectedVariable,
+    selectedValues,
+  });
+
+  const previewMarkerData = useMarkerData({
+    studyId,
+    filters, // no extra little filters; should reflect whole map and all time
     studyEntities,
     geoConfigs,
     selectedVariable,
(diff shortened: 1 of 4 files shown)


moontrip (approved):
After testing extensively I found no side effects. 👍 I was examining this issue from this morning without knowing that Bob was actually working on it 😅 But my investigation actually helped me understand this PR faster, so it is not that bad 😉

bobular:
> I want to make sure I understand the issue. Before, the legend and the markers were using different overlay configs, causing the color mapping to be different between the two. That is because we do not want little filters to impact the legends.
> 
> For what it's worth, I think your changes are good. I do prefer passing in the overlay config--it's a good example of inversion of control and separation of concerns.
> 
> I didn't test, but can shortly.

Well it felt right at the time! DK has kindly done extensive testing after I worked on it without assigning. oops.

Issue #1054, opened 2024-04-26
Tweak marker selection snackbar messages so they are more suitable for filter-related scenarios.


bobular:
I just noticed that the snackbar messages come up as a result of filtering. This was an inevitability of my simple approach - oops!

Example 1.

* Zoom in to Southern Africa + Madagascar
* Select markers in Madagascar then add a filter for country=Zimbabwe
* The `Selected markers that went off-screen have been deselected` message appears.

This could be ameliorated with some minor verbiage changes. The code that produces the message doesn't know about filters and subsets AFAIK.

Example 2.

* Select some markers
* Add a filter that removes all markers from the screen (e.g. country = Poland)
* The `Marker selection has been cancelled because aggregation level has changed due to zooming` message appears.

This is coming up because `consolidatedMarkers.length` is zero.
This can be fixed with a new message for when there are no markers on the screen any more.

_Originally posted by @bobular in https://github.com/VEuPathDB/web-monorepo/issues/617#issuecomment-2079398286_

bobular:
The new message could be `Marker selection has been cancelled because there are no visible markers in the current view`

bobular:
The tweaked message could be `Selected markers that are no longer visible have been deselected`

PR #1061, opened 2024-04-29
tweak marker selection snackbar messages

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/1054.


bobular (approved):
Works great. Thanks!

PR #1065, opened 2024-04-30
added missing dep

### LINKED ISSUE #1064: Map - constraints violation allowed by front end with boxplot in megastudy.
I will fill in more details tomorrow but the crux of this (seen in latest `main`, on qa and production) is as follows

Start with a boxplot of mortality rate overlaid by insecticide:

image

Then change the marker variable (which is also the overlay variable for floaters) to something on a different branch, e.g. "Plasmodium falciparum infection status"

It should just be disabled with a nice message saying why (and this happens with a histogram of mortality rate). But instead there's a 400 

image

Bizarrely, switching back to insecticide gives you the warning that we should have had after switching to Pf infection status!
Need to do more tests and look at the constraints provided by the back end.

Here's the analysis shared from production

https://vectorbase.org/vectorbase/app/workspace/maps/n1k6Rqg/import

bobular:
Fixes #1064 

@dmfalke was spot on about the missing dependency!


dmfalke (approved):
🎉

PR #1073, opened 2024-05-03
Disable histogram timeline for b68 release

bobular:
Fixes https://github.com/VEuPathDB/EdaNewIssues/issues/784

Is it worth changing `StartPage` to either remove or improve the greyed out icon?

NO. There is a back end fix to go with this. https://github.com/VEuPathDB/service-eda/pull/31 

image


dmfalke:
With https://github.com/VEuPathDB/service-eda/pull/31, I don't think we need this PR. I'll test and confirm before Monday

dmfalke:
> With [VEuPathDB/service-eda#31](https://github.com/VEuPathDB/service-eda/pull/31), I don't think we need this PR. I'll test and confirm before Monday

Confirmed. I think we should close this PR without merging it.

bobular:
Closing without merge. Thanks for testing that.

PR #1079, opened 2024-05-08
Fix search links for WGCNA network nodes

### LINKED ISSUE #1078: WGCNA Genomics EDA - Nodes are linking to qa sites in prod
This bit of code needs to be updated:

https://github.com/VEuPathDB/web-monorepo/blob/d7c2e91ecf9bc26a6c52b830c433a9b8c7e9d830/packages/libs/eda/src/lib/core/components/computations/plugins/correlation.tsx#L102-L108

dmfalke:
Closes #1078

This PR makes it so the WDK search links in WGCNA EDA correlation viz use the correct url prefix. The url prefix is derived from the host website (not to be confused with HostDB website). This means the links will likely not work on dev sites, unless the same dev site prefix exists for both plasmodb and hostdb.

In the future, this could be driven by configuration values.

_**The production site needs to be patched with this change.**_

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/correlation.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/correlation.tsx
@@ -99,12 +99,18 @@ export const plugin: ComputationPlugin = {
           const variable = variables.find((v) => v.id === variableId);
           if (variable == null) return [];
 
+          // E.g., "qa."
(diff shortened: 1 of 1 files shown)


asizemore (commented):
Why did we remove the https: part? Otherwise wonderful thank you for getting this done so quickly!

dmfalke:
I removed the protocol so that it matches the protocol used by the website the user is on.

PR #1082, opened 2024-05-09
Fix regression in RecordTable search filter

jernestmyers:
Resolves https://epvb.slack.com/archives/C01DS6HBY3Z/p1715270392560699


dmfalke (approved):
Looks good!

On a side note, I only just realized that the changes related to the orthologs table are in wdk-client. As a follow-up, it would be good to move these changes to web-common. Sorry I didn't notice before!

PR #1088, opened 2024-05-13
Use a Set for id lookup

dmfalke:
This PR is meant to address some inefficiencies when a study has >120,000 variables.

Changes include:
- Use a `Set` for lookups, rather than `Array.prototype.includes`

_For now, we will only support 30K variables._


asizemore:
Sweet! How would you like me to test? Just general testing to make sure it works on other datasets? Or would you like me to test it on the super big dataset? Let me know what would be most helpful!

dmfalke:
I think the main thing is to test that the test being performed is still correct

asizemore (approved):
lgtm!

Issue #1101, opened 2024-05-20
Barplot (EDA and Map) is guessing numerical wrongly for an edge case


bobular:
We have a user dataset (I will post it on the crows Slack if anyone needs it) via @nkittur-uga that has a variable with the values: 1, 4, 5, 66, "atypical/unique".

UD loading calls it correctly a string categorical

Boxplots look fine:

![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/31a6413a-7aba-4f0a-83aa-32f50704db08)

But barplots are treating it as a number for some reason. Probably because it's guessing from the content and we never told it specifically to use a categorical axis.

![image](https://github.com/VEuPathDB/web-monorepo/assets/308639/7f31fe25-f948-467a-81ee-833f36a0abaf)

Note that the "atypical/unique"s are lost completely from the plot. 😢

PR #1102, opened 2024-05-22
possibly address barplot x-axis with a mix of number strings and text strings

moontrip:
This will possibly address https://github.com/VEuPathDB/web-monorepo/issues/1101. I have looked at the current implementation of Barplot Viz but couldn't find specific issue why it was caused. Then, tested with some mock-up data that consists of a mixture of number strings and text strings like the screenshot, Surprisingly, it seems like bar chart at Plotly treats those specific case as number strings only, by excluding text strings. 

That being said, I found an option that the x-axis (technically independent axis) can always be treated as categorical values, by setting type: 'category'. From my tests (using codesandbox.io), it accepts such mixed strings correctly. 

So, @bobular it would be great if you can test this PR with the user dataset you mentioned.

### DIFF
--- a/packages/libs/components/src/plots/Barplot.tsx
+++ b/packages/libs/components/src/plots/Barplot.tsx
@@ -168,6 +168,7 @@ const Barplot = makePlotlyPlotComponent(
       range: data.series.length ? undefined : [0, 10],
       tickfont: data.series.length ? {} : { color: 'transparent' },
       showticklabels: showIndependentAxisTickLabel,
+      type: 'category',
     };
 
     // truncation


bobular (approved):
Looks good! I was unable to test with Nupur's user dataset so I spoofed some data in dev tools. A barplot which had x-axis values [0,1,2,3,4] to [0,1,2,3,49]

A. without your fix
image
B. with your fix
image

moontrip:
@bobular Thank you for your tests and review 👍 Glad it works! Will merge this

PR #1110, opened 2024-06-11
Add more control over npm tags

bobular:
git tag `v1.2.3` -> npm tag `latest`
git tag `v1.2.3-someword` -> npm tag `someword`
git tag `v1.2.3.anotherword.3` -> npm tag `anotherword`

falls back to `latest` if the pattern matching fails for whatever reason.

### DIFF
--- a/.github/workflows/npm-publish-sites.yml
+++ b/.github/workflows/npm-publish-sites.yml
@@ -39,13 +39,30 @@ jobs:
         env:
           NODE_OPTIONS: '--max-old-space-size=4096'
       - run: node tools/scripts/version.mjs packages/sites/${{ matrix.dir }} ${GITHUB_REF_NAME#v}
+      - id: extract-npm-tag
+        # figure out how to tag the release on npm - sets env.npm_tag
+        run: |
+          # Extract the part of the git tag after the 'v' prefix
+          git_tag="${GITHUB_REF_NAME#v}"
+          # Check if the tag is a plain version (e.g., 1.2.3) and tag 'latest' on npm
+          if [[ "$git_tag" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "npm_tag=latest" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish-sites.yml
```diff
@@ -39,13 +39,29 @@ jobs:
...
+            echo "npm_tag=latest" >> $GITHUB_ENV
+          # else look for a word-like suffix (ignoring optional trailing .1, .2 etc) and tag with that word
+          # e.g. v.1.2.3-experimental will be tagged 'experimental' on npm
+          elif [[ "$git_tag" =~ ^[0-9]+\.[0-9]+\.[0-9]+-([a-zA-Z0-9]+)(\.[0-9]+)*$ ]]; then
+            npm_tag="${BASH_REMATCH[1]}"
+            echo "npm_tag=$npm_tag" >> $GITHUB_ENV
+          # or fall back to 'latest' just in case
+          else
+            echo "npm_tag=latest" >> $GITHUB_ENV
+          fi
```
dmfalke, line 57: We can check the `prerelease` property of the `release` object. If it's `true`, we can use `prerelease`, otherwise we can use `latest`?

```suggestion
          # if prerelease is checked, use "prerelease" for the npm tag
          elif ${{ github.event.release.prerelease }}; then
            echo "npm_tag=prerelease" >> $GITHUB_ENV
          # or fall back to 'latest' just in case
          else
            echo "npm_tag=latest" >> $GITHUB_ENV
          fi
```

dmfalke (commented):
Looking good! I will take a closer look at the regexp. In the mean time, see my suggestion and let me know what you think.

.github/workflows/npm-publish-sites.yml
bobular: Great idea!

dmfalke (approved):
Sorry for the delay. I think the regex looks good. :rocket:

PR #1111, opened 2024-06-11
Wdk service cache fix

dmfalke:
This PR fixes an issue where the reload dialog would appear on an initial page load.

The issue is due to some requests being made before the WdkService `startupTime` property is initialized (as the variable `_version`) on the client. Such requests would set `_hasRequestBeenMade` to `true`, but the request would not include the header `x-client-wdk-timestamp`.

The fix is to always call `_initializeStore` before any requests are made. This took a little refactoring to avoid an infinite loop.

### DIFF
--- a/packages/libs/wdk-client/src/Service/ServiceBase.ts
+++ b/packages/libs/wdk-client/src/Service/ServiceBase.ts
@@ -351,7 +351,13 @@ export const ServiceBase = (serviceUrl: string) => {
       .then((storeConfig) => {
         if (storeConfig == null) {
           return fetch(serviceUrl)
-            .then((response) => response.json())
+            .then((response) => {
+              if (!response.ok) {
+                console.error(`Fetching ${serviceUrl} failed for _initializeStore: ${response.statusText}`);
+                throw new Error("Failed to initialize service");
+              }
+              return response.json()
+            })
             .then((serviceConfig: ServiceConfig) => {
               return _store
                 .setItem('/__config', serviceConfig)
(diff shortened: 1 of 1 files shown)


packages/libs/wdk-client/src/Service/ServiceBase.ts
```diff
@@ -350,14 +350,19 @@ export const ServiceBase = (serviceUrl: string) => {
       .getItem<ServiceConfig>('/__config')
       .then((storeConfig) => {
         if (storeConfig == null) {
-          return _fetchJson<ServiceConfig>('GET', '/').then((serviceConfig) => {
-            return _store.setItem('/__config', serviceConfig);
-          });
+          return fetch(serviceUrl)
+            .then((response) => response.json())
```
bobular, line 354: I think we should catch/throw a failed request

```suggestion
            .then((response) => {
              if (!response.ok) {
                console.error(`Fetching ${serviceUrl} failed for _initializeStore: ${response.statusText}`);
                throw new Error("Failed to initialize service");
              }
              return response.json()
            })
```

bobular (commented):
This looks good. I wondered if the vanilla `fetch` could/should be a `fetchWithRetry` to be consistent?

packages/libs/wdk-client/src/Service/ServiceBase.ts
dmfalke: Great idea!

dmfalke:
> This looks good. I wondered if the vanilla `fetch` could/should be a `fetchWithRetry` to be consistent?

I don't have a good reason not to do that, so I will make that change!

dmfalke:
@bobular I made the changes you suggested. Let me know what you think!

dmfalke:
> I approve fully!

:pray:

Issue #1108, opened 2024-06-05
Network: Add show label toggle and search box to show specific node labels


asizemore:
Underneath the viz, in a new Network Controls section, let's add a toggle and searchable multi-select that lets a user turn on/off node labels, and specify which node labels they'd like with the multi-select. Default all node labels.

<img width="616" alt="Screen Shot 2024-06-05 at 11 07 06 AM" src="https://github.com/VEuPathDB/web-monorepo/assets/11710234/f87d56f2-13a6-4f08-b767-c8a8731e567f">

Issue #1105, opened 2024-06-05
Network: Basic node label improvements


asizemore:
For the network viz, our node labels are a bit all over the place. 

<img width="950" alt="Screen Shot 2024-06-05 at 10 48 36 AM" src="https://github.com/VEuPathDB/web-monorepo/assets/11710234/c3785b45-387a-44ef-948b-880d08d125f0">

Let's do the following to improve node labels:
1. Hide node labels and only show on hover
2. Have node labels show atop everything else. Like tooltips?

Can branch off of #1090 until that work is merged

PR #1113, opened 2024-06-12
Network node label control

moontrip:
This will possibly address https://github.com/VEuPathDB/web-monorepo/issues/1105 and https://github.com/VEuPathDB/web-monorepo/issues/1108. However, please note the followings:

- This is branched out from https://github.com/VEuPathDB/web-monorepo/pull/1090
- Currently, it is tested with a story/storybook, not with actual network, thus this PR is set to be a draft one for now
- No. 1 requirement at https://github.com/VEuPathDB/web-monorepo/issues/1105 is made
- Concering No. 2 requrement at https://github.com/VEuPathDB/web-monorepo/issues/1105, I tried several ways to overcome it but could not find a solution. It is because of the nature of the nature of the SVG where a CSS like z-index does not work and later SVG element always overlaps former one. Thus, there is no clear solution for that, I suppose: perhaps one needs to rely on tooltip if a node label is not that visible due to overlap?
- As for https://github.com/VEuPathDB/web-monorepo/issues/1108, I made a mockup to follow the description and I think it works fine. 
- made a video capture for a quick review in the following

If this is okay, then I will try to implement this to actual network VIz

https://github.com/VEuPathDB/web-monorepo/assets/12802305/95a1bcd3-26eb-4bb1-ae39-651c69319c7c


moontrip:
@asizemore New node label control is also applied to Network Viz. Below is an example screenshot and video capture for Correlation (Taxa v. Taxa) Viz: 

network example


https://github.com/VEuPathDB/web-monorepo/assets/12802305/3fa53bc0-a06a-4419-9c87-124de20dab23

asizemore:
@moontrip this is looking awesome!! Thanks for getting this all together!! I'm back working on the `feature-1038-network-viz` now to hopefully remind myself a bit of what's going on, then i'll be back here to take a look!

packages/libs/components/src/stories/plots/NetworkPlot.stories.tsx
```diff
@@ -171,6 +177,120 @@ WithActions.args = {
...
+  // all values are selected as default
+  const [selectedNodeLabels, setSelectedNodeLabels] =
+    useState<(string | undefined)[]>(labels);
+  // show/hide node labels button
+  const [showNodeLabels, setShowNodeLabels] = useState<boolean>(false);
+
+  const ref = useRef<any>(null);
+
+  // custom switch based on MUI
+  const AntSwitch = withStyles((theme: Theme) =>
```
asizemore, line 206: If I remmeber correctly, i think coreui has a custom toggle/switch. Is there a reason not to use that one here?
(I do like the way this looks btw, just curious)

packages/libs/components/src/stories/plots/NetworkPlot.stories.tsx
```diff
@@ -171,6 +177,120 @@ WithActions.args = {
...
+      },
+      checked: {},
+    })
+  )(Switch);
+
+  return (
+    <>
+      <NetworkPlot
+        containerStyles={{ width: DEFAULT_PLOT_SIZE }}
+        selectedNodeLabels={selectedNodeLabels}
```
asizemore, line 246: i think it may have been me who suggested "selectedNodeLabels", but i just realized we already use "selected" to mean something different (when a user has clicked on a node). So instead let's change this prop name to `visibleNodeLabels` or something like that.

packages/libs/components/src/components/plotControls/MultiSelection.tsx
```diff
@@ -0,0 +1,80 @@
...
+import Input from '@material-ui/core/Input';
+import InputLabel from '@material-ui/core/InputLabel';
+import MenuItem from '@material-ui/core/MenuItem';
+import FormControl from '@material-ui/core/FormControl';
+import ListItemText from '@material-ui/core/ListItemText';
+import Select from '@material-ui/core/Select';
+import Checkbox from '@material-ui/core/Checkbox';
+// import Chip from '@material-ui/core/Chip';
+
+interface MultiSelectionProps {
```
asizemore, line 12: This looks very nice! We expect to have thousands of nodes so i think it'll be very crucial to have a search function in the select box as well. After looking around a bit i found https://react-select.com/home which has a searchable, multi-select component. Do you think that would work fine here? I think it has all the functionality we need!

packages/libs/components/src/plots/NetworkPlot.tsx
```diff
@@ -76,6 +80,8 @@ function NetworkPlot(props: NetworkPlotProps, ref: Ref<HTMLDivElement>) {
     labelTruncationLength = 20,
     emptyNetworkContent,
     annotations,
+    selectedNodeLabels,
+    showNodeLabels = false,
```
asizemore, line 84: originally i thought `false` would be the best default. But after playing around with it a bit i was wrong! Let's do `true` as the default please!

packages/libs/components/src/plots/Node.tsx
```diff
@@ -60,28 +67,48 @@ export function NodeWithLabel(props: NodeWithLabelProps) {
...
+          verticalAnchor="middle"
+          fontWeight={fontWeight}
+          fill={labelColor}
+          id="NodeLabelText"
+          style={{
+            cursor: 'ponter',
+            zIndex: 1000,
+            backgroundColor: 'red',
+            display:
+              label && showNodeLabels && selectedNodeLabels.includes(label)
```
asizemore, line 100: i think we can move this logic outside of the component and then simplify the props. Each `NodeWithLabel` operates on one node, so we could just do one prop called `showLabel` which is true if `showNodeLabels && selectedNodeLabels.includes(label)` is true and false otherwise. What do you think about that?

packages/libs/components/src/plots/Node.tsx
```diff
@@ -60,28 +67,48 @@ export function NodeWithLabel(props: NodeWithLabelProps) {
...
+                : hover
+                ? 'block'
+                : 'none',
+          }}
+        >
+          {label && truncateWithEllipsis(label, truncationLength)}
+        </Text>
+        <title>{label}</title>
+      </g>
+      {/* <use xlinkHref=".NodeLables" /> */}
```
asizemore, line 111: is this code still needed?

packages/libs/components/src/stories/plots/NetworkPlot.stories.tsx
moontrip: 👋 @asizemore Yes there exists a CoreUI's toggle/switch. But if I understand cocrrectly, it does not support labels at both sides (left and right of a toggle/switch UI), like **Hide** <toggle/switch UI> **Show**, which was suggested at the screenshot of  https://github.com/VEuPathDB/web-monorepo/issues/1108. That was the reason why I had to introduce a new switch based on Material-UI. Please correct me if I am incorrect.

packages/libs/eda/src/lib/core/components/visualizations/implementations/NetworkVisualization.tsx
```diff
@@ -247,6 +263,18 @@ function NetworkViz(props: VisualizationProps<Options>) {
       };
     });
 
+    // update selected node label
+    const labels =
```
asizemore, line 267: can we change this to `nodeLabels` just for clarity, please?

packages/libs/eda/src/lib/core/components/visualizations/implementations/NetworkVisualization.tsx
```diff
@@ -307,6 +335,17 @@ function NetworkViz(props: VisualizationProps<Options>) {
     </div>
   );
 
+  // list of node labels
+  const labels = useMemo(() => {
```
asizemore, line 339: i dont follow - why is `labels` defined twice? Once here and once above?

asizemore:
Looking good! I know i've been gone a bit and a lot has changed with the project since these issues got created, so please bear with me as we might change course a bit :) 

Overall this is really exciting! I have some questions i left as comments. Otherwise two larger questions:
1. Can we make the label controls more like the below where the styling mimics the other plots please? So for example instead of "Plot mode" in the bar plot we'd have "Network controls"? 
Screen Shot 2024-06-21 at 12 40 31 PM
2. Looking over the mockups i also realized we don't actually need the toggle, as long as the multiselect has a "show all | clear all" option! What do you think about the feasibility of this with the `react-select` multiselect component? Here's a [code sandbox with an example](https://codesandbox.io/s/react-multi-select-example-with-select-all-option-and-checkboxes-ejjc9) of what i mean!
3. I had trouble with the network story working. I can toggle on/off the labels in the plot but in the select the checkbox stays on? Not sure what's going on but here's a screencast:

https://github.com/VEuPathDB/web-monorepo/assets/11710234/539a2ded-d867-4d72-a0ab-7ff87f5dfa51

moontrip:
@asizemore Thank you for your feedbacks. I will take a look at them carefully next week, and try to accept those valuable suggestions 👍 By the way, perhaps some functionality of the storybook was broken while fixing some bugs for network Viz. I will also look into it.

moontrip:
@asizemore your suggestions are mostly addressed. I will leave my response at each of your question. Also this branch is synced with its parent branch and resolved merge conflicts. Here is a video capture with the same example study.

https://github.com/VEuPathDB/web-monorepo/assets/12802305/e408a865-4bb3-4648-a152-03d6fa2d5555

packages/libs/components/src/stories/plots/NetworkPlot.stories.tsx
moontrip: @asizemore with the new commits, MUI one is replaced with React-Select so MUI-related ones are removed.

packages/libs/components/src/stories/plots/NetworkPlot.stories.tsx
moontrip: @asizemore changed to visibleNodeLabels as suggested

packages/libs/components/src/components/plotControls/MultiSelection.tsx
moontrip: @asizemore as mentioned at other place, now the node label control uses React-Select, not MUI

packages/libs/components/src/plots/NetworkPlot.tsx
moontrip: @asizemore as the toggle switch is removed, showNodeLabels prop is also removed

packages/libs/components/src/plots/Node.tsx
moontrip: @asizemore as suggested, it is now passed as a prop from the parent component, `NetworkPlot`

packages/libs/components/src/plots/Node.tsx
moontrip: @asizemore Oops, removed 😅

packages/libs/eda/src/lib/core/components/visualizations/implementations/NetworkVisualization.tsx
moontrip: @asizemore changed to `nodeLabels` as suggested

packages/libs/eda/src/lib/core/components/visualizations/implementations/NetworkVisualization.tsx
moontrip: @asizemore Yes there are two `labels` variables, which are now changed to `nodeLabels` and they have different roles. The first one is defined inside `cleanedData` to set the default of the node label control to be "Select All" (all node labels are checked initially, which mimics "show all node labels in the network plot": since `cleanedData` is defined using useMemo hook with dependencies, it will be more efficient to set the default "Select All" in there. The second one is to pass the list of all node labels to node label control so that all node labels are shown in the drop down menu with checkboxes.

asizemore:
WOW it looks so great! I'll dig into the code later this week! Thanks so much @moontrip !!!

moontrip:
> WOW it looks so great! I'll dig into the code later this week! Thanks so much @moontrip !!!

@asizemore Glad you like it 😄 Please feel free to make suggestion(s) if necessary

p.s. Hmm... I didn't know that github changed emojis: personally I liked the previous ones better 😞

packages/libs/eda/src/lib/core/components/visualizations/implementations/NetworkVisualization.tsx
```diff
@@ -0,0 +1,561 @@
...
+        labelPosition:
+          scaleX(Number(node.x)) >
+          (plotContainerStyleOverrides?.width ?? 900) / 2
+            ? 'right'
+            : ('left' as LabelPosition),
+      };
+    });
+
+    // set initial visible node labels
+    const nodeLabels = nodesWithLabels.flatMap((node) => {
```
asizemore, line 271: For clarification, let's call this one `defaultNodeLabels` or `initialNodeLabels`. What do you think? I'm hoping to make it easier for someone searching for "nodeLabels" in the future :)

packages/libs/eda/src/lib/core/components/visualizations/implementations/NetworkVisualization.tsx
asizemore: thanks @moontrip ! That's very helpful. I made one further suggestion in the latest review about it.

asizemore (commented):
This looks really nice! I tested with the qa backend for mbio and it was very smooth. I really like the new selector and it all just works very well. I had two tiny comments/questions but that's it!

moontrip:
> This looks really nice! I tested with the qa backend for mbio and it was very smooth. I really like the new selector and it all just works very well. I had two tiny comments/questions but that's it!

@asizemore Thank you for your tests and reviews. 👍 I will address your feedbacks

packages/libs/components/src/plots/Node.tsx
```diff
@@ -60,28 +64,42 @@ export function NodeWithLabel(props: NodeWithLabelProps) {
...
+          textAnchor={textAnchor}
+          fontSize={fontSize}
+          verticalAnchor="middle"
+          fontWeight={fontWeight}
+          fill={labelColor}
+          id="NodeLabelText"
+          style={{
+            cursor: 'ponter',
+            zIndex: 1000,
+            backgroundColor: 'red',
```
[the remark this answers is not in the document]
moontrip: @asizemore Oops I think it was a part of some tests but forgot to delete it. Will remove.

packages/libs/eda/src/lib/core/components/visualizations/implementations/NetworkVisualization.tsx
moontrip: @asizemore sounds good 👍 Will do that.

moontrip:
@asizemore Feedbacks are addressed and merge conflicts are also resolved :)

asizemore (approved):
Woot this is awesome! Thanks @moontrip !

moontrip:
> Woot this is awesome! Thanks @moontrip !

@asizemore Thank you for your tests, reviews, and suggestions 👍

PR #1114, opened 2024-06-18
Add ability to make user datasets public (aka, community datasets)

dmfalke:
### Summary

This PR adds the ability to make a user dataset public. This is done by setting the new `visibility` property introduces by VDI.

_This feature is only visible on clinepi and mbio sites_.

#### Toggle in sharing modal
image

#### New "Community" column in User Datasets table
image

#### New header menu section
image

#### New section on Study Summaries page
image

### DIFF
--- a/packages/libs/user-datasets/src/lib/Components/Sharing/UserDatasetSharingModal.jsx
+++ b/packages/libs/user-datasets/src/lib/Components/Sharing/UserDatasetSharingModal.jsx
@@ -469,6 +469,15 @@ class UserDatasetSharingModal extends React.Component {
         </div>
       );
     } else {
+      // Determine the value for the community visibility toggle:
+      // If a single user dataset is selected, use its value.
+      // If multiple are selected, and they all have the same value,
+      // use that value; otherwise, use undefined.
+      //
+      // Since the former is a subset of the latter, we can implement
+      // only the latter by:
+      // 1. Get an array of unique visbility values.
+      // 2. If the length of the array is 1, use that value; otherwise use undefined.
       const visibilities = uniq(datasets.map((d) => d.meta.visibility));
       const visibility =
         visibilities.length === 1 ? visibilities[0] : undefined;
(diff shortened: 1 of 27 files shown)


packages/libs/user-datasets/src/lib/Components/Sharing/UserDatasetSharingModal.jsx
```diff
@@ -460,8 +469,35 @@ class UserDatasetSharingModal extends React.Component {
         </div>
       );
     } else {
+      const visibilities = uniq(datasets.map((d) => d.meta.visibility));
+      const visibility =
+        visibilities.length === 1 ? visibilities[0] : undefined;
```
jernestmyers, line 474: Nit: add a comment here?

This is clever, but it took me a bit of effort to understand the how and why of `visibilities` and `visibility`, and how it works to determine the value of the toggle.

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -607,6 +607,7 @@ function validateForm<T extends string = string>(
       datasetType: datasetUploadType.type,
       projects: [projectId],
       dataUploadSelection,
+      visibility: 'private',
```
jernestmyers, line 610: Are we setting this as `private` during upload for simplicity, rather than allowing the user (in clinepi and mbio) to set visibility during the upload phase?

jernestmyers (approved):
These changes look great. Just a couple non-blocking comments.

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
dmfalke: It's partially for simplicity, and partially because I think people should review their user dataset before setting it to public.

packages/libs/user-datasets/src/lib/Components/Sharing/UserDatasetSharingModal.jsx
dmfalke: Ah, yes, I will add a comment. Thanks!

dmfalke: Comment added. Let me know what you think.

PR #1115, opened 2024-06-21
New optional Mesa inline table behaviour using tooltips

bobular:
Note this targets the ongoing orthogroup branch.

The default Mesa behaviour when `options.inline` is `true`, is for cell contents to be truncated so the rows are equal height. Clicking on the rows expands them to show the full contents.

This wasn't acceptable for ortho, where the tree and table alignment is messed up by expanding rows. (Also browser scroll-to behaviour is unpredictable and you can't easily find the row you expanded.)

With this PR there's a new configuration option: `inlineUseTooltips` which removes the click behaviour and adds tooltips to all table cells.

It would be nice to do this for
a) only cells with content
b) only cells that overflow

but neither are economical to implement at this time.

### DIFF
--- a/packages/libs/components/src/components/tidytree/TreeTable.tsx
+++ b/packages/libs/components/src/components/tidytree/TreeTable.tsx
@@ -81,6 +81,7 @@ export default function TreeTable<RowType>(props: TreeTableProps<RowType>) {
       ...props.tableProps.options,
       deriveRowClassName: (_) => rowStyleClassName,
       inline: true,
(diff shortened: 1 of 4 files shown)


bobular:
I'm using it here!
It was a useful way to get fixed-height rows. Until someone clicked on them of course...

bobular:
I'm going to merge to `orthogroup-tree-table`. Going to be off grid for a few days and it would be nice to get this on the feature site.

Issue #1094, opened 2024-05-14
EDA - Featured variables styling


dmfalke:
From David Roos:

> Might be easier to visually parse Featured Variables if entity is in a separate tab, or highlighted in bold (see mockups for LLINEUP2). Option 1 would be easier to read, and also (presumably) very easy to implement. Option 2 takes more vertical space. Option 3 takes more horizontal space. Option 4 would probably be harder to implement properly
>
> ![image](https://github.com/VEuPathDB/web-monorepo/assets/365139/e22b97bb-0b3c-4f2a-95b2-548789a75a62)

bobular:
Proposal 2 agreed with five votes at a UX meeting as reported by @aurreco-uga

PR #1120, opened 2024-06-25
1094 reorganize featured variables

chowington:
Resolves https://github.com/VEuPathDB/web-monorepo/issues/1094

New version in context:

image

### DIFF
--- a/packages/libs/eda/src/lib/core/components/variableSelectors/VariableList.tsx
+++ b/packages/libs/eda/src/lib/core/components/variableSelectors/VariableList.tsx
@@ -694,13 +694,7 @@ export default function VariableList({
             {Object.entries(groupedFeaturedFields).map(
               ([entityName, fields]) => (
                 <li>
-                  <h4
-                    style={{
-                      fontSize: '1.05em',
-                      marginLeft: '1em',
-                      padding: 0,
-                    }}
-                  >
+                  <h4 style={{ fontSize: '1.05em', padding: 0 }}>
                     {entityName}
                   </h4>
                   <ul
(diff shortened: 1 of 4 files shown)


dmfalke (commented):
The code looks good, but I'm not able to test since the data isn't currently available. I will try again over the weekend or on Monday.

bobular:
I though because qa.clinepidb.org was up (qa.vectorbase.org has been down for a few days), I'd be able to test, but it seems that no studies are available on the qa back end :-(

dmfalke:
I was able to play with this a little, using the prod site as a backend. I think it might look a little better if the entity names were indented to the same level as the variable names... and then indent the variable names by another space. I would also use semantic tags for the entity names and replace the `<span>` with `<h4>`. It will require some additional styling to make it work in this context.

bobular:
Looking at this now! I know it's out of the remit of this ticket but we have some serious entity diagram label overflow going on.

image

dmfalke:
> Looking at this now! I know it's out of the remit of this ticket but we have some serious entity diagram label overflow going on.
> 
> image

Perhaps we can make the width of the boxes based on the longest entity name, on a per-study basis.

bobular (approved):
All the nuts and bolts look good.
As for styling, I don't think indenting more is a great use of space. I'd probably remove the :small_red_triangle_down: and show/hide from the "Featured variables" heading (but make that stronger?)

Probably not best to rely on my design skills though!

As for the semantic headings, maybe it should be nested `<ul>`s like the main variable tree?

Either way, I'd be happy for this to go out!

dmfalke:
> As for the semantic headings, maybe it should be nested `<ul>`s like the main variable tree?

That's fine, too. I'm also okay with keep the indentation as-is.

PR #1121, opened 2024-06-26
Userdatasets rnaseq fixes

### LINKED ISSUE #1117: Genomics UD RNA-Seq: txt files do not need link to JBrowse
Screenshot 2024-06-23 at 3 17 20 PM

dmfalke:
Fixes #1117

This PR addresses two issues with RNA-Seq datasets:
1. Prevents a WDK search request when dependencies are not yet installed.
2. Prevents non-bigwig files from appearing in the **Genome Browser Tracks** table

When addressing the first issues, I noticed that it's feasible for there to be multiple genomic dependencies, but the code does not account for it. It will always use the last one in the array. It's possible that this will never happen in the real world, but it might be worth asking about. I will post a question on slack.

### DIFF
--- a/packages/libs/user-datasets/src/lib/Components/Detail/BigwigDatasetDetail.jsx
+++ b/packages/libs/user-datasets/src/lib/Components/Detail/BigwigDatasetDetail.jsx
@@ -25,6 +25,7 @@ class BigwigDatasetDetail extends UserDatasetDetail {
     const { wdkService } = this.context;
     const { dependencies } = userDataset;
     let genome;
+    // FIXME Can a UD have multiple of these dependencies?
     dependencies.forEach(function (dependency) {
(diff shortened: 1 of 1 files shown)


jernestmyers (approved):

dmfalke:
> When addressing the first issues, I noticed that it's feasible for there to be multiple genomic dependencies, but the code does not account for it. It will always use the last one in the array. It's possible that this will never happen in the real world, but it might be worth asking about. I will post a question on slack.

I confirmed on slack that there will only ever be one entry for `dependencies`. I will add a comment to the code, for posterity.

dmfalke:
I updated the comment. I also added a check to make sure the dataset is compatible with the current project (component site). Otherwise, if the user is on a different project than the one the dataset is installed, the search will return an empty result set.

PR #1124, opened 2024-07-01
Display invalid filters so they can be removed

dmfalke:
This PR renders EDA filters with invalid variable references, so that they can be removed from an analysis. A tooltip in included, explaining why the filter is invalid.

image


aurreco-uga (approved):
I think I would switch the order of sentences.

Issue #946, opened 2024-03-11
Export plot button does not export the legend


danicahelb:
When you export a plot with an overlay, the legend should be included in the exported plot. Currently it is not, making the exported plot useless

jernestmyers:
This came up in our `ui-infra` scrum and we decided this needs further discussion.

bobular:
@steve-fischer-200 and @bobular suggest that ONE file is likely too much work

user will be editing the SVG anyway

So we provide two files: plot and legend

This assumes we can get SVG from our legend div relatively easily. (can we @dmfalke ?)

PR #1126, opened 2024-07-03
export both plotly plot and legend images

moontrip:
Hi @dmfalke This is an attempt to possibly address https://github.com/VEuPathDB/web-monorepo/issues/946, especially based on the last/latest comment, https://github.com/VEuPathDB/web-monorepo/issues/946#issuecomment-2117931778.

I have tried many ways, even trying to use html-to-image, which is a forked version of the dom-to-image and more updated library, but realized that dom-to-image was already installed and worked just fine for our purpose so I sticked to it. There were several issues but I could manage to find ways. From my various tests, this seems to work as intended, but just in case, I made this as a draft PR as there may be a better way 🤔

### DIFF
--- a/packages/libs/components/src/components/plotControls/PlotLegend.tsx
+++ b/packages/libs/components/src/components/plotControls/PlotLegend.tsx
@@ -34,6 +34,8 @@ export default function PlotLegend({
         type === 'colorscale' ||
         type === 'bubble') && (
         <div
+          // set id for export legend
+          id="plotLegend"
           style={{
             display: 'inline-block', // for general usage (e.g., story)
             border: '1px solid #dedede',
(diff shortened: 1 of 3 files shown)


dmfalke:
Thank you for working on this, @moontrip. I think this approach seems reasonable, but it will require some UX discussion. Also, we don't want to hard code DOM element IDs, as it creates an implicit coupling between the legend implementation and the `ExportPlotToImageButton` component.

I noticed that the associated issue is still in "Needs triage". I'm not sure this is considered urgent, at the moment. However, I think this PR serves as a good proof-of-concept. Perhaps a future iteration would be to pass an array of objects (maybe specifying DOM refs and file names) to `ExportPlotToImageButton`, which would be used to generate the image files. And, instead of rendering `ExportPlotToImageButton` in the plot components, it could be rendered in the viz components.

PR #1135, opened 2024-07-14
Throw synchronous error in usePromise context, if `throwError` is `true`.

### LINKED ISSUE #1136: EDA User Dataset: Error thrown when installation is queued
See this screenshot:
image

dmfalke:
fixes #1136

This PR fixes the semantics of `throwError` such that the error can be caught by the caller of `usePromise(..., { throwError: true })`.


bobular (approved):
Looks good!

PR #1140, opened 2024-07-16
Orthogroup tree table  pfam filter

dmfalke:
This PR introduces table filters to the tree table section of group pages. There are three filters: pfam domains, core/peripheral, and species.

The first two filters are applied after the popover is dismissed, while the third one is applied immediately. This is a bug, as the intended behavior is to apply the selection after the popover is dismissed. **Question: Which behavior is preferable?** 

Here is a screenshot:
image

### TODO

- [ ] Remove "Phyletic distribution" section
- [ ] Move "Protein" section to top
- [ ] Add ability to highlight table rows from species filter

### DIFF
--- a/packages/libs/components/src/components/tidytree/HorizontalDendrogram.tsx
+++ b/packages/libs/components/src/components/tidytree/HorizontalDendrogram.tsx
@@ -55,6 +55,10 @@ export interface HorizontalDendrogramProps {
    * highlight whole subtrees ('monophyletic') or just leaves ('none')
    */
   highlightMode?: 'monophyletic' | 'none';
+  /**
+   * highlight color (optional - default is tidytree's yellow/orange)
+   */
+  highlightColor?: string;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


bobular:
Hi @dmfalke 

I just made some [adjustments](https://github.com/VEuPathDB/web-monorepo/pull/1140/commits/a42207269eb0b2f003ab98035da080c25db60617) to avoid the species button overflowing with taxon abbreviations.

I also made the [update behaviour consistent](https://github.com/VEuPathDB/web-monorepo/pull/1140/commits/c50a85eeeec9ffa33a794ded8f5c5823b27924e9) between all three filters.

I wonder if all the popovers should have a "close" or "apply" button? (or even a cancel/apply...?)

bobular:
For "Add ability to highlight table rows from species filter" why don't we just highlight the rows that have their checkboxes checked?  This way users can use any search they like to highlight rows.

bobular:
> For "Add ability to highlight table rows from species filter" why don't we just highlight the rows that have their checkboxes checked? This way users can use any search they like to highlight rows.

Here's what it looks like (after filter, select all, unfilter)

image

PR #1143, opened 2024-07-18
Filter out unimplemented viz for new viz picker (single app mode)

### LINKED ISSUE #1142: ClinEpi's pass-through new visualization icons should not include "Map markers" - also consider dropping "Coming soon!" entries
The regular map visualization (not the full-screen app) uses two backend plugins: `map-markers` and `map-markers-overlay`. The problem is that the `map-markers-overlay` is getting a viz icon:

image

The `apps` endpoint returns all the computes and their visualizations. The `pass` compute response lists `map-markers` and `map-markers-overlay`.

The front end, in single-app ClinEpi mode, renders the viz icons with [NewVisualizationPicker](https://github.com/VEuPathDB/web-monorepo/blob/main/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx#L315) 

Note that Mbio and the full screen map use a similar component unhelpfully named `StartPage` ([here](https://github.com/VEuPathDB/web-monorepo/blob/main/packages/libs/eda/src/lib/core/components/computations/StartPage.tsx))

Both components iterate through the list of visualizations from the `apps` endpoint. The icon is greyed out if the front end does not have a plugin to deal with it. See here https://github.com/VEuPathDB/web-monorepo/blob/c803fc5b457fb85483c17e6418d8cd46400827b5/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx#L351-L354

Given that we probably want to get rid of the "Coming soon" icons, the easiest fix is likely to replace the `orderBy` here: 

https://github.com/VEuPathDB/web-monorepo/blob/c803fc5b457fb85483c17e6418d8cd46400827b5/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx#L345-L348

with a filter.  There's no need to fix `StartPage` because it isn't used to render the `pass` viz plugins.

It's not clear how long this has been a bug. The front end code hasn't changed recently, and I'm pretty sure the `apps` response hasn't either. Maybe we just didn't notice it.

bobular:
Fixes #1142 

Has not been widely consulted!

Before:
image

After:
image

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -342,6 +342,10 @@ export function NewVisualizationPicker(props: NewVisualizationPickerProps) {
       <Grid>
         {/* Filter completely skips back-end plugins with no front-end implementations.
          * So there are no more "coming soon" placeholders.
+         * If we want "Coming soon..." icons, we should make a ComingSoonViz
+         * implementation where isEnabledInPicker function returns false.
+         * It should also include a selectorIcon. Other required props can be
+         * stubbed with noops.
          */}
         {visualizationsOverview
           .filter((viz) => viz.name && visualizationPlugins[viz.name])
--- a/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx
@@ -8,7 +8,6 @@ import {
 } from 'react-router-dom';
 import Path from 'path';
 import { v4 as uuid } from 'uuid';
-import { orderBy } from 'lodash';
 import {
   Link,
   Loading,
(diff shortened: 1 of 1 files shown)


dmfalke:
While I agree that we probably don't need "coming soon..." right now, I would prefer to keep the existing logic and find another way to hide this icon, on the off chance clinepi development picks up again one day.

dmfalke:
I think we could flag this viz as "internal" on the client.

For example, take this line of code:
https://github.com/VEuPathDB/web-monorepo/blob/886a9947a3f6d13faa568d9383926504d0c8245b/packages/libs/eda/src/lib/core/components/computations/Types.ts#L49

We could change the type to something like this:
```ts
  visualizationPlugins: Partial<Record<string, VisualizationPlugin<any> | '__INTERNAL__' >>;
```

And filter out any such visualizations.


Even better might be to flip this logic somewhat so that vizs w/out a client implementation are always hidden, and if we want a "comming soon" icon, that has to be specified on the client with a new `ComingSoonViz` implementation.

dmfalke:
I thought about this some more. Let's go with the approach in this PR, but add a comment similar to this:

> If we want "Coming soon..." icons, we should make a `ComingSoonViz` implementation where `isEnabledInPicker` function returns `false`. It should also include a `selectorIcon`. Other required props can be stubbed with noops.

dmfalke (approved):

PR #1144, opened 2024-07-19
Instant filters for the orthogroup page

bobular:
Note that this targets `orthogroup-tree-table`, so can be merged without review.

Maybe we demo both or make a video with both?

### DIFF
--- a/packages/libs/coreui/src/components/inputs/SelectList.tsx
+++ b/packages/libs/coreui/src/components/inputs/SelectList.tsx
@@ -34,8 +34,6 @@ export default function SelectList<T>({
   );
 
   const onClose = () => {
-    if (instantUpdate) return; // the next two effects take care of the following when constantly updating
-
     onChange(selected);
     setButtonDisplayContent(
       selected.length ? selected.join(', ') : defaultButtonDisplayContent
@@ -61,6 +59,7 @@ export default function SelectList<T>({
    */
   useEffect(() => {
     setSelected(value);
+    if (instantUpdate) return; // we don't want the button text changing on every click
     setButtonDisplayContent(
       value.length ? value.join(', ') : defaultButtonDisplayContent
     );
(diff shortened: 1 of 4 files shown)


dmfalke:
Sorry for the delay. I will take a look at this today or tomorrow

packages/libs/coreui/src/components/inputs/SelectList.tsx
```diff
@@ -29,12 +34,23 @@ export default function SelectList<T>({
...
   };
 
+  /**
+   * Keep caller up to date with any selection changes
+   */
+  useEffect(() => {
+    if (instantUpdate) {
+      onChange(selected);
+    }
+  }, [onChange, selected, instantUpdate]);
```
dmfalke, line 52: This should be called in a callback used as the `onChange` prop of `CheckboxList`. We don't currently have one defined, so it will be a new callback.

dmfalke (changes requested):
I think the name of the new prop should be consistent with what we use for the checkbox select component. I'm fine with both being `instantUpdate`.

Also, I don't think `buttonDisplayContent` needs to be tracked with React state. I know this was already in place, but would be good to remove all unnecessary state and effects.

bobular:
Hi @dmfalke - thanks for the review.

I removed all but one effect thanks to your suggestion. While tesing things were still working in EDA/Lineplot I found a [new bug](https://github.com/VEuPathDB/web-monorepo/issues/1160) (already on qa and presumably prod) which I have a fix for too.

I don't think it's wise to get rid of the local state for `buttonDisplayContent `. The local state allows us to delay changing the button label until the drop-down is closed, which prevents it moving around as its contents change. I think this is OK.

dmfalke:
> I don't think it's wise to get rid of the local state for buttonDisplayContent . The local state allows us to delay changing the button label until the drop-down is closed, which prevents it moving around as its contents change. I think this is OK.

Makes sense. Thanks for checking!

Issue #712, opened 2023-12-07
Map - Donut and bar marker configuration panels make unused data requests


bobular:
I've been looking into this and don't see a quick fix :cry: 

The problem is in `ConfigPanelComponent` of Donut and Bar map type plugins. Preview-related data is fetched for both categorical and continuous variables here:

https://github.com/VEuPathDB/web-monorepo/blob/3bd842bbc41f2e0f03cdce3c3c4bc0f35e69cd74/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/DonutMarkerMapType.tsx#L124-L158

The `useCateogoricalValues` isn't so bad because it returns undefined pretty quickly if the variable is the wrong data shape. No back end request made.

However this bit (for continuous variables) `previewMarkerResult = useMarkerData(...)` does an actual back end request even if the variable is categorical.

That extra request will be cached by react-query but every click on a categorical marker config table will likely create a new un-cached query that is never used.

Maybe we can make it so that `useMarkerData` can "bail" if the wrong variable shape is given. This would likely require passing a new prop through `useMarkerData` and `useDistributionMarkerData` to set `useQuery`'s `enabled` to false. Not very nice.

**A better solution to try first** would be to create a `ContinuousMarkerPreview` component that is analogous to the already existing `CategoricalMarkerPreview` component that is already implemented. Conditional rendering chooses between the two and we would just neeed to make sure it gets the props it needs. Remember to do the same in BarMarker mode.

Should hopefully be low cost.

PR #1155, opened 2024-07-30
712 refactor continuous marker preview

chowington:
Fixes https://github.com/VEuPathDB/web-monorepo/issues/712


bobular (approved):
Thanks Connor - this refactor tidies things up a lot.

Dave and I could no longer reproduce/detect the "unused requests" in dev tools. What I suspect has happened is that work subsequent to the issue being created has standardised the `useMarkerData` requests so that both the request needed for the old `continuousMarkerPreview` and the current legend data request [here](https://github.com/VEuPathDB/web-monorepo/blob/30425bd0e89e8de23dc85f38ec244416e481bf87/packages/libs/eda/src/lib/map/analysis/mapTypes/plugins/DonutMarkerMapType.tsx#L464) are identical and with react-query caching we only see one in the logs.

Issue #1153, opened 2024-07-29
Unintended consequences of Network work upsetting the bipartite network


asizemore:
I would guess this happened because of the recent updates done to Network. What should happen is that the bipartite network should scroll within it's box.

<img width="1079" alt="Screen Shot 2024-07-29 at 10 43 54 AM" src="https://github.com/user-attachments/assets/ecad5fa9-2e01-4ab1-b6c4-fc81f0b80a9e">

The other unintended consequence is that all labels are hidden by default. For the bipartite network there are never issues with label overlap so let's change it so that the bipartite network has labels shown all the time.

moontrip:
@asizemore I have no idea of the first one concerning scrolling, but I think I know the reason for the second one, showing labels.
As you can remember, we have node label control for Network Viz, not for Bipartite network Viz. When I looked into Bipartite, I realize that it calls like: Bipartite network Viz > BipartiteNetworkPlot (BipartiteNetworkPlot.tsx) > NetworkPlot (NetworkPlot.tsx) > NodeWithLabel (Node.tsx)
On the other hand Networ Viz's process is like:  Network Viz > NetworkPlot > NodeWithLabel

The problem arises because we have changed NetworkPlot and NodeWithLabel components to implement node label control, more specifically, this `showLabel` prop (https://github.com/VEuPathDB/web-monorepo/blob/6b9bba0f55bee8b43cfa055977fa3b7d321e73c1/packages/libs/components/src/plots/NetworkPlot.tsx#L311). For Bipartite network viz that does not have node label control, `showLabel` prop becomes false, thus no label is shown.

To overcome this, I think we have one of the following two options: a) pass a new prop from BipartiteNetworkPlot to NetworkPlot (e.g., the name of "BipartiteNetwork") and add an additional OR (||) condition for `showLabel` prop to make it true for the case of Bipartite network; b) add node label control to Bipartite network Viz too.

asizemore:
Thanks @moontrip ! Given the two options, i'd pick (a). The bipartite network never has trouble with overlapping labels so the labels should always be shown, and therefore we don't need controls like we do for the regular network.

I gave a quick look at the components again, thanks to your link. What do you think about having the BipartiteNetwork send in all node labels as the `visibleNodeLables`? Then all labels would be in `visibleNodeLabels` and we wouldn't need a new prop.

moontrip:
[...]

PR #1157, opened 2024-07-31
fixing the issues of plot box and node labels

moontrip:
@asizemore This will address https://github.com/VEuPathDB/web-monorepo/issues/1153. A screenshot is attached here:

BipartiteNetwork fix

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
@@ -19,7 +19,7 @@ import {
   CorrelationBipartiteNetworkResponse,
 } from '../../../api/DataClient/types';
 import { twoColorPalette } from '@veupathdb/components/lib/types/plots/addOns';
-import { useCallback, useMemo } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 import { scaleOrdinal } from 'd3-scale';
 import { uniq } from 'lodash';
 import { usePromise } from '../../../hooks/promise';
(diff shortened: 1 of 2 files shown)


packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
```diff
@@ -342,6 +355,8 @@ function BipartiteNetworkViz(props: VisualizationProps<Options>) {
     labelTruncationLength: 40,
     emptyNetworkContent,
     getNodeMenuActions,
+    // pass visible node labels
+    visibleNodeLabels: visibleNodeLabels,
```
asizemore, line 359: Why not just set `visibleNodeLabels` to be `cleanedData.nodes.flatMap((node) => {
      return { value: node.label, label: node.label };` or similar? Do we need to use state here?

asizemore (commented):
Looks good @moontrip ! I just had a question on why we needed to use state with the change. Is there an advantage over just passing the labels directly?

packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
moontrip: @moontrip Thank you for your comment! I may be wrong without further test, but if I remember correctly, it caused some issues of updating visibleNodeLabels correctly, which was due to the fact that it was indirectly related to the Promise of the data request. But it happened with node label control (i.e., for network viz), so in this Bipartite network case, it may just work without using the useState. I will check it out next week if that is the case.

moontrip: @asizemore I checked it carefully and found that there was no issue without using useState as you guessed. 👍 Thus I simplified it by removing unnecessary codes. :)

asizemore (approved):
Looks great! Thank you!!

PR #1161, opened 2024-08-02
Prevent numerator/denominator checkbox confusion by replacing unstable empty array reference literals

### LINKED ISSUE #1160: EDA - lineplot numerator/denominator checkboxes get reset if you work very fast
To reproduce

* qa.clinepidb.org
* EDA
* any study
* make a lineplot
* choose x
* choose categorical y variable
* AS QUICK AS YOU CAN go and click a numerator checkbox
* you should see your checkbox be reset to unchecked

This appears to be due to the `visConfig` coming back from the user database after a PATCH update. You may not see it on faster connections?

The useEffect in `packages/libs/coreui/src/components/inputs/SelectList.tsx` is presumably resetting the local state of the component because `vizConfig.numeratorValues` has changed (even though it has actually not changed from an empty array to an empty array?)  Is there a rogue `[]`? Yes there is. Fix to follow

bobular:
Fixes #1160 

Fairly simple fix!


packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
```diff
@@ -2797,7 +2799,7 @@ export function AggregationInputs<F extends string, P extends Array<string>>(
   props: AggregationConfig<F, P>
 ) {
   const classes = useInputStyles();
-
+  console.log(props);
```
chowington, line 2802: Can you undo this change, please? Thanks!

chowington (approved):
Looks good after taking care of a leftover console log!

packages/libs/eda/src/lib/core/components/visualizations/implementations/LineplotVisualization.tsx
bobular: Oops - yes of course. Thanks for spotting!

Issue #1165, opened 2024-08-06
Networks: set cursor over label to pointer - third time's a charm


asizemore:
Seen on qa in both the network and bipartite networks.
A small thing, but when we mouse over the node label the cursor is not becoming a pointer. This is odd to me because a quick look through the code says it's set as a pointer in two places. So why it the mouse becoming a text selector instead?

It's a mystery to me!

moontrip:
@asizemore I will take a look at it 🤞

PR #1166, opened 2024-08-07
set cursor pointer for network node label

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/1165. 

@asizemore Simply put, it was a typo, though it took some time to find the reason as it was hard to notice 😓


asizemore (approved):
Hahaha no way! I read that line so many times! I'm glad you found it. Thank you!

moontrip:
@asizemore yes indeed. Hard to find it: I missed it several times too😄

PR #1168, opened 2024-08-08
Add tooltip reasons for disabled RadioButtonGroup options and style Map marker configuration panel a bit better

### LINKED ISSUE #1167: Map: comment out standard deviation option for continuous variable markers
As a temporary stop-gap to https://github.com/VEuPathDB/EdaNewIssues/issues/790 we could just remove the standard deviation option from `BarPlotMarkerConfigurationMenu` and `PieMarkerConfigurationMenu`.

bobular:
Fixes #1167 

Before / After
image image

(tooltip is `This option is currently disabled for maintenance reasons`)

Before / After (notice the removal of an all-disabled "Marker X-axis controls" section)
image image

Before / After
image image


moontrip (approved):
Looks good to me @bobular 👍

PR #1169, opened 2024-08-09
Place static help image next to group stats

bobular:
Here are the basics. Note the PR targets `orthogroup-tree-table`.

image


bobular:
It's now mostly done. We may prefer just to hardcode the width at 480px - I don't mind.  

For the screenshot I have removed the broken image from the model HTML blurb using dev tools. This will need editing by whomever edits this in the relevant file. There no longer needs to be any `` tag in that blurb above the table.

image

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/GroupStats.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useEffect, useRef, useState } from 'react';
+import { RecordTable } from './SequenceRecordClasses.SequenceRecordClass';
+import { RecordTableProps, WrappedComponentProps } from './Types';
+
+import eval_hist_img from '../../../images/eval-hist.png';
```
bobular, line 5: I tried to set up a path alias thingy in tsconfig.json (there was one already in webpack.config.js - but we don't seem to be using it) but I couldn't get it to work :-(

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/GroupStats.tsx
```diff
@@ -0,0 +1,54 @@
...
+            style={{
+              fontSize: '90%',
+            }}
+          >
+            This histogram is provided to aid the interpretation of E-values in
+            the adjoining table. E-values have been transformed using a negative
+            logarithm, so higher significance is represented further to the
+            right.
+          </figcaption>
+        ) : null}
```
dmfalke, line 50: I would set the `height` and `width` attributes of the `` tag. This will cause the browser to reserve the space needed for it, while the image is loading. I think this will also allow you to remove the `width` style from `<figure>`.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/GroupStats.tsx
bobular: OK, I'll do that! Thanks!

dmfalke (commented):
Take a look at my comment regarding width, and let me know what you think.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/GroupStats.tsx
bobular: So we shouldn't bother with the dynamic width determination?

bobular: Happy to remove it! Seems OK to hardcode the image dimensions if the image itself is hardcoded into the page!

bobular: Removed it!

dmfalke: Just seeing these comments. Hardcoding the image size is safer for a couple of reasons:
1. It reserves space on the page for the image.
2. If the image gets replaced with a huge version, it won't impact the layout.

dmfalke (approved):

PR #1171, opened 2024-08-09
Restore the buttonless phyletic distribution

bobular:
Here's the phyletic distribution back at the top of the page. Species filter-drop down still works as before!

image


dmfalke (approved):
Looks great!

PR #1172, opened 2024-08-09
Eda user dataset updates

dmfalke:
This PR addresses two issues:
1. Incorporate user datasets into the "All Analyses" and "Public Analyses" tables.
2. Allow analyses of community datasets to be made public.


jernestmyers (approved):
LGTM!

PR #1174, opened 2024-08-09
Improve NoDataError handling

### LINKED ISSUE #1173: Map - client-side fix for 500 error from `continuous-variable` that is currently exposed to user
Follow-up from https://github.com/VEuPathDB/web-monorepo/issues/708 - note that the examples in this ticket are fixed - the empty-bodied 204 response is correctly handled.

However, it can still be reproduced by a new example

1. VB megastudy map
2. Change the donut marker variable to Collection -> Specimen collection start date
3. Filter on species = Aedeomyia catasticta

image

Let's see if a client-side fix is possible here.

bobular:
Fixes #1173 

Before this fix, we were successfully intercepting `Could not generate continuous variable metadata` from the error thrown by the 500 response, so `isNoDataError(error)` was returning `true`. However the `MapFloatingErrorDiv` banner was displaying the raw error (see screenshots in issue).

Since we know it's a "no data error" it's OK to display a more user-friendly message (the same one we use when receiving a formal NoDataError that comes from the 204/null body response).  So now it says `The visualization cannot be made because no data remains after filtering.`

image


jernestmyers (approved):
Simple enough. Looks good!

PR #1175, opened 2024-08-09
Take a less agressive path when requesting user datasets from within eda

dmfalke:
This PR removes a code path that causes an error to be thrown, in favor of a conditional statement, for requesting user datasets within eda.

Resolves this error

```
ERROR
In order to use this feature, a VdiCompatibleWdkService must be configured.
```


bobular (approved):
LGTM. I'm not familiar with the bug this resolves but we're definitely not throwing an exception any more from `assertIsVdiCompatibleWdkService`

bobular:
> This addresses a bug you reported when running the `start` command for `@veupathdb/eda` 😄

Guess I'd better test that then!

bobular:
libs/eda dev server works for me but see https://github.com/VEuPathDB/web-monorepo/pull/1195 (I needed this too)

dmfalke:
> libs/eda dev server works for me but see #1195 (I needed this too)

ok, will look!

Issue #1180, opened 2024-08-14
Map: timeslider arrows layout with short variable name


bobular:
My UD has a short "date" variable name

![image](https://github.com/user-attachments/assets/ab19470c-fbe5-44ea-96b3-fb8f0e4b1e62)

Needs to be

![image](https://github.com/user-attachments/assets/874e56e3-9458-4000-9f8d-4d889df10c99)

add `margin-left/right: auto` to div containing the date range selector

bobular:
Maybe the arrows should be snug with the date picker inputs.

The arrows can have `margin-left: auto` and `margin-right: auto` respectively.

moontrip:
This is resolved by https://github.com/VEuPathDB/web-monorepo/pull/1185 so closed

PR #1185, opened 2024-08-20
adjusting the layout of the timeslider

moontrip:
This will address https://github.com/VEuPathDB/web-monorepo/issues/1180

Screenshots after adjusting layouts:
- temporarily reduced the width of the input variable button to mimick your screenshot with the variable, "date"

timeslider layout01

- 
timeslider layout02


packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
```diff
@@ -406,10 +406,9 @@ export default function TimeSliderQuickFilter({
                   constraints={timeSliderVariableConstraints}
                 />
               </div>
-              <div>
+              <div style={{ marginRight: 'auto', marginLeft: 'auto' }}>
```
bobular, line 409: Could you try it like this too please, and post some screenshots? (see other suggestion below)

```suggestion
              <div style={{ marginLeft: 'auto' }}>
```

packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
```diff
@@ -424,28 +423,29 @@ export default function TimeSliderQuickFilter({
...
+                  valueType={'date'}
+                  containerStyles={{
+                    flex: 1,
+                  }}
+                  // change the height of the input element
+                  inputHeight={30}
+                  disabled={!active}
+                />
+              </div>
+              <div style={{ marginRight: 'auto', marginLeft: 'auto' }}>
```
bobular, line 446:
```suggestion
              <div style={{ marginRight: 'auto' }}>
```

packages/libs/eda/src/lib/map/analysis/TimeSliderQuickFilter.tsx
bobular: Oh, so they are too close now? I hoped they'd be a bit closer. Maybe a fixed 10px margin on the side closest to the date pickers?

moontrip: @bobular Aha I see. Then we may set value instead of "auto". For example
- removing marginRight or marginLeft


- set "1em": I think it would be either: Didn't check the parent's css carefully but usually 1em = 12 or 16px

bobular: Looks great with 1em!

bobular: (yes, one auto, one fixed)

moontrip:
@bobular Thank you for your review and tests. I committed the change as you suggested.
I noticed an issue at main branch concerning React 18 by Jemery so I am discussing with him. I will merge this work after resolving the issue (if it is indeed an issue)

PR #1188, opened 2024-08-22
Fix how validation bundle error responses are handled

dmfalke:
This PR fixes how validation bundle errors are handled. We were passing the Error object to the `ok` function from the `JSON` module, which expects a parsed json object. This PR addresses that by first parsing `error.response`, and then passing that parsed value to the `ok` function. It also moves the checks to the `clientError` block, which is when we would expect a validation bundle error.

This can be tested by attempting to register a new user using an email address associated with an existing user. The validation error should be unpacked from the validation bundle and displayed as a human-readable message:

Screenshot from 2024-08-22 14-02-57


jernestmyers (approved):
LGTM!

PR #1189, opened 2024-08-23
Remove custom scroll anchoring implementation

dmfalke:
This PR remove custom JavaScript code to implement scroll anchoring. [Scroll anchoring](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring) is now supported by modern browsers, so this code is no longer necessary.

To test this feature, go to a gene page and open every section on the page. Then, scroll halfway down the page and reload. When the page reloads, content on screen should remain visible while content above is loaded, instead of being pushed off screen.


dmfalke:
After some more testing, I don't think we can merge this. I'm going to close this, and maybe revisit later.

PR #1191, opened 2024-08-26
Remove `searchTree` from redux store

### LINKED ISSUE #1184: Remove searchTree (derived state) from redux store
Derived state does not belong in the redux store. Not only is it an anti-pattern, it is also causing a delay in some rendering since it is dependent on `static/all-data-loaded`, which depends on other resources being loaded.

See https://github.com/VEuPathDB/web-monorepo/blob/501d0bbce2e850358c7ccf134068c85fb9c8d1a4/packages/libs/web-common/src/wrapStoreModules.js#L89-L93. Turn this transformation into a selector function.

dmfalke:
Fixes #1184

This PR removed the derived `searchTree` from the redux store.


jernestmyers (approved):
🚀

PR #1192, opened 2024-08-26
Eda user dataset detail

dmfalke:
This PR introduces rendering the user dataset detail page in the **Study details** in EDA (both the EDA Workspace and EDA Map applications).

A few things of note:
- User Data type-based customization for the details page are not included. Currently, the only addition includes links to the eda workspace and map, which are not needed in this new context.
- New props are included in the UserDatasetDetailController component to allow hiding some header elements. 

Here is a screenshot of what this looks like in the EDA Workspace:

image


jernestmyers (approved):
🚀

PR #1197, opened 2024-09-04
Add diamond support to multi-blast, and enable on ortho-site

### LINKED ISSUE #1201: OrthoMCL - Replace protein mapping tool with diamond multi-blast tool
Summary
----
We are going to replace the existing protein mapping tool with a multi-blast based tool. The following tasks are needed:

WDK model changes:
----
- [x] Add a question to the model. It only needs two params: input_sequence, and job_description.
- [x] Move input_sequence instruction from client to wdk model param's visibleHelp (ortho and genomics)

Static content:
----
- [x] Add page for protein mapping

Client changes:
---- 
- [x] The input_sequence param should only have a file upload option.
- [x] When a file is provided, use `multipart/form` content-type to upload file, rather than reading into memory.
- [x] Add space for visible help at the top of the question form
- [x] The title of the blast workspace should say something like "Map Proteins"
- [x] The results will show a preview of the tsv file. The easiest way to do this is to override the route for blast results and use a custom `BlastWorkspaceResult` component.
- [x] Only provide a single download of entire tsv.
- [x] Configure workspace to use protein mapping help page from static content, for the Help tab. 
- [x] Include header line in download file.
- [x] Make sure all internal links are relative (i.e., remove the hardcoded blast path).
- [x] #1210

### LINKED ISSUE #1210: Stop job polling when navigating away from result page for an in-progress job.

dmfalke:
This PR adds support for diamond-blastp to the multi-blast package. It also adds the multi-blast workspace to ortho-site.

Notes:
- The results view for a diamon-blastp job will only show a combined result.
- Only a TSV file is available for download.
- There is some hard coding that maps the `subject-title` column to three columns. This allows us to include the orthomcl sequence and group ids as separate columns.

Closes #1201 
Closes #1210


packages/libs/multi-blast/config-overrides.js
```diff
@@ -10,18 +10,37 @@ module.exports = function override(config, env) {
...
-    resolveLoader: {
-      ...config.resolveLoader,
-      modules: [
-        path.join(__dirname, 'node_modules'),
-        ...(config.resolveLoader.modules || ['node_modules']),
+    plugins: [
+      ...config.plugins,
+      new webpack.ProvidePlugin({
+        // This is needed by shape2geohash, used by MapVEuMap
+        process: 'process/browser',
```
bobular, line 26: Does multi-blast use shape2geohash? I'd guess it doesn't. Does something else need the `process` global JS variable?

dmfalke: I just copied that in an attempt to get the local dev server for multi-blast working. It can probably be removed.

bobular: I couldn't get the local dev server working with qa or feature back ends.

Following https://github.com/VEuPathDB/web-monorepo/tree/main/packages/libs/multi-blast README I set all but WDK_AUTH_KEY - which I wasn't sure about (I thought we haven't used this for years?)

Should I have been able to get it working? @dmfalke

dmfalke: No, I wasn't able to get it working, either. I've been using a local ortho-site using dfalke-b.orthomcl.org as a backend. I will try to get it working, maybe after the main work is complete.

PR #1199, opened 2024-09-07
Add new WDK record scope: record-collapsed

bobular:
See https://github.com/VEuPathDB/OrthoMCLModel/commit/23305fb7457254d80c05198a6c346d9a24ce3e41 where an example configuration change has been made on the `diamond` branch.

Here we handle it on the client side.

To do:
- [ ] documentation

### DIFF
--- a/packages/libs/wdk-client/src/Controllers/RecordController.tsx
+++ b/packages/libs/wdk-client/src/Controllers/RecordController.tsx
@@ -111,7 +111,7 @@ class RecordController extends PageController<Props> {
           return (
             attributes?.includes(id) ||
             tables?.includes(id) ||
-            scopes.includes('record-internal')
+            scopes.includes('record-internal') // do we need to handle 'record-collapsed'?
           );
         })
         .toArray();
--- a/packages/libs/wdk-client/src/Controllers/RecordController.tsx
+++ b/packages/libs/wdk-client/src/Controllers/RecordController.tsx
@@ -111,7 +111,7 @@ class RecordController extends PageController<Props> {
           return (
             attributes?.includes(id) ||
             tables?.includes(id) ||
-            scopes.includes('record-internal') // do we need to handle 'record-collapsed'?
+            scopes.includes('record-internal')
           );
         })
         .toArray();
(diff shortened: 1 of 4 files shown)


packages/libs/wdk-client/src/StoreModules/RecordStoreModule.ts
```diff
@@ -84,11 +84,22 @@ export function reduce(state: State = {} as State, action: Action): State {
...
+
+      const collapsedSections = getLeaves(categoryTree, (node) => node.children)
+        .filter(
+          (
+            node
+          ): node is CategoryTreeNode & { properties: { name: string[] } } =>
+            !!node.properties.scope?.includes('record-collapsed') &&
+            !!node.properties.name
+        )
+        .map((node) => node.properties.name[0]);
```
bobular, line 96: Is there ever >1 element in the `name` array?

packages/libs/wdk-client/src/Controllers/RecordController.tsx
```diff
@@ -111,7 +111,7 @@ class RecordController extends PageController<Props> {
           return (
             attributes?.includes(id) ||
             tables?.includes(id) ||
-            scopes.includes('record-internal')
+            scopes.includes('record-internal') // do we need to handle 'record-collapsed'?
```
dmfalke, line 114: No, it isn't needed here. This code comes into play when attributes or tables are specified in the url, which is a way to have a record page with only those attributes and tables. Clearly this needs to be better documented.

packages/libs/wdk-client/src/StoreModules/RecordStoreModule.ts
dmfalke: I don't think so, but the ontology properties are always multi-value.

FYI, you can use this function:
https://github.com/VEuPathDB/web-monorepo/blob/c02d237d2116a452a8195a51ad570b10e6ccdb2e/packages/libs/wdk-client/src/Utils/CategoryUtils.tsx#L91-L93

dmfalke (approved):
This looks great!

PR #1204, opened 2024-09-16
VDI: fix guest errors with community datasets

dmfalke:
This PR fixes some issue when guest users interact with community vdi datasets:

- Forbidden network response error
- Runtime reference error
- In EDA, the wrong default variable can be selected when switching studies. In addition to fixing this bug, the UI now includes the variable tree when this happens, allowing the user to select a valid variable.

_The prod branch will also get these changes._

### DIFF
--- a/packages/libs/eda/src/lib/core/hooks/promise.ts
+++ b/packages/libs/eda/src/lib/core/hooks/promise.ts
@@ -1,4 +1,5 @@
 import { useEffect, useState } from 'react';
+import { useValueBasedUpdater } from './state';
 
 export type PromiseHookState<T> = {
   value?: T;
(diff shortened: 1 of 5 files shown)


dmfalke:
Note that there is a backend issue, where guest users cannot request community dataset details. This is in progress (see [slack thread](https://epvb.slack.com/archives/C04M6TCHC66/p1726420460287049)).

dmfalke:
> The backend issue has been resolved. However, there is more front end work to do:

This is fixed now.

I also addressed a subtle state synchronization issue that caused an incorrect default variable to be selected.

aurreco-uga (approved):

PR #1214, opened 2024-09-26
Changes to HistogramFilter range selection bounds behaviour for integer variables

### LINKED ISSUE #1200: EDA Subsetting - Special handling of integer variable histogram
https://epvb.slack.com/archives/C0132PM1MT5/p1725654457552689?thread_ts=1725049427.107699&cid=C0132PM1MT5

Background
---

In EDA, range filters are treated as both min- and max-inclusive. However, histogram bins are treated as min-inclusive and max-exclusive. When brushing a histogram to select a range filter, the min of smallest selected bin, and the max of the largest selected bin are used for the min and max values of the filter, respectively. (We call this the "selected range".) This causes an inconsistent UI when selecting a bin of an integer variable, because the max value of the largest bin is not included in the count of the bin (as seen in the tooltip of the histogram), but that value _is_ included in the filter. That creates a mismatch in the counts of the selected bins, and the count of the subset.

The proposed solution is to subtract 1 from the max value of the selected range when defining the filter. This will make the histogram and the subset consistent. Because the UI is determined by the filter, we also have to add 1 to the selected range that is passed to the histogram.

In addition, we will append the text `, both included` to the filter chip tooltip.

The complete list of TODOs:
---

- [ ] Change [`HistogramFilter`](https://github.com/VEuPathDB/web-monorepo/blob/main/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx) so that we **subtract 1** from the `selectedRange` in the `onSelectedRangeChange` callback prop, and **add 1** to the `selectedRange` props, **for integer variables**.
- [ ] Add the text `, both included` to the filter chip, **for integer variables**.

bobular:
Fixes #1200 

I made a comment on the issue regarding the filter chip tooltip. Otherwise, all looks good!

### DIFF
--- a/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
@@ -514,6 +514,25 @@ function HistogramPlotWithControls({
     return { min: filter.min, max: filter.max } as NumberOrDateRange;
   }, [filter]);
 
+  // For integer variables, the graphical range highlighting needs to extend to (max + 1).
+  // This compensates for the (max - 1) adjustment in handleSelectedRangeChange.
+  // The (max - 1) logic is necessitated by the both-ends-inclusivity of filters, which is most
+  // noticable with integers. This approach does not handle real-valued variables with values
+  // that coincide with bin boundaries, and we don't have a plan yet how to deal with it. Subtracting
+  // 1e-8 does not seem attractive!
+  // Full description in https://github.com/VEuPathDB/web-monorepo/issues/1200
+  const selectedRangeForHighlighting = useMemo(():
+    | NumberOrDateRange
+    | undefined => {
+    if (selectedRange == null || variable == null) return;
+    if (variable.type === 'integer') {
+      return {
+        min: selectedRange.min,
+        max: (selectedRange.max as number) + 1,
+      } as NumberRange;
+    } else return selectedRange;
+  }, [selectedRange, variable]);
+
   // selectedRangeBounds is used for auto-filling the start (or end)
   // in the SelectedRangeControl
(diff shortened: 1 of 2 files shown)


dmfalke (commented):
This looks pretty good, but we are missing logic to handle the text inputs. E.g., if I type in the same number in both text inputs, the max value gets subtracted by 1.

https://github.com/user-attachments/assets/cf96ba08-7dd5-4566-b885-fbc2ad65ef93


I think the solution is to add 1 to the max when the text input field changes.

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -596,11 +578,7 @@ function HistogramPlotWithControls({
         {}, // no overrides
         true // use inclusive less than or equal to for min
       ),
-    [
-      defaultUIState.independentAxisRange,
-      defaultDependentAxisRange,
-      uiState.dependentAxisRange,
-    ]
+    [defaultUIState.independentAxisRange, dependentAxisMinPosMaxRange, uiState]
```
bobular, line 581: These dependencies were all wrong - yellow truncation stuff seems to work fine still after the fix.

packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
```diff
@@ -547,35 +546,18 @@ function HistogramPlotWithControls({
...
-              } as NumberOrDateRange,
-              selectedRangeBounds
-            )
-          );
-        } else {
-          updateFilter(); // clear the filter if range is undefined
-        }
-      }
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bobular, line 553: This wasn't needed :-)

PR #1215, opened 2024-09-27
OrthoMCL - updates to protein mapping tool

### LINKED ISSUE #1213: OrthoMCL - protein mapping with diamond updates
- [x] Allow all file types
- [ ] Update verbiage on results page
- [x] Remove revise link
- [x] Make it clear that you can drag and drop
- [x] regarding expired diamond jobs, remove the link to the results, and the "rerun" link. de-emphasize row

dmfalke:
Fixes #1213


bobular:
Ready for review. Maybe add an "also" into the drag-and-drop guidance? ("You can **also** drag...")

bobular (approved):
Looks good!

bobular:
Hi @dmfalke - this is hopefully all done now. I tested the individual results page with an expired job and now it doesn't keep polling or trying to rerun!

dmfalke:
> Hi @dmfalke - this is hopefully all done now. I tested the individual results page with an expired job and now it doesn't keep polling or trying to rerun!

Great! I'm taking a look now.

dmfalke:
Regarding the latest change, I'm not sure about the **Page not found** heading. I think a better heading would be **Diamond job - expired**.

I also noticed that there is a moment where the page heading says **Blast job - pending**. I think this is an intermediate component that needs to use the configured workspace title, or something like that.

Finally, the last request that hasn't been addressed is to add the number of rows in the preview table. It should say something like:

```
Showing the first 100 sequences in your query file.
```

If the number of number of rows is less than 100, it can say

```
Showing all 48 sequences in your query file.
```

This can replace the existing text of **Mapped proteins**.

bobular:
Thanks @dmfalke - all done except for the "Blast job - pending" verbiage. This is a bit trickier than it looks. I'll think about it tomorrow.

dmfalke:
> Thanks @dmfalke - all done except for the "Blast job - pending" verbiage. This is a bit trickier than it looks. I'll think about it tomorrow.

Take a look at the following to see how we are passing a configurable header value to other components: https://github.com/VEuPathDB/web-monorepo/blob/89b02fe70dea193addccef5d21b5961854507b67/packages/libs/multi-blast/src/lib/controllers/BlastWorkspaceRouter.tsx#L27

bobular:
> > Thanks @dmfalke - all done except for the "Blast job - pending" verbiage. This is a bit trickier than it looks. I'll think about it tomorrow.
> 
> Take a look at the following to see how we are passing a configurable header value to other components:
> 
> https://github.com/VEuPathDB/web-monorepo/blob/89b02fe70dea193addccef5d21b5961854507b67/packages/libs/multi-blast/src/lib/controllers/BlastWorkspaceRouter.tsx#L27

Yeah I've seen that, but what we provide to that prop from proteinMappingRouter is rather long-winded.

dmfalke:
I've reviewed and tested this, using both ortho and plasmo sites. It all looks great to me!

PR #1217, opened 2024-10-03
Fix issue with how a 409 response from the WDK service is handled

### LINKED ISSUE #1216: Websites will sometimes show a "Sorry an error occurred" popup before automatically reloading
On occasion, when first loading a website, an error popup will briefly appear before the page automatically reloads. It appears this is related to how the client code handles the 409 status from the WDK service.

dmfalke:
fixes #1216

This PR addresses an issue where errors are thrown when the client is clearing the data cache store and reloading the page. 

Changes include:

- Move a return statement outside of an if-block
- Expose `_isInvalidating` and use that to suppress the error modal.


bobular (approved):
Looks/works great!

PR #1225, opened 2024-10-09
Do not include referrer for jbrowse iframe, and address linting warnings

### LINKED ISSUE #1224: JBrowse returns a 413 for large configurations
This happens when we load jbrowse in an iframe. The parent window url includes the query parameters that are used by the iframe. The iframe request includes the parent window as a referrer, which causes the header payload to be very large, causing a 413 response.

dmfalke:
Fixes #1224

This PR adds a referrerPolicy that disabled the referrer header for jbrowse iframes.


aurreco-uga (approved):

dmfalke:
I may wait to merge this, to see if systems can increase the payload limit.

ryanrdoherty:
> I may wait to merge this, to see if systems can increase the payload limit.

Yes, please wait so we can see if we can figure out both fixes.

PR #1228, opened 2024-10-11
Use `useQuery` in place of `usePromise` for EDA subsetting and visualizations

bobular:
- [x] refactor data requests
- [x] make sure user db PATCH requests don't keep going forever
- [x] spinner behaviour
- [x] cancel main variable behaviour
- [x] check date histograms work (I saw a date-related runtime exception once when other things weren't working very well either)
- [x] test bad variable behaviour in MapVEu
- [x] barplot
- [x] boxplot
- [x] test mbio
- [x] lineplot
- [x] mosaics
- [x] scatter plot (maybe set `cacheTime` in `useQuery` to 60 seconds (in ms) for scatterplot?) - default is 5 minutes
- [x] test mbio scatter behaviour
- [ ] other plots (volcano, networks etc) --> https://github.com/VEuPathDB/web-monorepo/issues/1265


packages/libs/eda/src/lib/core/components/filter/MultiFilter.tsx
dmfalke, line 1: All "MultiFilter" variables appear to be broken. An example can be found with a clinepi-site (using qa.clinepi): http://localhost:8080/a/app/workspace/analyses/DS_841a9f5259/new/variables/PCO_0000024/ENVO_00003064.

image

dmfalke: I think it's because one of the deps is `null`ish:

https://github.com/VEuPathDB/web-monorepo/blob/65004b37ee7936ec7895b7eb4b19b40d86405032/packages/libs/eda/src/lib/core/hooks/cachedPromise.ts#L24

dmfalke (commented):
I started testing with a clinepi site, and ran into an issue with a MultiFilter. See details below.

packages/libs/eda/src/lib/core/components/filter/MultiFilter.tsx
bobular: Thanks @dmfalke  - I have fixed this now.
I did test it but I must have already checked some checkboxes before deploying the useQuery.

bobular:
Hi @dmfalke thank you for the review.
I cleaned up a lot of dependency warnings and the multifilter bug you spotted.

PR #1243, opened 2024-10-22
use `useDeferredValue` to improve responsiveness of filter checkboxes 

bobular:
Interestingly this works pretty much as expected - the filter checkboxes update way before the table is rerendered. Unclicking them is the same.

All in all it's a better UX... except... when the organism filter is filtered using the text search, e.g. for 'gam'.  Then when you click on the Anopheles gambiae Pest checkbox, the behaviour is OK, but when unchecking the same box the checkbox doesn't update until the tree-table has updated. If the text search is cleared and the same checkbox is unchecked, it unchecks fairly quickly, as intended. I wonder what's going on?

Note, we could encapsulate the `useState` and `useDeferredValue` with something like this:

```
import { useState, useDeferredValue } from 'react';

function useDeferredState<T>(initialValue : T) : [T, T, Dispatch<SetStateAction<T>>] {
  const [state, setState] = useState(initialValue);
  const deferredState = useDeferredValue(state);

  return [state, deferredState, setState];
}
```
If so, where would we put it?

### DIFF
--- a/packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
+++ b/packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
@@ -1,5 +1,6 @@
 import React, {
   CSSProperties,
+  useCallback,
   useDeferredValue,
   useMemo,
   useState,
@@ -289,6 +290,14 @@ export function RecordTable_Sequences(
     [mesaColumns]
   );
 
+  const handleSpeciesSelection = useCallback(
+    (species: string[]) => {
+      setSelectedSpecies(species);
+      setTablePageNumber(1);
+    },
+    [setSelectedSpecies, setTablePageNumber]
+  );
+
   if (
     !sortedRows ||
     (numSequences >= MIN_SEQUENCES_FOR_TREE &&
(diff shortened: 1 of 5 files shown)


dmfalke:
This is super cool! I'm at a loss regarding the organism tree filter. I'll poke around tomorrow.

About where to put the custom hook... coreui or components might make the most sense, since I could see either of those wanting to use the hook.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -440,10 +449,7 @@ export function RecordTable_Sequences(
       <RecordTable_TaxonCounts_Filter
         key={`taxonFilter-${resetCounter}`}
         selectedSpecies={volatileSelectedSpecies}
-        onSpeciesSelected={(species) => {
-          setSelectedSpecies(species);
-          setTablePageNumber(1);
-        }}
+        onSpeciesSelected={handleSpeciesSelection}
```
bobular, line 452: Note that this doesn't fix the problem!

bobular:
Still probably needs the custom hook but I'll do that later.

dmfalke:
I just pushed a commit that sets fixed dimensions for the organism filter. This prevents it jumping around the page and nodes expand/collapse.

dmfalke (approved):
This all looks great!

PR #1252, opened 2024-10-28
Orthogroup tree table  protein filter

bobular:
<s>There's an edge case where say you are filtered on 6 proteins and you select-all 6 proteins and then click "refine filter" the filter widget doesn't close automatically (because our `key` doesn't change)</s> Solved :heavy_check_mark:

### DIFF
--- a/packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
+++ b/packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
@@ -196,9 +196,11 @@ export function RecordTable_Sequences(
   // 2. core-peripheral radio button
   // 3. checked boxes in the Pfam legend
 
-  const [selectedColumnFilters, setSelectedColumnFilters] = useState<string[]>(
-    []
-  );
+  const [
+    selectedColumnFilters,
+    setSelectedColumnFilters,
+    volatileSelectedColumnFilters,
+  ] = useDeferredState<string[]>([]);
 
   const filteredRows = useMemo(() => {
     if (
@@ -626,7 +628,7 @@ export function RecordTable_Sequences(
           onSearchTermChange={setSearchQuery}
           recordDisplayName="Proteins"
           filterAttributes={filterAttributes}
-          selectedColumnFilters={selectedColumnFilters}
+          selectedColumnFilters={volatileSelectedColumnFilters}
           onColumnFilterChange={(keys) => setSelectedColumnFilters(keys)}
         />
         <div className="MesaComponent" style={{ marginRight: 'auto' }}>
--- a/packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
+++ b/packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
@@ -373,6 +373,17 @@ export function RecordTable_Sequences(
     setTablePageNumber,
   ]);
 
+  const treeProps = useMemo(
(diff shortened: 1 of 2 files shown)


bobular:
Note: there seems to be no urgent need to address the debouncing quirks in `RealTimeSearchBox` (unused `delayMs` prop) - performance seems OK as-is.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -457,6 +472,90 @@ export function RecordTable_Sequences(
...
+  };
+
+  const proteinFilter = (
+    <PopoverButton
+      buttonDisplayContent={`Proteins${
+        volatileProteinFilterIds.length > 0
+          ? ` (${volatileProteinFilterIds.length})`
+          : ''
+      }`}
+      key={volatileProteinFilterIds.join(':')}
```
dmfalke, line 496: This isn't needed
```suggestion
```

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -457,6 +472,90 @@ export function RecordTable_Sequences(
...
+    setHighlightedNodes([]);
+  };
+
+  const proteinFilter = (
+    <PopoverButton
+      buttonDisplayContent={`Proteins${
+        volatileProteinFilterIds.length > 0
+          ? ` (${volatileProteinFilterIds.length})`
+          : ''
+      }`}
```
dmfalke, line 495: Should we change the label when rows are selected, to indicate that the underlying filter is operable?

dmfalke (approved):
Overall, this is a good straw man. I anticipate lots of opinions and potential changes, so this is a really good and simple starting point.

I did make a couple of suggestions (see below).

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
bobular: I think it is necessary - or if we remove it we need another way to close the popover after an action button is clicked

image

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
bobular: I've added an asterisk just as a placeholder (though the asterisk is also used inside the popover, so maybe it's good)

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
dmfalke: I see. This could use a comment.

I'm not a big fan of using keys like this, but it does solve this problem. 

There are two ways we can avoid using key, here:
1. Have `PopoverButton` take an `open` prop that can be used to control the internal state.
2. Have `PopoverButton` take a ref that exposes a `close()` and an `open()` method.

Both have their pros and cons. The first option means introducing more state for the consumer, and more state management for the `PopoverButton` component, but it uses props, which is more common. The latter doesn't not introduce new state and works better with event handlers, but custom refs are not nearly as common to work with.

bobular: I didn't expose an `open()` method just yet. We can grow into that!

PR #1253, opened 2024-10-29
Orthogroup tree table - tooltips

dmfalke:
This PR address some perf issues with huge tables, and removes some unneeded react state.

### DIFF
--- a/packages/libs/components/src/components/tidytree/TreeTable.scss
+++ b/packages/libs/components/src/components/tidytree/TreeTable.scss
@@ -1,0 +1,6 @@
+.TreeTable {
+  --tree-table-row-height: 1em;
+  tr {
+    height: var(--tree-table-row-height);
+  }
+}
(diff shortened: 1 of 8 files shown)


bobular (commented):
I don't think we should change the behaviour of the popover button text in "instantUpdate" mode.

With this PR, the button text changes with each filter checkbox click, and this can change the layout of the buttons and then the popover moves with them. It's not very nice UX.

Hopefully `resetCounter` can be removed, but at one point it was necessary! (I don't remember why.)

Also, while we are tweaking the mouse-over/'tooltips' I saw that the mouse pointer is in `cursor: pointer` mode for all table cells in this table, whether or not they have clickable or mouse-over-able contents. (Compare this with the Pfam and other tables on the page.) I tried to find the culprit but couldn't. The file DataRow.scss looked likely but the class is spelled differently: `DataRow-Inline` (.scss file) vs `DataRow--Inline` (dev tools).

PR #1257, opened 2024-11-01
Marker animation mostly fixed

### LINKED ISSUE #1227: Map: semantic marker animation is broken on zoom in (but not zoom out)
This is in Firefox and Chrome, on production and qa!

The markers aggregate nicely (swarm in to a single marker) on zooming out. But on zooming in, the markers do not animate when disaggregating :sob:

bobular:
Fixes #1227 

The only problem I see is that when zooming in, it would be nice to have the old aggregated marker on top, and have the new disaggregated markers animate out from underneath it. I couldn't make this happen :-( (see the comments in updateMarkers.tsx)  It's not a deal-breaker for me.

Needs a lot more heavy testing though.


packages/libs/components/src/map/animation_functions/geohash.tsx
```diff
@@ -33,10 +33,9 @@ export default function geohashAnimation({
      **/
     zoomType = 'in';
     const hashDif = currentGeohash.length - prevGeoHash.length;
-    // Get a new array of new markers with existing position property
-    // Set final render markers to the cloneArray which holds the new markers with
-    // their new starting location
-    consolidatedMarkers = updateMarkers(markers, prevMarkers, hashDif);
+    // Get array of new markers with old positions
+    const cloneArray = updateMarkers(markers, prevMarkers, hashDif);
+    consolidatedMarkers = [...prevMarkers, ...cloneArray];
```
bobular, line 38: Reversing the order of the spreads here does not fix the "old marker hidden under new markers" problem.

dmfalke:
> it would be nice to have the old aggregated marker on top

Did you try setting a higher z-index value for the old markers?

packages/libs/components/src/map/SemanticMarkers.tsx
dmfalke, line 1: Can you explain how removing `enqueueZoom` fixes the issue?

dmfalke (commented):
Everything seems to work! I don't really understand how your change fixed things. Can you provide a brief explanation?

bobular:
> > it would be nice to have the old aggregated marker on top
> 
> Did you try setting a higher z-index value for the old markers?

Yes, I tried `zIndexOffset` and setting CSS z-index via classes. Neither made any difference, in either direction. Nor did the ordering in the spread operators making the combined array of markers. I tested it with a very slow animation time and adding an offset to the marker's `position.lng` so that I could see both old and new side by side.

dmfalke:
> > > it would be nice to have the old aggregated marker on top
> > 
> > 
> > Did you try setting a higher z-index value for the old markers?
> 
> Yes, I tried `zIndexOffset` and setting CSS z-index via classes. Neither made any difference, in either direction. Nor did the ordering in the spread operators making the combined array of markers. I tested it with a very slow animation time and adding an offset to the marker's `position.lng` so that I could see both old and new side by side.

That surprising. If the markers are in the same stacking context, z-index should work. Might look into it later :slightly_smiling_face:

dmfalke (approved):

packages/libs/components/src/map/SemanticMarkers.tsx
bobular: To be honest, not entirely sure. I suspect React 18-related timing changes messed something up. There was nothing else to "blame" as far as I could see in the animation-related source files.

Although there's a remaining niggle with the z-index ordering of old vs new markers during animation, I am happy to remove code.

The enqueued `setTimeout` isn't necessary because on the next render (which is now debounced to the same milliseconds as animation (300ms)) the animation function will basically be a no-op because the geohash level is the same for `prevRecenteredMarkers` and `recenteredMarkers` - so it returns `recenteredMarkers` which will replace the previous concatenated old+new markers.

bobular:
Hi @dmfalke  - I snuck in a new [fix](https://github.com/VEuPathDB/web-monorepo/pull/1257/commits/aa42bc8a0f5a7fb01b344f089f44886a5382b8e3). Looks harmless to me*, so I'm merging. 

(* the worst that could happen is that some animation is broken, which it already was before this PR! - though it all looks good to me)

PR #1259, opened 2024-11-01
Orthogroup tree table  1254 fixes

### LINKED ISSUE #1254: Ortho 7 - minor UX upgrades
- [x] warning about filtered trees not being same as a tree from filtered protein seqs
- [x] tree/table mismatch warning verbiage and link to contact page, make sure all cases are handled (tree is superset of table not currently caught?)
- [x] Filter should capitalize Core/Peripheral
- [x] See if core/peripheral counts are easy
- [x] Text search instructions are wrong... fixed the search behaviour to match the other tables
- [x] table paging and protein filtering are a bit dodgy. Probably just need to reset to page 1 when applying the protein filter.
- [x] Test `SelectList` button labelling in EDA/MBio (lineplot numerator/denominator, mbio diff abundance condition selectors and elsewhere?)
- [x] Handle excessive "Previous Groups" in header with a "more..." link.

bobular:
Resolves #1254 

I looked into improving the responsiveness of the table paging controls with useDeferredState, but it was not possible due to the memoisation of `mesaState` - which has a mix of "slow" and "volatile" dependencies. I wonder if that memoisation isn't actually necessary? (Answer: I removed it and this reintroduced the janky search field checkboxes...)

### DIFF
--- a/packages/libs/coreui/src/components/inputs/SelectList.tsx
+++ b/packages/libs/coreui/src/components/inputs/SelectList.tsx
@@ -1,6 +1,9 @@
 import { ReactNode, useCallback, useEffect, useState } from 'react';
 import PopoverButton from '../buttons/PopoverButton/PopoverButton';
-import CheckboxList, { CheckboxListProps } from './checkboxes/CheckboxList';
+import CheckboxList, {
+  CheckboxListProps,
+  Item,
+} from './checkboxes/CheckboxList';
 
 export interface SelectListProps<T> extends CheckboxListProps<T> {
   children?: ReactNode;
@@ -30,13 +33,13 @@ export default function SelectList<T>({
 }: SelectListProps<T>) {
   const [selected, setSelected] = useState<SelectListProps<T>['value']>(value);
   const [buttonDisplayContent, setButtonDisplayContent] = useState<ReactNode>(
-    value.length ? value.join(', ') : defaultButtonDisplayContent
+    getDisplayContent(value, items, defaultButtonDisplayContent)
   );
 
   const onClose = () => {
     onChange(selected);
     setButtonDisplayContent(
-      selected.length ? selected.join(', ') : defaultButtonDisplayContent
+      getDisplayContent(selected, items, defaultButtonDisplayContent)
     );
   };
 
@@ -61,9 +64,9 @@ export default function SelectList<T>({
     setSelected(value);
     if (instantUpdate) return; // we don't want the button text changing on every click
     setButtonDisplayContent(
-      value.length ? value.join(', ') : defaultButtonDisplayContent
+      getDisplayContent(value, items, defaultButtonDisplayContent)
     );
-  }, [value, defaultButtonDisplayContent]);
+  }, [value, items, defaultButtonDisplayContent]);
 
   const buttonLabel = (
     <span
@@ -102,4 +105,18 @@ export default function SelectList<T>({
       </div>
     </PopoverButton>
   );
+}
+
+// Returns button display content based on `value` array, mapping to display names from `items` when available.
+// If no matching display name is found, uses the value itself. Returns `defaultContent` if `value` is empty.
+function getDisplayContent<T>(
+  value: T[],
+  items: Item<T>[],
+  defaultContent: ReactNode
+): ReactNode {
+  return value.length
+    ? value
+        .map((v) => items.find((item) => item.value === v)?.display ?? v)
+        .join(', ')
+    : defaultContent;
 }
--- a/packages/libs/coreui/src/components/inputs/SelectList.tsx
+++ b/packages/libs/coreui/src/components/inputs/SelectList.tsx
@@ -18,7 +18,7 @@ export interface SelectListProps<T> extends CheckboxListProps<T> {
   instantUpdate?: boolean;
 }
 
-export default function SelectList<T>({
+export default function SelectList<T extends ReactNode>({
   name,
   items,
   value,
@@ -109,14 +109,16 @@ export default function SelectList<T>({
 
 // Returns button display content based on `value` array, mapping to display names from `items` when available.
 // If no matching display name is found, uses the value itself. Returns `defaultContent` if `value` is empty.
-function getDisplayContent<T>(
+function getDisplayContent<T extends ReactNode>(
   value: T[],
   items: Item<T>[],
   defaultContent: ReactNode
 ): ReactNode {
   return value.length
     ? value
-        .map((v) => items.find((item) => item.value === v)?.display ?? v)
-        .join(', ')
+        .map<ReactNode>(
+          (v) => items.find((item) => item.value === v)?.display ?? v
+        )
+        .reduce((accum, elem) => (accum ? [accum, ',', elem] : elem), null)
     : defaultContent;
 }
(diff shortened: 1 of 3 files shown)


packages/libs/coreui/src/components/inputs/SelectList.tsx
```diff
@@ -15,7 +18,7 @@ export interface SelectListProps<T> extends CheckboxListProps<T> {
   instantUpdate?: boolean;
 }
 
-export default function SelectList<T>({
+export default function SelectList<T extends ReactNode>({
```
dmfalke, line 21: This seems strange to me. `T` is the type of the value. I don't think we ever want that to be a ReactNode. I think we should have `T extends string`, to be honest.

The change you made for `buttonDisplayContent` is perfect.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -39,6 +39,10 @@ import {
 import { RecordTable_TaxonCounts_Filter } from './RecordTable_TaxonCounts_Filter';
 import { formatAttributeValue } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
 import { RecordFilter } from '@veupathdb/wdk-client/lib/Views/Records/RecordTable/RecordFilter';
+import {
+  areTermsInStringRegexString,
+  parseSearchQueryString,
+} from '../../../../../../../libs/wdk-client/lib/Utils/SearchUtils';
```
dmfalke, line 45:
```suggestion
} from '@veupathdb/wdk-client/lib/Utils/SearchUtils';
```

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -404,7 +408,8 @@ export function RecordTable_Sequences(
   if (
     mesaRows != null &&
     sortedRows != null &&
-    mesaRows.length !== sortedRows.length
+    (mesaRows.length !== sortedRows.length ||
+      mesaRows.length !== leaves?.length)
```
dmfalke, line 412: Are both comparisons necessary?

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -419,8 +424,13 @@ export function RecordTable_Sequences(
       <Banner
         banner={{
           type: 'warning',
-          message:
-            'Tree and protein list mismatch. Please contact the helpdesk',
+          message: (
+            <span>
+              A data processing error has occurred on our end. We apologize for
+              the inconvenience. If this problem persists, please{' '}
+              <Link to="/contact-us">contact us</Link>.
```
dmfalke, line 431: Maybe add `target="_blank"`

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -622,6 +634,22 @@ export function RecordTable_Sequences(
...
+        {MAX_SEQUENCES_FOR_TREE.toLocaleString()} sequences
+      </span>
+    ) : filteredRows.length < sortedRows.length ? (
+      <span>
+        Note: The ortholog group's phylogeny has been pruned to display only the
+        currently filtered proteins. This may differ from a tree constructed{' '}
+        <i>de novo</i> using only these sequences.
+      </span>
+    ) : undefined;
+
```
dmfalke, line 652: I have a concern about putting this new message in the same place as the other message. It gets lost. Wouldn't it make more sense to add this to the table description where we describe the ability to filter?

dmfalke (commented):
I think this is mostly great. I made several comments, but I don't think any are show stoppers.

packages/libs/coreui/src/components/inputs/SelectList.tsx
bobular: Thanks. I'll try that, but it was my understanding that T needed to be something "compatible with" ReactNode.

dmfalke: I think of `value` as what would be included in a form request, which is always a string. A number could also work, but I'm not convinced we need to do that.

This would be a breaking change, but I'm not sure if it would actually break anything.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
bobular: You mean it gets lost because it's the same colour and position as the "too many proteins" message that was almost certainly there just before they filtered?

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
bobular: Yes, I think so.

`sortRows()` returns the protein table rows in tree leaf order. Proteins that are missing from the tree are an obvious problem. But the tree can also be bigger (more leaves) than the protein table (though Rich says it won't happen) - it's best to play on the safe side.

Just using `mesaRows.length !== leaves?.length` won't protect against a protein table of X rows and a tree with X leaves, but where none of the proteins/leaves match each other.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
bobular: I'm not going to tackle this tonight. I don't really see it as a problem, personally. But we can discuss tomorrow.

packages/libs/coreui/src/components/inputs/SelectList.tsx
bobular: I've fixed it as you suggested. If it's a breaking change, it hasn't broken anything within this repo at least!

bobular:
I'd be grateful if you could have a quick look at the latest changes, @dmfalke - thanks!

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
dmfalke: > You mean it gets lost because it's the same colour and position as the "too many proteins" message that was almost certainly there just before they filtered?

Yes, that was my thinking. Let's leave it as-is and see if others raise it as an issue.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -98,7 +98,10 @@ export function RecordTable_Sequences(
   const numSequences = mesaRows.length;
 
   const treeResponse = useOrthoService(
-    (orthoService) => orthoService.getGroupTree(groupName),
+    (orthoService) => {
+      if (numSequences < 3) return Promise.resolve(undefined);
```
dmfalke, line 102: Should probably use `MIN_SEQUENCES_FOR_TREE` here.

PR #1263, opened 2024-11-08
JBrowse2 component integration

bobular:


packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/common/JBrowse2.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const location = 'AgamP4_2L'; // params.get('loc');
+  //  const tracks = params.get('tracks'); // comma-delimited track ids
+
+  const [config, setConfig] = useState<JBrowseConfig | undefined>();
+  const [loading, setLoading] = useState(true);
+
+  // Could use wdk usePromise? or react-query?
+  useEffect(() => {
+    // Fetch the config JSON
+    fetch('/jbrowse2/config.json')
```
dmfalke, line 34: Ideally, this would be a configuration value that gets set in configer. Something like `jbrowse2_root_url = /jbrowse2`. Then, you can update `__SITE_CONFIG__` in `EbrcWebsiteCommon/Model/lib/conifer/roles/conifer/templates/EbrcWebsiteCommon/appBase.html.j2` to include that value. Then, update `packages/libs/web-common/src/config.ts` to export that value. Finally, you can import that here via `import { jbrowse2RootUrl } from 'web-common/lib/config';`.

dmfalke (commented):
Looking good! I made a comment :)

PR #1270, opened 2024-11-12
Move record page perisistence to wdk-client package

dmfalke:
This PR moves the logic for persisting the state of a record page from genomics-site to wdk-client.

The state that gets persisted includes:
1. Page sections that are expanded.
2. Table rows that are expanded.
3. If navigation panel is expanded.

Changes include:
1. Moving persistence code to wdk-client.
2. Default navigation panel collapsed based on number of top-level categories
3. Add a method to `RecordController` component that can be overridden, to determine the default expanded sections.

I've tested this both with `main` and with `orthogroup-tree-table`, and all works well.

We may want to consider also persist the table settings: sorting, search filter, and filter columns.


bobular (approved):
Looks good. I see where most of the code has moved from/to and the new default section method. Tested in ortho-site and genomics-site and all looks good. Deleted local storage and gene model section is open by default. :-)

PR #1274, opened 2024-11-14
Keyboard a11y improvements

dmfalke:
This PR addresses some issues with keyboard accessibility:

- General: Enable focus ring.
- PopoverButton: Restore focus to button when menu is closed.
- SwissArmyButton: Do not override outline style by default.
- CheckboxTree: Retain focus on internal node when toggling expansion state via keyboard.
- UserMenu (header): Make fully keyboard accessible (previously could not tab into it).
- Tooltip: Make fully keyboard accessible.
- Strategy: Make step edit button fully keyboard accessible.
- Strategy: Improve focus ring style for basket button in results table.
- Strategy: Make organism filter show/hide button fully keyboard accessible.


aurreco-uga (approved):

dmfalke:
After merging this, I realized I should have used [:focus-visible](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible), instead of :focus. I will update this in a new PR.

PR #1276, opened 2024-11-15
coreui NoteBox component

dmfalke:
This PR introduces a `NoteBox` component. It is meant to be an alternative to `Banner`, integrating better on content-heavy pages, like record pages.

TODO: Add storybook entries.


packages/libs/coreui/src/stories/containers/NoteBox.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { mutedGreen, mutedMagenta } from '../../definitions/colors';
+
+export default {
+  title: 'Containers/NoteBox',
+  component: NoteBox,
+} as Meta;
+
+const Template: Story<Props> = function Template(props) {
+  return (
+    <UIThemeProvider
```
asizemore, line 14: Is there a reason to add the theme here? Or is it just to check that we get the correct colors regardless of theme?

packages/libs/coreui/src/components/containers/NoteBox.tsx
```diff
@@ -0,0 +1,47 @@
+import { css } from '@emotion/react';
+import React, { ReactNode } from 'react';
+import { error, warning, mutedBlue } from '../../definitions/colors';
+
+export interface Props {
+  type: 'info' | 'warning' | 'error';
+  children: ReactNode;
+}
+
+const baseCss = css`
```
asizemore, line 10: Perhaps additionally a default font? I believe coreui generally uses Inter or Roboto

packages/libs/coreui/src/stories/containers/NoteBox.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+        ultrices purus urna tristique mattis consequat. Posuere volutpat
+        facilisi natoque dictumst dignissim magna dapibus. Taciti vel a etiam
+        curabitur velit torquent. Fusce interdum dictum vulputate sollicitudin
+        nulla. Orci placerat congue odio aptent enim mauris. Turpis nec rhoncus
+        eleifend eleifend eget. Auctor sed nullam vestibulum quisque egestas;
+        nullam aenean ante.
+      </div>
+    ),
+  },
+});
```
asizemore, line 74: A story using a different type of child (for example, an image or some other non-text element) would be helpful if time allows

asizemore: or an expandable one

packages/libs/coreui/src/components/containers/NoteBox.tsx
```diff
@@ -0,0 +1,47 @@
...
+`;
+
+export function NoteBox(props: Props) {
+  const finalCss =
+    props.type === 'warning'
+      ? warningCss
+      : props.type === 'error'
+      ? errorCss
+      : infoCss;
+  return <div css={finalCss}>{props.children}</div>;
```
asizemore, line 46: Perhaps for later, but it'd be nice to pair the icon with the NoteBox type. We picked one icon for error, warning, and info ( a looong time ago and i think it was never actually impllemented). Of course, one could pass that as part of the child, but if we want to keep coreui very opinionated then there could be a `useIcon` prop and if that is true. the appropriate icon for that NoteBox style appears.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -662,23 +663,7 @@ export function RecordTable_Sequences(
...
-            background: 'rgb(255, 228, 228)',
-            gap: '1em',
-            marginBottom: '1em',
-            fontWeight: 500,
-          }}
-        >
-          {warningText}
-        </div>
-      )}
+      {warningText && <NoteBox type="error">{warningText}</NoteBox>}
```
asizemore, line 666: we should either change `warningText` to `errorText` or use the `type='warning'`. Having "warning" text in an error Notebox is confusing.

asizemore (commented):
Mostly just suggestions and questions. Otherwise looks really good! And it's 1000x better than what we have.

One big one - I'm not seeing the "Read more"/"Read less" button here, but I believe i should be?

Screen Shot 2024-11-20 at 5 46 27 PM

asizemore:
ahh i found it. 
Screen Shot 2024-11-20 at 5 49 27 PM

It's tough when so much is populated from whatever we're given from the backend!

dmfalke:
> Mostly just suggestions and questions. Otherwise looks really good! And it's 1000x better than what we have.
> 
> 
> 
> One big one - I'm not seeing the "Read more"/"Read less" button here, but I believe i should be?
> 
> 
> 
> Screen Shot 2024-11-20 at 5 46 27 PM
> 
> 

I fixed a bug related to this. Maybe on this branch? Anyway, I'll take a look.

dmfalke:
> > Mostly just suggestions and questions. Otherwise looks really good! And it's 1000x better than what we have.
> > 
> > 
> > 
> > One big one - I'm not seeing the "Read more"/"Read less" button here, but I believe i should be?
> > 
> > 
> > 
> > Screen Shot 2024-11-20 at 5 46 27 PM
> > 
> > 
> 
> I fixed a bug related to this. Maybe on this branch? Anyway, I'll take a look. 

Oh, I know what's going on here. This specific case is using a `<details>` element. It is likely closed by default, which means the overflow detection isn't triggered. This is actually kind of what we want, otherwise there would be two methods to collapse the content.

Generally, I prefer the `<details>` method, but I added the other one for an ortho use case. I probably should have argued my case 🙂

packages/libs/coreui/src/components/containers/NoteBox.tsx
dmfalke: For content, I think we should use the same font as the rest of the site, but I will take a look at using one of these other fonts.

packages/libs/coreui/src/stories/containers/NoteBox.stories.tsx
dmfalke: Yes, I agree! I will add a couple more, soon.

packages/libs/coreui/src/components/containers/NoteBox.tsx
dmfalke: Agreed!

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
dmfalke: This is one of those things where someone wanted the box to be red, even though it isn't strictly an error. I will change it to `errorText` for clarity.

dmfalke:
While testing this PR, I noticed a bug that was introduced by #1270:

The persistence of record page table state was also moved to wdk-client. There is at least [one place](https://github.com/VEuPathDB/web-monorepo/blob/7b84bdc7fadf994b1fe9c16632c60f9214125ea2/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/records/GeneRecordClasses.GeneRecordClass.jsx#L1643-L1650) that uses the old `updateTableState` action, which needs to be updated to use the new action. Additionally, the [logic that handles scrolling to the selected table row](https://github.com/VEuPathDB/web-monorepo/blob/7b84bdc7fadf994b1fe9c16632c60f9214125ea2/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/common/RecordTableContainer.jsx#L60-L98) also needs to be moved to RecordTableSection.

I will make a new issue for this.

packages/libs/coreui/src/components/containers/NoteBox.tsx
dmfalke: I think I'm going to leave the font as-is.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
dmfalke: Done!

dmfalke:
I've added icon support, and made some additional stories to highlight the option. I also modified the padding and border width to compensate for the additional space needed for the icon.

asizemore (commented):
Screen Shot 2024-11-26 at 8 03 51 AM

I was hoping for a good way to align the icon with the text. First i tried the `flex` + `align-items-center` route, but that only worked until the text was multiple lines. Next I tried the less elegant way of just adding a negative top margin to the icon. That worked, but requires all of the icons to be the same size. I believe that to be true, but am not 100% certain. What do you think?

PR #1283, opened 2024-11-21
Eda notebook

dmfalke:


packages/libs/eda/src/lib/core/components/VariableLink.tsx
```diff
@@ -70,12 +70,12 @@ export const VariableLink = forwardRef(
         tabIndex={0}
         style={finalStyle}
         onKeyDown={(event) => {
-          event.preventDefault();
```
dmfalke, line 73: This was breaking keyboard accessibility.

asizemore:
Wiring up the backend should probably be a different PR, especially given how big this has become. So i just added the wdk param below the notebook. That makes the search still functional, and would let us merge this PR, and then attack the different tasks that are left in a more modular way.

https://github.com/user-attachments/assets/9e4d5db3-cd83-4b40-be4f-91a1c3b16466

@bobular I'll be knocking on your door for a review soon!

bobular (commented):
I seem to have a review in progress. WIll submit it now!

bobular:
I've merged in my branch, but I've commented out the "persistence to WDK param" and brought back the wgcna module drop-down for now.

Next step will be to integrate the notebook with the broader plan laid out in the confluence doc.

PR #1284, opened 2024-11-21
Fix pfam filter button issues + replace group stats histogram image

### LINKED ISSUE #1280: Ortho 7 - beta bugs
- [x] Pfam filter button is now a bit crazy (probably related to the change @bobular recently made allowing `ReactNode` rendering
- [x] filters freeze up with multiple pfam filters active (explanation in comment below)

image

bobular:
Partially addresses #1280 

Was unable to tackle the second item in the issue. I think we should go ahead and merge this. That's likely to be a tricky one.


dmfalke (approved):
Nice!

PR #1298, opened 2024-12-09
Mesa fix layout

dmfalke:
This PR addresses some recently introduced layout issues with Mesa table, specifically when rows are either selectable or expandable.

With the relatively recently `position: sticky` css value, we can dramatically simplify the html and remove a decent amount of runtime javascript.

These changes impact all uses of Mesa table:
- Dataset and study pages
- Strategy results
- Record page tables
- EDA categorical filters
- TreeTable (ortho group record's protein table)

Seeing as this is an urgent fix (the bug is on production), I would appreciate a quick review. I'm happy to go over changes in more detail, after this is merged.

### DIFF
--- a/packages/libs/coreui/src/components/Mesa/Ui/DataTable.jsx
+++ b/packages/libs/coreui/src/components/Mesa/Ui/DataTable.jsx
@@ -7,22 +7,19 @@ import { defaultMemoize } from 'reselect';
 import HeadingRow from './HeadingRow';
 import DataRowList from './DataRowList';
 import { makeClassifier, combineWidths } from '../Utils/Utils';
-import { MESA_SCROLL_EVENT, MESA_REFLOW_EVENT } from './MesaContants';
 
 const dataTableClass = makeClassifier('DataTable');
 
 class DataTable extends React.Component {
   constructor(props) {
     super(props);
-    this.widthCache = {};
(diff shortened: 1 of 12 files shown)


aurreco-uga (approved):

packages/libs/coreui/src/components/Mesa/Ui/DataTable.jsx
```diff
@@ -123,11 +122,9 @@ class DataTable extends React.Component {
     if (this.props.rows.length === 0 || this.props.filteredRows.length === 0)
       return;
 
-    this.setState({ dynamicWidths: null, tableWrapperWidth: null }, () => {
-      this.widthCache = {};
-      const { columns } = this.props;
-      const hasSelectionColumn = this.hasSelectionColumn();
-      const { contentTable, getInnerCellWidth } = this;
+    this.setState({ dynamicWidths: null }, () => {
```
bobular, line 125: I'm confused as to where `this.state.dynamicWidths` is ever accessed?

This seems to be more about setting `this.widthCache`.

dmfalke: Thanks for catching. That was an oversight. I meant to purge all occurrences of dynamicWidths. Will fix in a follow-up.

PR #1299, opened 2024-12-11
Fix volcano plot thumbnail out of sync

### LINKED ISSUE #1297: Volcano thumbnail not loading
Seen on both qa and prod.

To reproduce, create a volcano plot, then go back to the thumbnail list. The first time we see the thumbnail is missing.

Screen Shot 2024-12-05 at 9 31 26 AM

Open the volcano plot again, go back to thumbnail list. The thumbnail is back!

But why?

asizemore:
Resolves #1297 

From the start the `visualization.descriptor.currentPlotFilters = undefined` if there are no filters. It works fine if we do have filters. In the other visualizations it initializes to `[]` as it should if the analysis has no filters. 

This PR has a fix, but I doubt it's really the direction we should go. I'd like to ask if folks have any other ideas!

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/VolcanoPlotVisualization.tsx
@@ -136,6 +136,9 @@ function VolcanoPlotViz(props: VisualizationProps<Options>) {
     plotContainerStyleOverrides,
   } = props;
 
+  // If no filters applied, the currentPlotFilters will be undefined. It should be an empty array.
+  visualization.descriptor.currentPlotFilters =
+    visualization.descriptor.currentPlotFilters ?? [];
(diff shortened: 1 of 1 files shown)


dmfalke:
I would try to modify this code https://github.com/VEuPathDB/web-monorepo/blob/467eb3b02e5147c044e54d85fe23e5942152521b/packages/libs/eda/src/lib/core/components/visualizations/VisualizationsContainer.tsx#L836

to something like this:

```ts
    if (!isEqual(filters ?? [], currentPlotFilters ?? [])) {
```

This way, we aren't mutating state that. Recall that state should almost always be treated as immutable, with few exceptions. This is especially true in React components and utility functions.

asizemore:
Thanks @dmfalke ! Done and done!

asizemore:
Updated with `main`. 
@dmfalke I totally forgot about this PR after the break. Do you mind giving it another look when you have a min please?

PR #1321, opened 2025-02-05
Add annotations to plotly plots

### LINKED ISSUE #1315: Phenotype data: Add annotations to scatterplot component
For the crispr phenotype plot, we'd like to replace the existing functionality of having two annotations on the plot. 

Image
(https://toxodb.org/toxo/app/record/gene/TGGT1_261490#category:phenomics)

Planning to use the [paper referenced annotations](https://plotly.com/javascript/text-and-annotations/) approach.

asizemore:
Resolves #1315 

Uses plotly's `layout.annotations` prop to add annotations to plots (see [examples](https://plotly.com/javascript/text-and-annotations/)). Examples included for `Scatterplot` and `Histogram`, but these annotations could be added to any plotly plot.

This PR does not add annotations for visx plots. I considered that as out of scope for the time being.


asizemore:
Ahh good point. While actually implementing the visx annotations might be out of scope, you're right that we should at least try to define a nice neutral interface that they could both use. I'll work on this!

dmfalke:
> Ahh good point. While actually implementing the visx annotations might be out of scope, you're right that we should at least try to define a nice neutral interface that they could both use. I'll work on this!

Cool! Let me know if you want to bounce ideas around. Something I'm wondering is, what is a reasonable approach to declaring _where_ an annotation will appear. for example, are the annotations for the crispr scatterplots based on min and max points? more generally, is this specific to the dataset type? do we need a crispr-specific scatterplot component to declare these annotations?

Anyway, just some random thoughts about this.

asizemore:
Thanks @dmfalke ! I just added a new type. Plotly and visx approach annotations pretty differently, so I attempted to keep the type minimal and let it expand in the future if we need. 

> Something I'm wondering is, what is a reasonable approach to declaring where an annotation will appear. for example, are the annotations for the crispr scatterplots based on min and max points?

There are two ways to define where an annotation goes (in plotly): using a fraction of the axis distance (plotly's 'paper' reference') or an actual point on the plot. The stories I put up have an example of each. But i think your question about a crispr-specific plot is a good one. We don't necessarily need one because we could have another attribute in john's table that declares the annotation position _as long as_ that position is with reference to the axes, not xmin and max. I believe that is the case for these plots - folks just want to see those annotations in the bottom left and top right, but I need to verify.

So we don't have to (at least not from what i see right now!) but of course we could!

asizemore:
To answer your earlier question, we do have a plot that uses annotations right now: the volcano visx plot. It has volcano-specific annotations already baked in, and the only thing it needs to know from the context is the text for those annotations. Our thinking was that all volcano plot should always have those annotations, so no need to worry about passing in more than just the annotation text (called the `comparisonLabels` prop).

dmfalke:
@asizemore overall, I think this looks great. I'm not sure if you saw this build error: https://github.com/VEuPathDB/web-monorepo/actions/runs/13252170333/job/36992249053?pr=1321.

dmfalke:
Once that build error is fixed, I think this can be merged.

asizemore:
oh i didn't! Glad you pointed it out!

Sounds good! I'll fix and merge!

PR #1324, opened 2025-02-07
only order variable tree entities alphabetically

dmfalke:
Only order entities alphabetically in the variable tree components.


dmfalke:
Thanks for the quick review!

PR #1326, opened 2025-02-17
replace record_overview attribute with react component on gene page

dmfalke:
This gives developers more flexibility with how content is rendered in the top section of the record page, including using table attributes for the overview properties.

### DIFF
--- a/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/records/GeneRecordClasses.GeneRecordClass.jsx
+++ b/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/records/GeneRecordClasses.GeneRecordClass.jsx
@@ -1,4 +1,4 @@
-import lodash, { memoize } from 'lodash';
+import lodash from 'lodash';
 import React, {
   Suspense,
   useCallback,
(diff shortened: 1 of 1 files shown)


asizemore:
I'd guess this isn't the only example of a template defined in the wdk that gets parsed and rendered. In what scenarios would you advise moving the template to react like you've done in this PR? Should we take the opportunity to do it whenever we need to make a change to a template, for example? Or only when there's no way around it? Somewhere in between?

dmfalke:
> I'd guess this isn't the only example of a template defined in the wdk that gets parsed and rendered. In what scenarios would you advise moving the template to react like you've done in this PR? Should we take the opportunity to do it whenever we need to make a change to a template, for example? Or only when there's no way around it? Somewhere in between?

This is a good question. It really comes down to what makes the most sense for the project, as a whole.

There are a few benefits for using textAttributes in the model:
1. They are familiar and relatively easy to add
2. The WDK java code handles parsing dependent attributes, and will run whatever attribute queries are needed to derive the final composite value.
3. They provide a nice encapsulation for the front end, which allows the front end to request a single attribute.
4. It works well with downloads, which do not use any client side rendering logic--in a textAttribute, you can define both a display string, and a plain text string. This might be the biggest reason to stay the course.

That being said, textAttributes are pretty limited in what they can do. For example, there are no conditional or looping structures available. In cases where this is needed, using a react component is the best solution. (However, there are time where this type of logic is implemented in the attribute query's sql :nauseated_face:). It's also worth noting that this approach will sometimes require a second attribute just for downloads, which can get confusing from a maintenance perspective.

Ultimately, it would be a lot of work to migrate all textAttributes to a react-based approach, and might introduce more complexity to the model.

I hope this helps.

PR #1327, opened 2025-02-17
Use wdk record page for study details for user datasets in eda

### LINKED ISSUE #1337: User datasets: Add new metadata fields to upload form
API for reference: https://veupathdb.github.io/vdi-service/vdi-api.html#resources:/vdi-datasets:post

1. Update the client code (the api client module). That means a) updating types, b) renaming types to VDI- and Legacy UserDatasetMeta, c) update usages of `userDatasetDetails` type
3. Update the upload form
4. Make an "edit" mode.
Important: this will be a hidden feature _only_ used by our staff. That makes 3 less important because our staff can deal with a rougher UI for awhile. 

Questions/comments
- Why is `createdOn` part of the api?

dmfalke:
This PR updates EDA to use the WDK record page for study details for analyses made from user datsets.

Resolves #1337


aurreco-uga (approved):

asizemore:
Types looking good. Can see expected properties in the console.

Screen Shot 2025-02-27 at 5 15 12 PM

asizemore:
@dmfalke before I get too far, what do you think of the Publications strategy? Basically i've made a `PublicationInput` component that we can add as many times as necessary. It seems to function well so far. I didn't put too much effort into styling but it might be good enough for now? Anyways, wanted to get a quick sanity check on this before moving too much further with extracting the styles to a css file and just generally cleaning up a bunch (i'm noticing we could benefit from more white space in multiple as well!). If this looks good to you, I'll do the same for the other new props where we're adding possibly multiple items in an array. 


https://github.com/user-attachments/assets/b765fd2b-7231-408f-b115-0b2b169d7e5a

dmfalke:
I think this looks really nice! I would plan to move forward with this approach.

asizemore:
Looking good, all the new fields are there. I didn't add `createdOn` because that seemed like something that a user shouldn't be able to set... Anyways, I think I can consolidate and organize the code a little better, and there's one weird behavior i haven't figured out yet - When trying to remove a contact that precedes an empty contact, the intended contact will get removed in the object, but the ui still shows the non-empty contact on the page. See last little bit of video.


https://github.com/user-attachments/assets/f419649a-2ded-45e6-ad60-c654c7cd76da

asizemore:
I investigated adding a function to remove empty nested input elements today. Turns out the form won't submit with those empty elements because they contain fields marked as required! So a user actually has to go back and delete the empty ones. That seems like a reasonable ui to me, but curious what others think as they play with it.

bobular:
I tested this and the mechanics look great. Haven't looked at the code yet.

In UX terms, I think it needs a lot more guidance for users.
One trivial thing is the placeholder text for Publications->Citation should be `Citation (e.g. Jones et al., 1999)`

Explanations (or examples) for the Short Name and Short Attribution would be good. And category obviously, but that may be a hidden form parameter - I know work is ongoing regarding this.

It doesn't say anywhere that `*` is "required" - not a biggie, and very obvious, but a small note just before the submit button?

bobular:
While reviewing I DRYed up the types a little bit ([commit](https://github.com/VEuPathDB/web-monorepo/pull/1327/commits/38dc7e563d94796883d368f0cac54e4932f02e54)) 

I think we can go a bit further and consolidate the io-ts objects with the types, but I haven't tried yet.

asizemore:
@bobular you're likely right we need more guidance. However, at this point the feature is only for internal use only, so I wouldn't mind getting some feedback on it before we put too much effort in that direction. 

Thanks for the commit btw!! I like the cleaned up types 😎

aurreco-uga:
sorry i had not looked at the thread here
i had posted this in slack
https://epvb.slack.com/archives/C012ZK4Q5CZ/p1742430720306269
basically we should remove the first 3 fields, they are just for internal use. 
and you are correct createdOn  not needed either;  ellie posted to my question:
"createdOn is an rfc3339 timestamp, VDI sets that value when it creates the dataset.  I don't believe it is stored in the app db"

asizemore:
Types fixed @bobular !

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -449,6 +501,287 @@ function UploadForm({
...
+                });
+                return (
+                  <PublicationInput
+                    n={index}
+                    pubMedId={publication.pubMedId}
+                    onAddPubmedId={(value: string) => {
+                      const updatedPublications = updatePublicationsObject(
+                        value,
+                        'pubMedId'
+                      ) as UserDatasetPublication[];
```
bobular, line 524: If you use the generic <T> type and function above then all the `as XXXXX[]` assertions here and below are not necessary :-)

```suggestion
                      const updatedPublications = updatePublicationsObject(
                        value,
                        'pubMedId'
                      );
```

packages/sites/genomics-site/webapp/wdkCustomization/js/client/userDatasetRoutes.tsx
```diff
@@ -19,7 +19,7 @@ const UserDatasetRouter = React.lazy(
 
 const availableUploadTypes = userDatasetsUploadTypes.trim()
   ? userDatasetsUploadTypes.split(/,\s*/g)
-  : ['genelist', 'bigwigfiles', 'rnaseq'];
+  : ['genelist', 'bigwigfiles', 'rnaseq', 'wrangler'];
```
bobular, line 22: I'm not so keen on `wrangler` being user-facing, albeit in a URL. Is this already set in stone or can we replace it?

I see that it might be a conifer thing (`userDatasetsUploadTypes` defined in packages/libs/web-common/src/config.ts)

So these are the default fallbacks?

I would call this `study` or `generic` or something? Can discuss tomorrow.

bobular (approved):
Had another look. I strongly suggest the typing improvements below - will shorten and make a bit more readable. If it compiles and works afterwards, then go ahead and merge, if appropriate!

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -133,6 +158,20 @@ function UploadForm({
   const [description, setDescription] = useState(
     urlParams.datasetDescription ?? ''
   );
+  const [shortName, setShortName] = useState(urlParams.datasetShortName ?? '');
+  const [shortAttribution, setShortAttribution] = useState(
+    urlParams.datasetShortAttribution ?? ''
+  );
+  const [category, setCategory] = useState(urlParams.datasetCategory ?? '');
```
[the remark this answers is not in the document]
asizemore: Yes! In fact they aren't used at all in the form anymore and my understanding is they're just set on the back end so we can ignore them. I've deleted these items from the rest of the file, too.

PR #1328, opened 2025-02-18
Add ability to highlight points to scatterplot

### LINKED ISSUE #1322: Allow scatterplot to highlight points
Requested for the phenotype data but generally useful. 

The goal is to let `ScatterPlot` take a new argument `highlightedPoints`, that is an array of point IDs. 

How will this interact with an overlay variable? I propose it will override the overlay. Other options exist and may be explored.

asizemore:
Resolves #1322 

This was a little funky because plotly already allows for highlighting - we just have to set the color appropriately in the trace! So this PR is more of a hijack of our current processing of scatterplot data to recolor the points.

A few considerations:
- Should the backend just break up the data into highlight and non-highlight for us? It could, and would basically add another "overlay"-type column, but the issue is when we highlight a scatterplot with anything that we have calculated. Smoothed means, best fit line, etc. Breaking up the data for highlgihting would move these points to a place that doesn't correspond with their place in a calculation. We don't want a highlight-smoothed mean and a non-highlight-smoothed mean, we would just want one smoothed mean with a few points highlighted. But we still want to know those highlighted points contributed to the calculation of the smoothed mean. This gets extra complicated with overlays and facets.
- On the frontend, should we recolor in place or make a new trace? Make a new trace so that there's an extra line in the legend that we can click on and off
- Where should the highlighting go on the frontend? Ideally in `Scatterplot` but that means we have to traverse the entire dataset _again_. We already traverse the dataset in `processInputData`, so we can add it there. Also we want to do it _after_ the points get their normal color (given by overlay, etc), so that in the case where we want the highlighted points to be a different shape instead of color (for example), we still have the opportunity to keep the points' original colors. 

I ran into a bit of trouble with tooltips. Despite my best efforts, I couldn't hide the tooltips of the original data. I tried changing marker sizes, colors, lines, everything, but the tooltips still appeared. I finally found [an option](https://codepen.io/etpinard/pen/VwZgjom?editors=1010) that I don't like because it removes some data, but at least the whole point isn't deleted from the trace. 

Remaining To Dos:

- [ ] General scatterplot testing to make sure nothing broke
- [x] Add prop that gives the highlight trace a specific name (would show up in the legend and hover text)

### DIFF
--- a/packages/libs/components/src/stories/plots/ScatterPlot.stories.tsx
+++ b/packages/libs/components/src/stories/plots/ScatterPlot.stories.tsx
@@ -474,7 +474,8 @@ const { dataSetProcess: dataSetCategoricalOverlayHighlight } = processInputData(
   dependentValueType,
   false,
   highlightPointIdsStyled,
-  highlightStyleOverride
+  highlightStyleOverride,
+  'My Highlight Trace'
 );
 
 export const HighlightPointsWithStyleOverride: Story<ScatterPlotProps> =
--- a/packages/libs/components/src/stories/plots/ScatterPlot.stories.tsx
+++ b/packages/libs/components/src/stories/plots/ScatterPlot.stories.tsx
@@ -453,7 +453,11 @@ HighlightPoints.args = {
   displayLegend: true,
 };
 
-// Highlight points with a specialized style
+// Highlight points with a specialized style.
+// This story uses data that's already been broken up by an overlay variable. The result
+// currently looks a bit ridiculuous - we have a bunch of gray Data rows in the legend. However, this is actually
+// the desired behavior. In practice each of those gray Data traces would have a name, so we'd see helpful group names. Additionally,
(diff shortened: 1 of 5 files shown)


asizemore:
Update: related backend PRs have been merged and are now deployed to QA.

asizemore:
Ready for a look! 
To test with highlights, use microbiome DailyBaby scatterplot, and have at least one of the axes from the 16S entity. For example 
Screen Shot 2025-02-20 at 9 58 38 AM

After approval, I'll remove lines 163 and 2217 and test scatterplots again. 

(This was hard for me to remember 😆 ) The goal is to have a solid `processInputData` function that does the highlighting so that our genomics components can use it, not for a fully functioning highlight scatterplot eda viz. So As long as we get highlights in the mbio scatter with our stubbed in point ids, we can be happy. Otherwise we'd be implmementing a lot of new logic in the eda scatterplot for a feature that won't be in the eda for some time.

dmfalke (approved):
@asizemore this looks great! I say merge!

PR #1342, opened 2025-03-19
bluesky icon

aurreco-uga:
i used a png and added the image in Site/htdocs where we have footer images.
I could work on using one of the images i added in monorepo, maybe svg instead.
This was just faster for me to get something up.
https://aurreco.plasmodb.org/plasmo.aurreco/app

Would be nice to upgrade fontawesome to the latest so we could use their new icons (our version 4 doesn't have the bluesky icon or the X)

To see this working you ll need the latest on EbrcWebsiteCommon:
https://github.com/VEuPathDB/EbrcWebsiteCommon/commit/a146bf1a422842a8b92c7d45c30140cd39018284

There is also a request to make the 3 right most icons bigger, of similar size as X and bluesky..
https://epvb.slack.com/archives/CBE9KRM0T/p1742415204385959?thread_ts=1741779618.262289&cid=CBE9KRM0T
I might address on a later commit.


asizemore:
Why is there an svg version added if we're only using the png? 

> Would be nice to upgrade fontawesome to the latest so we could use their new icons 

Agreed! Let's make a new issue for this

> There is also a request to make the 3 right most icons bigger, of similar size as X and bluesky..

I also agree that would be good for a different PR

asizemore (approved):
I didn't test it myself but i read through the changes and looked at your dev site. The new icon looks nice!

PR #1343, opened 2025-03-19
Fix race condition in useSetter() by improving state management

bobular:
This solves the "new visualization" bug in the EDA on qa.clinepidb.org (since eda-in-wdk updates), which gave an error in the client:

image

This PR fixes a **race condition** in `useSetter()` that occurred due to its reliance on a shadow/ref state of `analysis`. Specifically, `setComputation()` wasn’t updating quickly enough when switching to a new visualization.

To resolve this, I refactored state management to use the **functional update form** of `setAnalysis()` (i.e., `(prev) => { ... }`). This eliminates the need to manually pass `analysis` around, ensuring updates always use the latest state.

The original use of `useRef` was presumably to prevent `analysis` from being a dependency in `useCallback()` inside `useSetter()`, since adding it caused infinite state updates. This refactor removes that issue while keeping the function stable.

(Note: there was a quicker fix using `useLayoutEffect()` in `useSetter()` but this might have had more unintended consequences than this seemingly more complex fix.)


aurreco-uga (approved):
not familiar with the code so ill just approve

bobular:
Seems like some build issues that didn't come up in testing/compiling in packages/lib/eda alone.

Will address!

bobular:
Seems that the toxo CRISPR phenotype subsetting is broken because of this?

Need to do more testing.

bobular:
No the phenotype subsetting stopped working because of my "fix" to the clinepi bug.

asizemore (approved):

PR #1345, opened 2025-03-20
update default highlight marker styles

### LINKED ISSUE #1334: Highlighting default colors
We now have a great example

Image

And can use this to choose a better pair of colors. That gray is about as light as we can make it while keeping it accessible, but we can tweak the purple so that it stands out better.

asizemore:
Resolves #1334 

So. In order to hit accessibility standards, we can't lighten the gray. The _best_ solution is to put an outline around the highlighted markers (and I updated their colors a bit). Adding an outline is a better solution because it retains accessibility by preventing our use of color "alone" for distinguishing points (read more [here](https://observablehq.com/@frankelavsky/chartability-contrast-series)). Frankly this "no color alone" thing eats at other aspects of our plots like our overlay colors and so on, but that's a topic for another day.

This outline approach works perfectly in storybook.

However, when we get to either the real eda or the phenotype graph, we lose the marker strokes. I'm not yet sure where they go. 

If I can't figure it out quickly, I'll just update the colors to make the gray lighter (sad for accessibility but it won't be too bad) and make a new issue for the outlines.


asizemore:
Since the outline wasn't working, I fell back to the easier option of lightening the gray and increasing the point size. This is not the ideal option because the gray now falls below accessibility standards for contrast, but the whole plot is still better than it was in terms of portraying the important information.

Screen Shot 2025-03-21 at 8 38 16 AM


I'll circle back a bit later and see if I can figure out the outline thing. That could pose issues later.

asizemore:
If folks would like to test, use qa toxo and visit `a/app/record/gene/TGGT1_261490#EdaPhenotypeGraphs`.

aurreco-uga (approved):
i think the highlighted dot could be even bigger if needed.

PR #1346, opened 2025-03-20
Add bluesky and linkedin properties to local dev site configs

bobular:
Sorry for the late review of https://github.com/VEuPathDB/web-monorepo/pull/1342

I tested in a local `genomics-site` and found that both the bluesky and linkedin variables needed setting there too.

This is of no importance to rolling out to qa!


asizemore (approved):
Thanks for the quick update!

aurreco-uga (approved):

PR #1350, opened 2025-03-24
Factor out a new useAnalysisChangeHandler hook

bobular:


asizemore (approved):
Still trying to get the concepts to sink in, but changes made sense and testing in genomics and regular eda went smoothly.

PR #1355, opened 2025-03-28
App scatterplots should not return point Ids

asizemore:
With https://github.com/VEuPathDB/service-eda/pull/87, fixes the issue seen on QA with scatterplots.

Uses `options` to have apps tell the scatterplot viz not to request point Ids. Tested with the local backend on the above branch.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
@@ -286,6 +286,7 @@ interface Options
   ): VariableDescriptor | VariableCollectionDescriptor | undefined;
   hideTrendlines?: boolean;
   hideLogScale?: boolean;
+  returnPointIds?: boolean; // for ScatterplotRequestParams
 }
 
 function ScatterplotViz(props: VisualizationProps<Options>) {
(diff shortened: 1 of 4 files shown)


packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -752,7 +753,8 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
             ? [vizConfig.facetVariable]
             : undefined,
           showMissingness: vizConfig.showMissingness ? 'TRUE' : 'FALSE',
-          returnPointIds: vizConfig.returnPointIds ?? true,
+          returnPointIds:
+            options?.returnPointIds ?? vizConfig.returnPointIds ?? true,
```
bobular, line 757: Is this really something we want to store/control via `vizConfig`?

I think `vizConfig` should only be for visualisation state that the user can control.

I think
```suggestion
          returnPointIds:
            options?.returnPointIds ?? true,
```
should be fine (and then remove `returnPointIds` from `ScatterplotConfig`)?

bobular (commented):
Looks good, but see my question about `vizConfig`

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
asizemore: I'm happy with that rule! I'll make the changes to `ScatterplotConfig`.

asizemore:
Done and done!

bobular (approved):
Great - assuming it works (and the map) I am very approving!

PR #1357, opened 2025-04-01
set 2 decimals

aurreco-uga:
issue in toxo new eda histogram filter parameter

Screenshot 2025-04-01 at 3 28 44 PM

### DIFF
--- a/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
+++ b/packages/libs/eda/src/lib/core/components/filter/HistogramFilter.tsx
@@ -505,7 +505,9 @@ function HistogramPlotWithControls({
 
   const selectedRange = useMemo((): NumberOrDateRange | undefined => {
     if (filter == null) return;
-    return { min: filter.min, max: filter.max } as NumberOrDateRange;
+    const mymin = Number(filter.min).toFixed(2);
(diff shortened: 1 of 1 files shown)


bobular (commented):
Thanks for having a look. I'll have a think about this. Here we have truncated the values as they come "up" from the saved analysis config (user database) - so they are truncated for display only.  This might be the way to do it, but we can also do it on the way "down".

I'm also hesitant to truncate to two decimal places when a study might have variables with genuinely small numbers. `toPrecision(4)` will probably be more appropriate, but this can have the downside of aggressively converting things to scientific (exponential) notation.

I'll have a play with

aurreco-uga:
nupur said it is a problem in clinepi, and it does seem easy to generate this in clinepi

Screenshot 2025-04-01 at 3 32 35 PM

bobular:
I think it's possibly the `binWidth` coming from the back end. This is the `Temperature` variable from NECTAR-1 (clinepi)

```
{
	"distributionDefaults": {
		"rangeMin": 34,
		"rangeMax": 37.7999992370605,
		"binWidth": 0.25333328247
	}
}
```

bobular:
Not going to merge this, see #1358

PR #1361, opened 2025-04-01
Update for the phenotype graph legend.

### LINKED ISSUE #1360: Change phenotype graph legend to say "All Genes"
Based on feedback from Omar.

asizemore:
Resolves #1360 

This is not the best fix ever, but will somewhat scalable. At least we won't need to update it until we have some more use cases to think about.

All the highlighting information was getting a bit unwieldy so I added a new type `HighlightedPointsDetails`. This modification is responsible for most of the changes.

@bobular I couldn't install the pluralize package so that we can have better (real) pluralization. Do you have the authority to do so?


asizemore:
Pluralize package now added. Ready for a look!

aurreco-uga (approved):

PR #1366, opened 2025-04-09
Feature flag for community datasets

bobular:


packages/libs/coreui/src/components/inputs/SingleSelect.tsx
bobular, line 1: These are changes I made to honour the `disabled` prop on the `Item` object passed to `SingleSelect`.

Before this change, if you set `disabled: true` on an item it was ignored completely, even though the typing suggested functionality.

In the end we have not used this functionality in this PR but now we have it for next time.

packages/libs/eda/src/index.tsx
```diff
@@ -68,6 +68,9 @@ const singleAppMode = process.env.REACT_APP_SINGLE_APP_MODE;
 const showUnreleasedData =
   process.env.REACT_APP_SHOW_UNRELEASED_DATA === 'true';
 
+const enablePublicUserDatasets =
+  process.env.COMMUNITY_DATASETS_ENABLED === 'true';
```
bobular, line 72: Developers will need 
```
COMMUNITY_DATASETS_ENABLED=true
```
in their `packages/libs/eda/.env.local` to pass the flag through to the map app. Not sure what it does there, however...

packages/libs/coreui/src/components/inputs/SingleSelect.tsx
bobular, line 1: Ignore the changes in this file. I originally implemented greyed-out item functionality in the dropdown list, but then reverted it.

I left the implementation part in because there was already an `item.disabled` prop that suggested this feature should exist! See [here](https://github.com/VEuPathDB/web-monorepo/blob/7232fe5b6d2400a8f97c37efc30de22068a2f6bb/packages/libs/coreui/src/components/inputs/checkboxes/CheckboxList.tsx#L76).

ryanrdoherty (approved):

PR #1367, opened 2025-04-11
Conditionally disable AI expression summary and nav menu item if no WDK reporter available

bobular:
Hi @dmfalke - would you mind taking a look at this?

We decided not to have a front end feature flag, but instead switch the feature on/off based on the availability of the back end Reporter.

I found a reasonably elegant way to disable the left hand navigation item. <s>Is the `return null` the only way to disable the actual section?</s> Both disabled by recordStoreModule upon loading.


dmfalke (commented):
Just looking at the code, I would expect the changes to the record store module to be all that is needed. This should cause the attribute to not appear in the left nav section, nor in the main content section. Does that answer your question?

bobular:
Merging this into `ai-expression` (I've tested it!)

Issue #1303, opened 2024-12-18
Ortho 7 - group table filters: performance, bug and UX


bobular:
### performance

Chrome dev tools performance tab has a 6x slower CPU throttling - use it to measure the "page load" on the biggest group: https://beta.orthomcl.org/orthomcl.b7/app/record/group/OG7_0000011

When it's done you can interact with the page, but with 6x slower CPU :-)
Try clicking the second level checkboxes in the Organism filter - every 1/2 second or so.
You should see that there are major delays to updates - replicating what our colleague found.

Try removing domain architecture column to see if that makes a difference.

500, 250 rows instead? Or even 20.

Filtering logic can be improved with Set lookup for organisms (`selectedSpecies`) and proteins (`proteinFilterIds`)

Look into windowing libs (only render the rows you need) - could be a heavy lift to incorporate into Mesa...

### bug

Under certain circumstances, particularly with large groups, after interacting with filter popovers, the page is inactive because the popover backdrop (it's basically a modal) has not been removed. The popover may still be there as a "ghost" and you can click on parts of it.

Let's try updating MUI to see if that helps.

### UX

Due to the delays described above there is some confusion about how to dismiss the popover dialogues and whether or not this is what activates the filter.

If the performance issues are fixed, maybe we can get away with just having an `x` top right to dismiss the filter popovers (in addition to clicking elsewhere, and pressing ESC.

Another suggestion was to have `cancel` and `apply` buttons, which would be a bit more work and require further consultation before investing in that.

Having sticky items at the top of the popover would be useful in either case, e.g.
![image](https://github.com/user-attachments/assets/8c28e74d-ebaf-4943-8f55-29280e3a77fa)

bobular:
### UX update

Dave suggested that we need visual feedback for when the table is not yet updated upon filtering.

PR #1369, opened 2025-04-17
Ortho ghost overlay fixes

### LINKED ISSUE #1280: Ortho 7 - beta bugs

### LINKED ISSUE #1370: Ortho 7 - tree/table not always in alignment and associated text filter bug
This is from https://qa.orthomcl.org/orthomcl.b6_22/app/record/group/OG7_0000011#Sequences

Image

Note that there is an additional table entry between the two highlighted leaf nodes.

The `:RNA` in the sequence's `full_id` hasn't been processed properly (everything after the colon is discarded in patristic tree node IDs)

Also if you text-search the table for `:RNA` you get a runtime error!

bobular:
# Ready for review

Resolves https://github.com/VEuPathDB/web-monorepo/issues/1303

Resolves #1280

Resolves #1370 

<s>Implement dynamic Popover transition durations—enter fixed at 300 ms, exit set to Math.max(300, numSequences/10)—so that even with very large groups (e.g. 2.6 s for 26,000 sequences) the popover buttons go away cleanly without leaving behind ghost overlays that block the UI. (Main thread compute was disrupting the animation event processing. The extra animation time effectively waits for the compute to be done.) Groups <= 3,000 proteins won't see any UI behaviour differences at all.</s>

NEW APPROACH: figure out when the component `isFiltering` from `(realState !== volatileState)` and then do several things:
1. grey out the main table and overlay some LOADING text (our default `Loading` spinner doesn't work due to hardcoded 200ms delay)
2. tell the Popover components that they should `deferClosing` while `isFiltering`
   - while Popovers are deferring closing they are greyed out and unclickable

Amazingly it works very nicely.

Filtering logic was also optimized by using Sets for membership checks, ensuring responsive performance on large groups.

Note that the ClinEpi popover behaviour should be unchanged but I need to 

- [x] test ClinEpi

Here is the fix evidence for #1370

image

### DIFF
--- a/packages/libs/coreui/src/components/buttons/PopoverButton/PopoverButton.tsx
+++ b/packages/libs/coreui/src/components/buttons/PopoverButton/PopoverButton.tsx
@@ -7,12 +7,13 @@ import {
   useImperativeHandle,
   useCallback,
 } from 'react';
-import { Popover, PopoverProps } from '@material-ui/core';
+import { Popover } from '@material-ui/core';
 import SwissArmyButton from '../SwissArmyButton';
 import { gray } from '../../../definitions/colors';
 import { ButtonStyleSpec, PartialButtonStyleSpec } from '..';
 import { ArrowDown } from '../../icons';
 import { merge } from 'lodash';
+import { Dimmable } from '../../containers';
 
 // The default styles largely mimic the style of the MUI button was previously used in this component
 const defaultStyle: ButtonStyleSpec = {
@@ -89,18 +90,15 @@ export interface PopoverButtonProps {
   /** Allows for additional cleanup when popover closes */
   onClose?: () => void;
 
+  /** Don't let popover close while this is true */
+  deferClosing?: boolean;
+
   /** Used in SingleSelect component to set focus when popover opens */
   setIsPopoverOpen?: (isOpen: boolean) => void;
 
   isDisabled?: boolean;
 
   styleOverrides?: PartialButtonStyleSpec;
-
-  /** Optional to provide animated appear/disappear
-   * provide either an integer milliseconds (appear and disappear)
-   * or an object with separate timings: { enter: 300, exit: 600 }
-   */
-  transitionDuration?: PopoverProps['transitionDuration'];
 }
 
 /**
@@ -116,9 +114,10 @@ const PopoverButton = forwardRef<PopoverButtonHandle, PopoverButtonProps>(
       setIsPopoverOpen,
       isDisabled = false,
       styleOverrides = {},
-      transitionDuration,
+      deferClosing,
     } = props;
     const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
+    const [isClosePending, setIsClosePending] = useState(false);
 
     const finalStyle = useMemo(
       () => merge({}, defaultStyle, styleOverrides),
@@ -126,20 +125,32 @@ const PopoverButton = forwardRef<PopoverButtonHandle, PopoverButtonProps>(
     );
 
     const onCloseHandler = useCallback(() => {
+      if (deferClosing) {
+        setIsClosePending(true);
+        return;
+      }
       setTimeout(() => {
         anchorEl?.focus(); // return focus to button
       });
       setAnchorEl(null);
       onClose && onClose();
-    }, [anchorEl, onClose]);
+    }, [anchorEl, onClose, deferClosing]);
 
-    // Expose the `close()` method to external components via ref
+    // mop up any deferred closing that needs to be done
+    useEffect(() => {
+      if (isClosePending && !deferClosing) {
+        setIsClosePending(false);
+        onCloseHandler();
+      }
+    }, [isClosePending, deferClosing, onCloseHandler]);
+
+    // Expose a `close()` method to external components via ref
     useImperativeHandle(
       ref,
       () => ({
-        close: onCloseHandler,
+        close: () => setIsClosePending(true),
       }),
-      [onCloseHandler]
+      []
     );
 
     useEffect(() => {
@@ -167,9 +178,10 @@ const PopoverButton = forwardRef<PopoverButtonHandle, PopoverButtonProps>(
           horizontal: 'left',
         }}
         keepMounted
-        transitionDuration={transitionDuration}
       >
-        {children}
+        <Dimmable dimmed={isClosePending} blocking fullscreen>
+          {children}
+        </Dimmable>
       </Popover>
     );
 
--- a/packages/libs/coreui/src/components/buttons/PopoverButton/PopoverButton.tsx
+++ b/packages/libs/coreui/src/components/buttons/PopoverButton/PopoverButton.tsx
@@ -7,7 +7,7 @@ import {
   useImperativeHandle,
   useCallback,
 } from 'react';
-import { Popover } from '@material-ui/core';
+import { Popover, PopoverProps } from '@material-ui/core';
 import SwissArmyButton from '../SwissArmyButton';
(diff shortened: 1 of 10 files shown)


packages/libs/coreui/src/components/buttons/PopoverButton/PopoverButton.tsx
bobular, line 1: Just prop passing

packages/libs/coreui/src/components/inputs/SelectList.tsx
```diff
@@ -30,31 +37,36 @@ export default function SelectList<T extends string>({
...
   const [selected, setSelected] = useState<SelectListProps<T>['value']>(value);
   const [buttonDisplayContent, setButtonDisplayContent] = useState<ReactNode>(
     getDisplayContent(value, items, defaultButtonDisplayContent)
   );
+  const [isOpen, setIsOpen] = useState(false);
 
-  const onClose = () => {
-    onChange(selected);
+  const onClose = useCallback(() => {
+    if (!instantUpdate) onChange(selected);
```
bobular, line 50: Added this to be symmetrical with `handleCheckboxListUpdate`. No need to update on closing if instant updates were already made. This was probably a harmless oversight.

packages/libs/coreui/src/components/inputs/SelectList.tsx
```diff
@@ -30,31 +37,36 @@ export default function SelectList<T extends string>({
...
+  }, [instantUpdate, selected, items, defaultButtonDisplayContent, onChange]);
 
   /**
    * Keep caller up to date with any selection changes, if required by `instantUpdate`
    */
   const handleCheckboxListUpdate = useCallback(
     (newSelection: SelectListProps<T>['value']) => {
+      // only allow updates while open
+      // seems obvious, but animated transitions blur the lines
+      if (!isOpen) return;
```
bobular, line 63: Here we stop any selection updates being made while the popover is closed. This is possible during animations or when the ghost popover component remains active (but not visible) along with the persistent backdrop.

packages/libs/coreui/src/components/inputs/SelectTree/SelectTree.tsx
```diff
@@ -50,6 +61,15 @@ function SelectTree<T>(props: SelectTreeProps<T>) {
     onSelectionChange(localSelectedList);
   }, [onSelectionChange, localSelectedList]);
 
+  // only used when there's a popover
+  // prevents selections being made while popover is transitioning away
+  const handleUpdateWithPopover = useCallback(
+    (update: string[]) => {
+      if (isPopoverOpen) setLocalSelectedList(update);
```
bobular, line 68: Same as with SelectList - don't allow selection changes when popover is closed.

packages/libs/coreui/src/components/inputs/SelectTree/SelectTree.tsx
```diff
@@ -94,7 +114,9 @@ function SelectTree<T>(props: SelectTreeProps<T>) {
       customCheckboxes={props.customCheckboxes}
       isMultiPick={props.isMultiPick}
       name={props.name}
-      onSelectionChange={setLocalSelectedList}
+      onSelectionChange={
+        hasPopoverButton ? handleUpdateWithPopover : setLocalSelectedList
+      }
```
bobular, line 119: We only need the fancy handler when there's a popover button

packages/sites/ortho-site/webapp/wdkCustomization/js/client/components/phyletic-distribution/PhyleticDistributionCheckbox.tsx
bobular, line 1: Just prop passing

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/RecordTable_TaxonCounts_Filter.tsx
bobular, line 1: More prop passing

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -97,6 +99,11 @@ export function RecordTable_Sequences(
 
   const numSequences = mesaRows.length;
 
+  const popoverButtonTransitionDuration = {
+    enter: 300,
+    exit: Math.max(300, numSequences / 10),
+  };
+
```
bobular, line 106: Here's the dynamic animation time calc.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -224,11 +231,16 @@ export function RecordTable_Sequences(
       selectedSpecies.length > 0 ||
       proteinFilterIds.length > 0
     ) {
+      // these two are likely to be selected in large numbers
```
bobular, line 234: Below are some filtering performance mods

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -276,7 +286,8 @@ export function RecordTable_Sequences(
       leaves == null ||
       tree == null ||
       filteredRows == null ||
-      filteredRows.length === 0
+      filteredRows.length === 0 ||
+      filteredRows.length > MAX_SEQUENCES_FOR_TREE
```
bobular, line 290: We were filtering the tree even when we're not showing it. Duh! Not any more.

Probably need to add `< MIN_SEQUENCES_FOR_TREE` also.

packages/sites/ortho-site/webapp/wdkCustomization/js/client/records/Sequences.tsx
```diff
@@ -331,12 +342,17 @@ export function RecordTable_Sequences(
     [mesaColumns]
   );
 
-  const handleSpeciesSelection = useCallback(
-    (species: string[]) => {
-      setSelectedSpecies(species);
+  const handleSpeciesSelection = useCallback((species: string[]) => {
```
bobular, line 345: memoised a couple of handlers, just in case (they are hook dependencies downstream)

bobular (commented):
Just added a few comments to aid review.

aurreco-uga (commented):
i cant reproduce the issue and the fix makes things slower.. but if if we could confirm this is a problem for overseas users i think it is a fine fix..

bobular:
> i cant reproduce the issue and the fix makes things slower.. on big groups. on small groups makes no difference if this is a problem for overseas users i think it is a fine fix..

Thanks @aurreco-uga - actually this isn't a problem related to network speed. It's more to do with client CPU. The slow-down for big groups manifests in the form of slow animations, making it impossible to click quickly between filter buttons, but at least the UI doesn't freeze up completely.

It's not ideal but the only other solution is to use "web workers" which would require a lot more engineering and potentially new bugs.

asizemore:
Code looks good! Will test on Monday!

asizemore:
The tree seemed well-aligned and the funky pfam filter looks all better. The animation to prevent the ghost popover is clever. I had some issues with the popover waiting a few moments before fading out, but i think it's probably better than it was so i'm happy to approve.

asizemore (approved):

PR #1372, opened 2025-04-21
Updates to user upload form

### LINKED ISSUE #1371: DIY: UI Updates to user upload (steves feedback)
Feedback:

1.  make links colors the same we use in genomic sites (see Org Prefs on the subheader)
2. remove "optional" since we use a star to show what is required.
3. say "Related organisms". 
4. Hide organisms for mbio and clinepi
5. steve suggested to add the 3 or 4 new fields same as Summary and Description, no background, not hidden under "Additional details". Not sure what Omar/Danica would prefer.

asizemore:
Resolves #1371 

This PR currently implements all the styling changes mentioned in the issue. 

Note the colors are tough to "match". The outlined buttons I have slightly lighter than the "#006699" color of the organism preference link, because the thickness of the font and outline make it look darker. The Remove buttons all have the "#069" color, so you can see what that looks like.

Screen Shot 2025-04-21 at 10 47 45 AM

Current status - not ready for code review but worth a style check to see if you like the updates, @aurreco-uga !

Note - I made some changes to `coreui`. Please run `yarn build-npm-modules` in the `coreui` directory before attempting to run this branch!


packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -490,143 +492,144 @@ function UploadForm({
...
+                    setPublications(updatedPublications);
+                  }}
+                  citation={publication.citation}
+                  onAddCitation={(value: string) => {
+                    const updatedPublications = updatePublicationsObject(
+                      value,
+                      'citation'
+                    );
+                    setPublications(updatedPublications);
+                  }}
```
bobular, line 530: This could be made a little more streamlined if we made the `createNestedInputUpdater` factory take a `setState()` updater (from `useState`) instead of the actual publications list, and use its functional update mode (`setState((prev) => modifySomehow(prev))`)

The goal would be to replace these lines with just this:
```
  onAddCitation={(value: string) => updatePublications(value, 'citation')}
```

Here's an _untested_ implementation of the setter-based factory from our friendly assistant. It also implemented the `enforceExclusiveTrue` that I had asked about a good name for that prop.

```
const createNestedInputUpdater = function <T>(
  props: {
    index: number;
    setNestedInputObject: React.Dispatch<React.SetStateAction<T[]>>;
    enforceExclusiveTrue?: boolean;
  }
) {
  const { index, setNestedInputObject, enforceExclusiveTrue } = props;

  return function (newValue: string | boolean, inputName: string) {
    setNestedInputObject((prev) => {
      const updated = [...prev];

      if (enforceExclusiveTrue && newValue === true) {
        updated.forEach((item, i) => {
          if (i !== index) {
            updated[i] = { ...item, [inputName]: false };
          }
        });
      }

      updated[index] = { ...updated[index], [inputName]: newValue };
      return updated;
    });
  };
};
```

Creating the update in the `map((publication, index) => ...)` would be like this
```
const updatePublications = createNestedInputUpdater({
  setNestedInputObject: setPublications,
  index,
  enforceExclusiveTrue: true,
});
```

If you needed an exclusive boolean setter and a non-exclusive boolean setter, you'd need to make two updaters: `updatePublications` and `updatePublicationsExclusiveTrue` or similar. I don't think there's a use case for that right now though. Or we make `exclusiveTrue` an arg to the updater, not the factory.

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -490,143 +492,144 @@ function UploadForm({
...
+                      value,
+                      'description'
+                    );
+                    setHyperlinks(updatedHyperlinks);
+                  }}
+                  isPublication={hyperlinks[index]?.isPublication}
+                  onAddIsPublication={(value: boolean) => {
+                    const updatedHyperlinks = updateHyperlinksObject(
+                      value,
+                      'publication'
```
bobular, line 598:
```suggestion
                      'isPublication'
```

This was breaking the radio button.

bobular: This would have been caught by TypeScript if we added `inputName: keyof T` to the factory. (This is the original.)

```
const createNestedInputUpdater = function <T>(
  props: AddNestedInputChildProps<T>
) {
  const { nestedInputObject, index } = props;

  return function (newValue: string | boolean, inputName: keyof T) {
    const updatedNestedInputObject = [...nestedInputObject];
    updatedNestedInputObject[index] = {
      ...nestedInputObject[index],
      [inputName]: newValue,
    };
    return updatedNestedInputObject;
  };
};
```

packages/libs/user-datasets/src/lib/Components/UploadForm.scss
```diff
@@ -38,7 +38,6 @@
     display: flex;
     flex-direction: column;
     gap: 0.5em;
-    padding-left: 1em;
     padding-right: 1em;
```
bobular, line 41: Maybe remove this padding-right, so the textarea boxes' right edges line up with the input group borders?

bobular (changes requested):
I think the colours are fine! What bugs me more is the change in outline style when you go from Name/Summary/Description to Publications/Hyperlinks etc.  I know these are different elements: textarea vs. outline of input group.  Maybe just remove the padding-right as commented below, so they line up nicely?

Functionally, I wonder if we can make it impossible to select two primary contacts (for example). Yes it looks possible. I think we'd need a third optional prop to the `createNestedInputUpdater` factory function called `enforceExclusiveTrue`. This would, when a true boolean value is set for a given input name, automatically set that field to false for all other items in the array.

I'm also wondering if we can make it so you can navigate to and from the help tab without losing one's work. No, that's tricky - see #1394

The style/DRY-up suggestions are optional.

The uniqueness of the contacts' `isPrimary` I think is a required behaviour?

The fix for `isPublication` is needed of course.

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -651,124 +654,121 @@ function UploadForm({
...
+                    setContacts(updatedContacts);
+                  }}
+                  isPrimary={contact.isPrimary}
+                  onAddIsPrimary={(value: boolean) => {
+                    const updatedContacts = updateContactsObject(
+                      value,
+                      'isPrimary'
+                    );
+                    setContacts(updatedContacts);
+                    return;
```
bobular, line 746: I just spotted this rogue `return`. Doesn't do any harm of course.

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -651,124 +654,121 @@ function UploadForm({
...
+                  }}
+                  onRemoveContact={(
+                    event: React.MouseEvent<HTMLButtonElement>
+                  ) => {
+                    event.preventDefault();
+                    const updatedContacts = [...contacts];
+                    updatedContacts.splice(index, 1);
+                    setContacts(updatedContacts);
```
bobular, line 754: Since these are repeated a few times, we could consider something a bit more concise, that again makes use of the functional updater.

```suggestion
                    setContacts(prev => prev.filter((_, i) => i !== index));
```

bobular (commented):
Just a few more things I spotted before closing my browser tab!

bobular:
Maybe a `border-radius` on these outlines to distinguish them from the input form boxes (with regular corners)?

image

asizemore:
Screen Shot 2025-05-12 at 5 25 01 PM

I agree the border isn't pretty. I don't want to add a border radius because it'll break with the right-angle-corner convention above on that page. Though then what about those big outlined buttons...

Anyways we could also just do a really light gray or light blue background

Screen Shot 2025-05-12 at 5 28 09 PM

Neither is great, but do you like any better than what we have? I liked your padding suggestion btw and applied it!

asizemore:
Looking good! Ready for you to take a quick peek again, @bobular !

bobular:
> 
> Neither is great, but do you like any better than what we have? I liked your padding suggestion btw and applied it!

I like both of those shaded background versions. You choose!

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
```diff
@@ -683,75 +672,52 @@ function UploadForm({
...
                     }}
                     isPrimary={contact.isPrimary}
                     onAddIsPrimary={(value: boolean) => {
-                      const updatedContacts = updateContactsObject(
-                        value,
-                        'isPrimary'
-                      );
-                      setContacts(updatedContacts);
+                      updateContactsObject(value, 'isPrimary');
+                      // setContacts(updatedContacts);
```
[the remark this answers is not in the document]
asizemore: Oh dear I didn't! I'll go get those out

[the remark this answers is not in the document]
asizemore: All cleaned now :)

bobular (approved):
Exclusive "is primary" buttons work great! Code is tighter - looks great!

Go ahead with a shaded background instead of the outline, if you like.

I'm approving but there are a few UX/accessibility niggles. I'll make a [new ticket](https://github.com/VEuPathDB/web-monorepo/issues/1400) for these.

bobular:
Sorry, I missed one missing `styleOverrides` in `MesaButton/index.tsx` sorry, could you add it?

```
  const finalStyle = useMemo(
    () => merge({}, defaultStyle, themeStyle, styleOverrides),
    [themeStyle]
  );
```

asizemore:
No worries, i added it! Thanks Bob!

PR #1373, opened 2025-04-22
Handle 300 responses from the record service (returned when ID maps to >1 new IDs)

ryanrdoherty:


bobular (approved):
Approving but would like to test/understand the context.

bobular:
Still very approving, but I updated/improved the verbiage. Without it I think some users would think "Why search for something that isn't there?"

image

PR #1375, opened 2025-04-28
implement showExtraMetadata flag

asizemore:
Adds a conifer flag that is used to control showing/hiding the extra metadata fields we have been discussing. 

Should be default to false, and also set to false in all .env files.


aurreco-uga (approved):

asizemore:
The second commit is just prop passing. I realized that the `user-datasets` can't grab the config from `web-common` because it's a dependency! That was causing the github actions failure. 
So, I had to go to a way higher up component to get the config value, then pass it all the way down to the user dataset components.

Tested locally on mbio, clinepi, and genomics.

PR #1380, opened 2025-04-29
Make it easier to read the numbers from the phyletic distribution component

### LINKED ISSUE #1379: Phyletic distribution checkbox/tree can be so wide that the counts are hard to interpret
Here's an example (and my screen isn't crazy-wide)

Image

Two options come to mind
1. dynamic width calculations and styling (depending on how many nodes/rows have been expanded)
2. hover highlighting

The highlighting is very much the simpler option.

bobular:
Fixes #1379 

Check out the video for the component at the top of the orthogroup table.

The highlighting is also there on the "taxon filter" on the main ortho "list of proteins" table.

https://github.com/user-attachments/assets/1e909326-c243-4741-9152-4bb68212c2aa

### DIFF
--- a/packages/sites/ortho-site/webapp/wdkCustomization/js/client/components/phyletic-distribution/PhyleticDistributionCheckbox.scss
+++ b/packages/sites/ortho-site/webapp/wdkCustomization/js/client/components/phyletic-distribution/PhyleticDistributionCheckbox.scss
@@ -17,6 +17,10 @@
     &__species {
       font-weight: normal;
     }
+
+    &:hover {
+      background-color: rgba(0, 0, 0, 0.05); /* subtle grey */
+    }
   }
 
   .wdk-CheckboxTreeItem .wdk-CheckboxTreeNodeWrapper:hover {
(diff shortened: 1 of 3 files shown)


aurreco-uga:
was it considered to make the tree/table narrower?

bobular:
Good idea. How about this?

image

aurreco-uga:
keep the grey highlight too?

bobular:
> keep the grey highlight too?

Yes, I think so.

bobular:
PS. the tree component is only half-width in the ortho group record page section. It is full width in the taxonomy filter popup!

asizemore:
Love it. Big fan of the gray highlight. Works great. Two comments
1. Can we extend the gray a few pixels to the right of the number? For styling purposes so that the number has some breathing room within the highlight
2. For a separate issue, but i think the table should just have a column title that says "Number", instead of having help text elsewhere that explains what the number is

asizemore (approved):
No blockers, just the styling suggestion above.

PR #1385, opened 2025-05-01
Payment form invoice number input

bobular:
Here's what it looks like

image

Notice there are now placeholders and the dollar amount is right-justified.


ryanrdoherty (approved):

bobular:
I've added copy-paste invoice number validation on the front end too - was this a good idea?

image

ryanrdoherty:
I don't have a problem with validation duplication.  The mantra I've always heard is: back-end validation is necessary, front-end, sublime.  Better to catch this right away with a nice message than show the spinner and then possibly not handle as well.

PR #1392, opened 2025-05-07
Fix user dataset subsetting distribution OOM

### LINKED ISSUE #1391: Edge case single row UD with large integer causes OOM
The front end is causing a back end OOM when a variable like this

```
{
	"distributionDefaults": {
		"rangeMin": 101,
		"rangeMax": 101,
		"binWidth": 1
	}
}
```

Gets sent to the EDA back end `distribution` endpoint with the following bin spec:

```
{
	"binSpec": {
		"binWidth": 1,
		"displayRangeMax": 101,
		"displayRangeMin": 0
	},
	"filters": [],
	"valueSpec": "count"
}
```

If the max is a large number, then the back end creates a large number of bins and that can OOM.

Clearly the zero `displayRangeMin` is a problem. We could also guard against this on the back end? (New ticket if so, please!)

A simple test case is a user dataset with the following data file:

```
SampleID	Name	Age	Weight
Bobby001	Bob	101	123
```

bobular:
FIxes #1391

### DIFF
--- a/packages/libs/eda/src/lib/core/utils/default-axis-range.ts
+++ b/packages/libs/eda/src/lib/core/utils/default-axis-range.ts
@@ -163,10 +163,9 @@ function getSafeLowerBound({
   // and the number of bins between 0 and rangeMax would be "too many"
   if (
     rangeMin === rangeMax &&
-    binWidth != null &&
-    rangeMax / binWidth > MAX_EDGE_CASE_BINS
+    rangeMax / (binWidth ?? 1) > MAX_EDGE_CASE_BINS
   ) {
-    return rangeMin - MAX_EDGE_CASE_BINS * binWidth;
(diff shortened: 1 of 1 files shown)


packages/libs/eda/src/lib/core/utils/default-axis-range.ts
```diff
@@ -24,7 +24,7 @@ export function numberDateDefaultAxisRange(
             min:
               logScale &&
               observedMin != null &&
-              (observedMin <= 0 ||
+              ((observedMin as number) <= 0 ||
```
bobular, line 27: This was just a tidy-up. My IDE is on a higher version of TypeScript, and it didn't like it.

aurreco-uga (approved):
sorry i dont have time to test so ill approve and we can test in main.

PR #1397, opened 2025-05-10
Handle bad graphs response

ryanrdoherty:
@bobular I cannot figure this out.  It's an very close import to the one a few lines above it but I break on `WdkService is not defined` in the browser (build seems fine).  Any ideas?


packages/libs/web-common/src/components/DatasetGraph.jsx
```diff
@@ -9,6 +9,7 @@ import { safeHtml } from '@veupathdb/wdk-client/lib/Utils/ComponentUtils';
 import ExternalResource from './ExternalResource';
 import { JbrowseIframe } from './JbrowseIframe';
 import { EdaScatterPlot } from './eda/EdaScatterPlot';
+import WdkService from '@veupathdb/wdk-client/lib/Service/WdkService';
```
bobular, line 12: Hi @ryanrdoherty 
The problem here is mainly that `WdkService` is a type, not an object, but because this is a JSX file we don't get any help from the IDE :-( 
I'll see if there's a way to get hold of the actual wdkService object.

bobular: I don't think there's a way*, partly because it's a class-based component, and partly because the calls to dataPlotter are outside the regular WDK ecosystem, so the WDK hasn't been wired into this component.

(* well, not without extensive refactoring, of course)

bobular: Yeah I think once it's a functional component `useWdkService` should be usable? Though probably `useWdkEffect` is the more appropriate hook.

The appropriate context is "provided" here:
https://github.com/VEuPathDB/web-monorepo/blob/526752faa0268931f6c3d821b6c9617bf49643d6/packages/libs/wdk-client/src/Core/Root.tsx#L163

packages/libs/web-common/src/components/DatasetGraph.jsx
```diff
@@ -110,7 +111,18 @@ export default class DatasetGraph extends React.PureComponent {
     this.setState({ loading: true });
     this.request = httpGet(baseUrl + '&declareParts=1');
     this.request.promise().then((graphs) => {
-      this.setState({ graphs, visibleGraphs: [0], loading: false });
+      let visibleGraphs = [0];
+      if (typeof graphs === 'string') {
+        // indicates an error from the server; log
+        let error = 'Error from dataPlotter: ' + graphs;
+        // FIXME: Don't understand why this does not exist
+        //WdkService.submitError(error);
```
bobular, line 119:
```suggestion
        // FIXME: ideally we would do something like `wdkService.submitError(error)` perhaps via `useWdkEffect` if we make this a functional component
```

bobular (approved):
Approve the improvement! Don't forget to comment the unused `import` too.

PR #1401, opened 2025-05-14
AI expression - buttons for Was this helpful? 

bobular:
Targets `ai-expression-cost-limit` at the moment!

The :+1: :-1: buttons are regular links to a Qualtrics survey with a couple of fields auto-filled from URL parameter values (hopefully!), but we can add client-side JavaScript too, if needed.

Wide browser:
image

Narrow browser:
image

### DIFF
--- a/packages/libs/coreui/src/components/buttons/SwissArmyButton/index.tsx
+++ b/packages/libs/coreui/src/components/buttons/SwissArmyButton/index.tsx
@@ -39,11 +39,25 @@ export default function SwissArmyButton({
     [buttonState, disabled]
   );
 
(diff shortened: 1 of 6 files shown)


aurreco-uga:
the thumbs might require some background or border? in google search:
Screenshot 2025-05-14 at 8 16 50 AM

bobular:
> the thumbs might require some background or border?

I've made them even bigger (per Omar's suggestion) and I think the mouse-over shading makes it clear they are buttons? Also, outlines make keyboard accessibility tricky (because the browser uses outlines to show keyboard focus).

PR #1409, opened 2025-05-23
Fix aspect ratio of new ortho help image and add flex-wrap layout

bobular:
Image was a little squashed in one direction. I made it square and slightly bigger.

I also changed the layout to side-by-side unless narrow browser makes it wrap below.

image


ryanrdoherty (approved):
Thanks, Bob.  I didn't notice we were hard-coding the aspect ratio and size.  Yes, please go ahead an put it out on QA.  I believe Rich wants it all the way out to prod so if you are happy with it, feel free to push there too.

bobular:
> Thanks, Bob. I didn't notice we were hard-coding the aspect ratio and size. Yes, please go ahead an put it out on QA. I believe Rich wants it all the way out to prod so if you are happy with it, feel free to push there too.

Yeah I think image sizing is still in the 1990s but I could be wrong.

PR #1419, opened 2025-06-12
don't try and fix user input on pubmedids

Foxcapades:


bobular:
Thanks for fielding this Ellie.

The handling of DOIs (below) looks like a sensible middle ground for cleaning whitespace and extra commas delimiting nothing of import

https://github.com/VEuPathDB/web-monorepo/blob/a857ab4207c7b8782244650573077ba22626c91f/packages/sites/genomics-site/webapp/wdkCustomization/js/client/controllers/UserCommentFormController.tsx#L614-L618

PR #1420, opened 2025-06-13
Support VDI API Changes

Foxcapades:
## Changes

* Updates io-ts definitions with move from `projectIds` field to `installTargets`
* Update REST API wrapper to use updated URL path
* Change misc references to `projectIds` throughout hooks and tsx files to `installTargets`
* Add `.nvmrc` file for NVM users to ensure the correct node version is used when working on web-monorepo
* Correct adoc syntax and reformat readme for readability.


bobular (approved):
This looks good. I haven't tested though (not sure against what service?)

I'm glad the old endpoint `/vdi/vdi-datasets` is being replaced with `/vdi/datasets` - all is right with the world now ;-)

PR #1428, opened 2025-06-30
Wgnca notebook improvements 1381

### LINKED ISSUE #1381: WGCNA notebook improvements for prime time
UPDATE 07/09:

- [x] Fix the prevalence input so that the user doesn't have to interact with it to have the configuration considered complete.
- [x] Have all sections be open by default.
- [x] Pin Data 1 based on the site.
- [x] Add check for ensuring app is valid for the project
- [x] Make clear that 0.75 correlation threshold is ANDed with module membership
- [x] (3) "get my list of genes" or "run the search for genes" or other improved verbiage
- [x] Ensure visualization requests data when the compute finishes

_The following was completed in previous PRs. Leaving here for nostalgia's sake_ 

Outside of it just working, a few things we need to do:

- [x] Connect the module dropdown with the visualization in a way that makes sense. Users should be able to see both at once on the screen. Replace `getNodeMenuActions` in `correlation.tsx`. We don't need it anymore.
- [x] Remove/update tooltip dropdown from nodes.
- [x] Fix sizing of network
- [x] Remove the Study Explorer box
- [x] Remove Unnamed Analysis
- [x] Remove extra help
- [x] Add the 1,2,3 numbers to match eda
- [ ] Connect "Get Answer"
- [x] Add titles for bipartite network

In order to replace the whole area with the notebook, we have to replace a "questionForm" with our own component. Something like
```
  {
    type: 'questionForm',
    test: ({ question }) => question?.queryName === 'GenesByWGCNAModule',
    component: EdaNotebookQuestionForm,
  }
```

The component then can do anything. Check out `GenesByEdaSubset` for an example.

asizemore:
Resolves #1381 

This PR adds functionality for the wgcna notebook. Specifically the goals are to
1. Add wdk parameter cell that renders inputs based on a specified list of parameters
2. Update the appropriate states/redux magic so that when a user hits "Get Answer", their param inputs are sent to the wdk.
3. Add interaction between network and param cell.

Some notes:
- At this time 06/30, parts 1 and 2 are functional. There's an issue with the organism, but I'm hoping it'll become more clear as I continue to poke around. I think organism is a hidden parameter for this search, so I'm having trouble finding it and finding where it's supposed to go.
- This PR does NOT address how the analysis object gets stored or how any new parameters are read in. I don't think we have the new eda notebook parameters in the model yet.

Testing:
- qa plasmo
- navigate to `a/app/search/transcript/GenesByRNASeqEvidence#GenesByRNASeqpfal3D7_Lee_Gambian_ebi_rnaSeq_RSRCWGCNAModules`

### DIFF
--- a/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/EdaNotebookQuestionForm.tsx
+++ b/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/EdaNotebookQuestionForm.tsx
@@ -1,0 +1,18 @@
+import DefaultQuestionForm, {
+  Props,
+} from '@veupathdb/wdk-client/lib/Views/Question/DefaultQuestionForm';
+import React from 'react';
+import { EdaNotebookParameter } from './EdaNotebookParameter';
+import { ParameterGroup } from '@veupathdb/wdk-client/lib/Utils/WdkModel';
+
+export const EdaNotebookQuestionForm = (props: Props) => {
+  console.log('EdaNotebookQuestionForm props:', props);
+
+  // question params are in props.state.question.paramNames
+
+  const renderParamGroup = (group: ParameterGroup, formProps: Props) => {
+    return <EdaNotebookParameter value={'test'} />;
+  };
+
+  return <DefaultQuestionForm {...props} renderParamGroup={renderParamGroup} />;
+};
--- a/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/EdaNotebookQuestionForm.tsx
+++ b/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/EdaNotebookQuestionForm.tsx
@@ -6,13 +6,22 @@ import { EdaNotebookParameter } from './EdaNotebookParameter';
 import { ParameterGroup } from '@veupathdb/wdk-client/lib/Utils/WdkModel';
 
 export const EdaNotebookQuestionForm = (props: Props) => {
-  console.log('EdaNotebookQuestionForm props:', props);
-
-  // question params are in props.state.question.paramNames
+  // value will be the serialized analysis object
 
   const renderParamGroup = (group: ParameterGroup, formProps: Props) => {
-    return <EdaNotebookParameter value={'test'} />;
+    return (
+      <EdaNotebookParameter
+        value={'test'}
+        parameters={props.state.question.parameters}
+      />
+    );
   };
 
-  return <DefaultQuestionForm {...props} renderParamGroup={renderParamGroup} />;
+  return (
+    <DefaultQuestionForm
+      {...props}
+      renderParamGroup={renderParamGroup}
+      resetFormConfig={{ offered: false }}
+    />
+  );
 };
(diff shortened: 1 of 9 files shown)


asizemore:
Some things for later:
- We should add max/min values to the wgcna_correlation_cutoff parameter so we can read those in and use them for the `NumberInput`.
- all around the `updateParamValue` I'm just not sure if this is the right way to abstract out some of the functions. 
- I also got a little lost trying to manage the `getVizPluginOptions` function. Not sure if there's a simpler way to do things

packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/EdaNotebookQuestionForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+        parameters={props.state.question.parameters}
+        wdkUpdateParamValue={wdkUpdateParamValue}
+      />
+    );
+  };
+
+  return (
+    <DefaultQuestionForm
+      {...props}
+      renderParamGroup={renderParamGroup}
```
bobular, line 49: This relies on there only being **one** non-hidden paramGroup in the question. If there were two, we'd get two similar-looking EDA notebooks.
Here's the relevant code from `DefaultQuestionForm` L207
```
            {question.groups
              .filter((group) => group.displayType !== 'hidden')
              .map((group) => renderParamGroup(group, props))}
```

If this works in practice for now, no biggie, but it's something to consider in the future.

packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/EdaNotebookQuestionForm.tsx
```diff
@@ -0,0 +1,53 @@
...
+        parameter,
+        paramValues,
+        paramValue: newParamValue,
+      });
+    },
+    [props, searchName]
+  );
+
+  // An override that renders the notebook instead of any default parameter or parameter group ui.
+  const renderParamGroup = (group: ParameterGroup, formProps: Props) => {
```
bobular, line 36: My IDE is warning about unused args `group` and `formProps`.  If I prefix with `_` the warning goes away. This also seems to compile fine:

```

File: packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/EdaNotebookQuestionForm.tsx [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+      paramValues: ParameterValues
+    ) => {
+      props.eventHandlers.updateParamValue({
+        searchName,
+        parameter,
+        paramValues,
+        paramValue: newParamValue,
+      });
+    },
+    [props, searchName]
```
bobular, line 32: first dependency could probably be `props.eventHandlers.updateParamValue`

bobular (commented):
Just leaving some comments before I get into the meat of it. github is behaving a bit strangely and I'm worried I might have lost previous comments (I've done about 4 but it's only saying 1 pending comment...). Want to start with a new review to be sure.

Edit to add: ah, looks like the other comments were submitted but I don't remember submitting a review...

PR #1435, opened 2025-07-07
Mbio diff abund remove values from picker that have been filtered out

### LINKED ISSUE #1417: Variable values removed because of a different variable filter still showing in app dropdowns
https://github.com/user-attachments/assets/879ecec5-a45c-4743-a1a6-64185a3d931a

Seen on qa and prod.

What is the correct fix? The user didn't subset on this second variable, so they may not be aware that those options are effectively gone. My favorite idea right now is to show the options but have them greyed out, so that they're not selectable.

asizemore:
Resolves #1417 

This PR adds functionality that asks the backend for the filtered distribution for the comparator variable. The result gives us a way to disable variables values in the drop downs that have been totally filtered out of the subset. 

Note this fix is only for categorical variables. Distributions for continuous variables already took into account the current filters.

Testing
- tested with local eda with qa mbio backend

### DIFF
--- a/packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
+++ b/packages/libs/eda/src/lib/core/components/computations/plugins/differentialabundance.tsx
@@ -1,6 +1,8 @@
 import {
   ContinuousVariableDataShape,
+  DistributionResponse,
   LabeledRange,
+  NumberVariable,
   usePromise,
   useStudyMetadata,
 } from '../../..';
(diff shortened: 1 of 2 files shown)


asizemore:
@bobular see if the error still exists. I couldn't reproduce your error but i did get a different one that complained about `filteredComparatorVariableDistribution.value.histogram` not having the `some` method. I'd guess they're related.
Let me know how this fix does/doesn't work for you!

PR #1438, opened 2025-07-09
Add margins to fix clipped red text bug

### LINKED ISSUE #1430: Floating window/popup for column customisation of mbio curated studies table formatting issue
The red text is truncated, and shouldn't be.

Actually, given that this one is a modal rather than a floater, it doesn't need to be movable at all.

Image

bobular:
Fixes #1430 

Tested in gene page AI expression tool too. Extra margins do not cause window width to change :-)


asizemore (approved):
Looks nice! Curious - was 42 found by playing around with it and checking visually what looked good or was it calculated somehow?

PR #1447, opened 2025-07-24
Add opt-in keyboard movement for Dialog components

### LINKED ISSUE #1446: `Dialog` component should only be keyboard-movable by choice, not by default.
I see too many bugs introduced when I added keyboard accessiblity for moving draggable Dialog boxes. e.g. here is one already squashed: https://github.com/VEuPathDB/web-monorepo/issues/1430

But I've seen another - the login Dialog (`packages/libs/wdk-client/src/Views/User/LoginForm/LoginForm.tsx`) goes into keyboard mode when you type an 'm' in your email address (the 'm' isn't swallowed up, at least). (Note: this login form may only be visible on `@veupathdb/genomics-site` developer instances, as logins are handled by the portal for regular users.)

But there will be more.

The solution is to add a `allowKeyboardMoving: boolean` prop to the component and only switch this on where it is needed. Probably only in the AI expression summary UI for now.

Ideally do this in one commit so it's easier to cherry-pick.

bobular:
Fixes #1446 

  Changes Made:
  - Dialog Component: Added allowKeyboardMoving prop (defaults to false) to make keyboard movement opt-in
  - AiExpressionSummary: Opted into keyboard movement with allowKeyboardMoving={true}
  - WordCloudModal: Opted into keyboard movement with allowKeyboardMoving={true}

  Problem Solved:
  - Login forms and other text inputs won't accidentally trigger keyboard movement when users type 'M'
  - Components designed for keyboard movement still work by explicitly opting in
  - Maintains backward compatibility for all existing Dialog usages

### DIFF
--- a/packages/libs/wdk-client/src/Components/Overlays/Dialog.tsx
+++ b/packages/libs/wdk-client/src/Components/Overlays/Dialog.tsx
@@ -32,6 +32,12 @@ type Props = {
 
   draggable?: boolean;
   resizable?: boolean;
+  /**
+   * Enable keyboard movement of the dialog. Defaults to false.
+   * When enabled, pressing 'M' will enter keyboard-positioning mode
+   * where arrow keys can be used to move the dialog.
+   */
+  allowKeyboardMoving?: boolean;
   className?: string;
   onOpen?: () => void;
   onClose?: () => void;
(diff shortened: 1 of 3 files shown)


asizemore (approved):
Looks good! Tested the ai summary modal.

So the overarching idea would be to keep the press-m-for-keyboard-moving only if the dialog does not involve a user typing something in the dialog, correct?

Issue #1445, opened 2025-07-23
Subscription tracking front end work


bobular:
We will start with some technical implementation notes but should add mockups and other specs at the top of this description as they are made. See https://veupathdb.atlassian.net/wiki/spaces/UI/pages/1445789713/Subscription+Management+Design+and+Implementation for the general design situation.

- [x] conditional banner basics (currently using `user.isGuest`)
- [x] time-limited banner dismissal
- [x] basic banner contents
- [x] basic modal form wired to banner
- [x] user icon badges  ✔️ ❌ and add subscribed/not-subscribed to mouse-over and wire in modal form
- [x] refine modal form

PR #1448, opened 2025-07-24
Subscriptions user management

bobular:
Fixes_ #1445

A major change here is that banner dismissal "memory" was held in session storage, which is tab-specific and is cleared out when browser or tab are closed. We've moved it to local storage which persists across tabs and sessions, as it's likely the user will close their browser during the 48 hour period.

Added placeholder text for the banner (the hairline border and grey background are optional - easily removed) and the user profile icon status markers:

# unsubscribed (simulated with logged out/`isGuest` status)
image

# subscribed (simulated with logged-in user)
image

# very narrow screen (unsubscribed)
image

### DIFF
--- a/packages/libs/wdk-client/src/Actions/UserActions.ts
+++ b/packages/libs/wdk-client/src/Actions/UserActions.ts
@@ -625,16 +625,15 @@ export function submitPasswordReset(
   };
 }
 
-async function logOut() {
-  await fetch(`/service/logout`, {
-    credentials: 'include',
-  });
-  window.location.assign('/');
-}
-
-export function deleteAccount(): ActionThunk<void> {
+// FIXME: this method does not properly log out and errors silently
+export function deleteAccount(): ActionThunk<EmptyAction> {
   return function run({ wdkService, transitioner }) {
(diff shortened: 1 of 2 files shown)


bobular:
We'll need a bit more logic not to show some of this stuff for guest users!

packages/libs/web-common/src/App/UserMenu/UserMenu.tsx
```diff
@@ -0,0 +1,148 @@
...
+}
+
+const UserMenu: React.FC<UserMenuProps> = ({ user, actions }) => {
+  if (!user) return null;
+
+  const { isGuest, properties = {} } = user;
+  const iconClass = 'user-circle' + (isGuest ? '-o' : '');
+
+  const renderMenu = (): JSX.Element => {
+    const items: MenuItem[] = !!user?.isGuest
```
asizemore, line 35: Can this just be `= isGuest`?

asizemore:
After a discussion with our friend chatgpt about best practices for save buttons, i can finally sleep tonight.

https://github.com/user-attachments/assets/6e96d2e5-bc69-4058-9372-236f59bf06d5

PR #1451, opened 2025-08-02
"Tabbed" user profile navigation and form behaviour changes

bobular:
Still definitely WIP but the important things are in place

- [ ] navigation style compatible with gene pages etc
- [ ] warnings and resetting when changing from one unsaved tab to another

Styling and verbiage still TBD!


asizemore:
And what do you mean by "form upload status chip"? I'm not sure what that is

bobular:
> And what do you mean by "form upload status chip"? I'm not sure what that is

That's the `*** You have unsaved changes ***` thing in my screenshot.

If we auto-save, then do you think it should also have a "Your changes are being saved" and "Your changes have been saved" status bar?

On the whole I think I'll need a bit more convincing on auto-save!

asizemore:
Screen Shot 2025-08-11 at 3 14 30 PM


current status. 

Still need to do small fix to the save/reset buttons.

asizemore:
A few thoughts on the save button before i forget. 
For some reason the disabled Save didn't look right when i played around with it. Not sure why because it makes sense not to save if we have nothing to save. But also it causes a funky transition when the button changes color from active to disabled after saving is complete.

I removed the success green because I think the button should stay the same primary color, instead of changeing from primary to green. Maybe if we nicely faded the green to the primary color it'd look better? Or we could always have the save button be green and never the primary color? It is a special action and often times the save button is green... but idk...

I used the primer linked above to help. It suggested we never need to disable the save button, which helped alleviate some of my confusion. 

The other bit i struggled to get super smooth was the unsaved changes. I removed "unsaved changes" from the nav because we can't actually leave the tab with unsaved changes thanks to the nice modal popup. The "Discard Changes" button now pops up when we have unsaved changes, which i like a little better than always being there (that could be just me though!) because when the Save button changes width, it's not pushing around another button so much. 

Which brings me to the Save button changing width with the presence/absence of an icon. I love the spinny loady thing and i love the check mark and i absolutely want to see them. Maybe we should make the save button a fixed size so that it doesn't change so much? Would that make the change from Save -> Saved too subtle? Also not sure.


@bobular so sorry for all the monstrous edits to your PR. Let's chat when you return!

asizemore:
Last thing to do is to update styling for the pop up that we get when we have unsaved changes. But i'll leave that for tomorrow!

asizemore:
https://github.com/user-attachments/assets/f51e3da9-9f71-49df-b65e-17f233bcab6c

PR #1454, opened 2025-08-18
Add custom user icons

asizemore:
Adds special icons `UserWarn` and `UserCheck` so they can be used for the subscription badge project.

The actual icons themselves took very little time to create. Figuring out the css of the icon is now giving me a small headache :) 

Screen Shot 2025-08-18 at 2 27 45 PM


asizemore:
@bobular no worries! Found and fixed. There was another icon with an addition that was getting clobbered as well, so it was an easy fix to upgrade `svgr/cli` in order to use their new `ignore--existing` flag. No more unnecessary overwriting!

bobular:
> @bobular no worries! Found and fixed. There was another icon with an addition that was getting clobbered as well, so it was an easy fix to upgrade `svgr/cli` in order to use their new `ignore--existing` flag. No more unnecessary overwriting!

Nice one!

PR #1456, opened 2025-08-20
Groups backend integration

ryanrdoherty:
Add calls to the new subscription groups endpoint, then use the data to:
- determine person icon in corner
- determine whether subscription banner should be shown
- render subscription section of profile page and let user select


packages/libs/web-common/src/App/UserMenu/UserMenu.tsx
```diff
@@ -33,6 +35,23 @@ const UserMenu: React.FC<UserMenuProps> = ({ user, actions }) => {
   const { isGuest, properties = {} } = user;
   const iconClass = 'user-circle' + (isGuest ? '-o' : '');
 
+  const subscriptionGroups =
```
bobular, line 38: The early return is a convenient way to avoid excessive checks for defined-ness but unfortunately does break the [rules of hooks](https://react.dev/reference/rules/rules-of-hooks), which protect us from weird edge case bugs in the future.

One option to solve this is to move the "early return" to a conditional render in `packages/libs/web-common/src/components/homepage/Header.tsx` where we do something like

```
  {user ? (
    <UserMenu webAppUrl={webAppUrl} user={user} actions={actions} />
  ) : (
    <GuestMenu webAppUrl={webAppUrl} actions={actions} />
  )}
```

A new `GuestMenu` component would handle log in/register only and would not need the <hr> and subscribed/unsubscribed (should probably be "Not subscribed"!) addition at the bottom.

This would be a clean separation of concerns (channeling Dave!).

bobular (commented):
Just a quick comment :point_down:

bobular:
I think I've made the final UX-related changes:

* merged in the modal-properly-on-top sub-branch
* ensured that all user and subscription state ultimately comes from the database. Now updates on-screen all happen at the same time
* there was a bug where if you 1. joined a group. 2. left the group. 3. started to join another group but pressed 'discard changes'; the UI would show you as belonging to a group. The previous item fixed this.
* more tweaks probably


TO DOs

- [ ] Maybe implement the full `subscriptionGroups` check for the "Subscribed / not subscribed" status subheading below "Account: Joe User" heading of /user/profile. When subscriptions have expired, we will show conflicting information.
- [ ] Subscribed/Not subscribed verbiage where needed
- [ ] Merge in Ann's `UserMenu` work when ready
- [ ] Test registration and ProfileModal again

packages/libs/wdk-client/src/Views/User/UserFormContainer.tsx
```diff
@@ -136,6 +141,11 @@ function UserFormContainer(props: UserFormContainerProps) {
     }
   }
 
+  // for this purpose, easier to not confirm against groups; any value will do
+  // (though technically it could clash with the subscriptionGroups-based checks elsewhere)
+  const subscribed =
+    props.globalData?.user?.properties?.subscriptionToken !== '';
```
bobular, line 147: Maybe do more here?

packages/libs/wdk-client/src/Views/User/UserAccountForm.tsx
```diff
@@ -145,25 +147,46 @@ function UserAccountForm(props: UserAccountFormProps) {
...
+                <legend>Delete My Account</legend>
+                <p>
+                  All your personal information will be removed from our systems
+                  and any contributions you have made will be anonymized.
+                </p>
+                <p>This action cannot be undone. Please be sure.</p>
+                <OutlinedButton
+                  text="Delete My Account"
+                  onPress={(e) => onDeleteAccount()}
+                  themeRole="primary"
```
bobular, line 185: Could we make it red with `themeRole="error"`?

PR #1457, opened 2025-08-20
Banner styling updates

asizemore:
So far i've just attacked the banner.

To make a long story short, we were halfway to creating the new banner (shown in figma mockups) when the first round of cuts happened. So we have these gorgeous mockups of banners, but hand't implemented them. This PR doesn't do all of the implementing (that should be branched from main), but adds enough for us to get rolling with the subscriptions stuff.

So far
- theme palettes should generally have warning, error, etc. colors defined, so i've added ours. This should have no effect anywhere other than the banner
- banner gets two options for call to action buttons. Primary (filled) and secondary (outlined). Both look nice with the close button :) 

One open question - currently, the banner uses specific levels (eg. 600) to set the border color and so on. This color should match the button colors. Right now it does because our theme colors are set to level 600. But in the future they may not be. So we could 1. have the banner component define it's CTA button colors using level 600 regardless (pretty nice but verbose), 2. the Button components could take a `themeLevelOffset` prop that shifts their button color 100 up or down, or 3. have the banner set its borders and such based on the theme color level instead of level 600. I like the third one best right now...
Update: Done. Now, for example, if we set the error theme to red: 400, the buttons (which listen to the 400 part), will match the banner (who now also listens to the 400 part instead of asserting it must be 600).

Screen Shot 2025-08-20 at 3 03 45 PM

Screen Shot 2025-08-20 at 3 11 10 PM

For adding the newly styled banners to the announcements, I aimed to leave alone as much as the current strategy as possible and just hijacked the rendering of the component. The idea is that now if the announcement has `renderAsIs=true`, the render method will be called outside of all the wdk announcement containers. In other words, we'll just render it as-is.

### DIFF
--- a/packages/libs/coreui/src/components/banners/Banner.tsx
+++ b/packages/libs/coreui/src/components/banners/Banner.tsx
@@ -285,6 +285,13 @@ export default function Banner(props: BannerComponentProps) {
               {secondaryActionButtonProps != null && (
                 <OutlinedButton
                   text={secondaryActionButtonProps.text}
+                  themeRole={
+                    type === 'danger'
+                      ? 'error'
+                      : type === 'normal'
+                      ? 'primary'
+                      : type
+                  }
                   styleOverrides={{
                     container: {
                       marginRight: '0.5em',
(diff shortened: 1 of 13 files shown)


asizemore:
Tested that ortho, mbio, and clinepi also build just fine.

Issue #1460, opened 2025-08-25
Redundant `dataPlotter.pl` back end calls are made and immediately aborted.


bobular:
This is an issue on the production site that is possibly wasting back end resources.

To reproduce, either

1. Go to a gene page, and the Transcript Expression section and minimise ALL the rows of the table. Then open devtools network tab and reload the page. Now open one of the rows of the table. You should see two identical `dataPlotter.pl` calls, the first of which is immediately aborted.
2. Same as above. Open some rows of the Transcript Expression table. Open devtools and reload the page. You should see several aborted `dataPlotter.pl` calls.

Even though the client aborts the requests, it is possible that the Perl CGI processes on the back end are continuing. We need to check this. It's possible also that a SIGPIPE signal from Linux aborts the Perl process, but, again, we should check.

In any case it's not very elegant behaviour, and still wastes resources even if the processes are eventually terminated.

**The cause of the problem** is that the `rowData` object in
`packages/libs/web-common/src/components/DatasetGraph.jsx`
soon to become
`packages/libs/web-common/src/components/DatasetGraph.tsx`
is not stable.

Somewhere in the upstream components (or probably in redux) the object is being recreated unnecessarily.

PR #1461, opened 2025-08-25
DatasetGraph functional component and TypeScript refactor for potential WDK error reporting

bobular:
DataPlotter errors were not being handled cleanly so we tried to rectify this in https://github.com/VEuPathDB/web-monorepo/pull/1397

We thought that also reporting the errors through the WDK would be useful (even though the back end errors are from Perl CGI, not WDK services).  This can still be up for debate.

However, to use WDK error reporting we needed to modernise the `DatasetGraph.jsx` component:
1. convert class-based component to a functional component
2. convert to TypeScript (for our sanity)

I have done this in this PR, and I also added the WDK error reporting (though it is easily removed).

While testing I noticed some issues with redundant dataPlotter.pl XHR calls being made and aborted. This problem also existed in the prior JSX version (on production) - fixing it is beyond the scope of this work - here is the new ticket: https://github.com/VEuPathDB/web-monorepo/issues/1460


bobular:
> Hard to compare the changes from jsx -> tsx. A scan of the tsx seems ok and looks nice though. What are you hoping to get out of the review? Any question points?

Thanks for having a look. I was mostly aiming for validation and general praise :rofl: 

I already tested that gene page Transcript Expression graphs worked. Difficult to test the dataPlotter exceptions! I'll merge this after the dust has settled from more urgent things.

bobular:
Tested again, just to be sure, before merging.

PR #1462, opened 2025-08-25
Suggested subscriptions update-and-submit-now approach

bobular:
The `onPropertyChange` enhancement I was suggesting during our call should work, I think.

I haven't done any testing but this all compiles.

Approximate usage in the `UserSubscriptionManagement` component (to which `formStatus` and `onPropertyChange` are passed as new props) is like this

```typescript
    const handleLeaveGroup = () => {
      // Call onPropertyChange with submitNow=true for atomic update+submit operation
      onPropertyChange('subscriptionToken', true)(''); // Clear the field and submit
      setShowConfirmModal(false);
    };
```

```tsx
                <div style={{ display: 'flex', gap: '1em', marginTop: '1em' }}>
                  <SaveButton
                    formStatus={formStatus}
                    onPress={handleLeaveGroup}
                    customText={{
                      save: 'Yes, Leave',
                      saving: 'Leaving...',
                      saved: 'Left Organization'
                    }}
                    themeRole="primary"
                  />

                  <button 
                    onClick={() => setShowConfirmModal(false)}
                    style={{
                      background: '#6c757d',
                      color: 'white',
                      border: 'none',
                      padding: '0.5em 1em',
                      borderRadius: '4px'
                    }}
                  >
                    Cancel
                  </button>
                </div>
```

### DIFF
--- a/packages/libs/wdk-client/src/Views/User/Profile/UserProfile.tsx
+++ b/packages/libs/wdk-client/src/Views/User/Profile/UserProfile.tsx
@@ -19,11 +19,6 @@ type UserProfileProps = Omit<
   | 'onSubmit'
 > & {
   globalData: { user?: User; config?: ServiceConfig };
-  userEvents: {
-    submitProfileForm: (userData: UserProfileFormData) => void;
-    updateProfileForm: (newState: UserProfileFormData) => void;
-    resetProfileForm?: (formData: UserProfileFormData) => void;
-  };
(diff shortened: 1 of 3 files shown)


bobular:
We will need to add an optional `onComplete` callback prop to `SaveButton` so we can let it do its thing in the modal before it closes.

This will be useful elsewhere I imagine.

ryanrdoherty (approved):

PR #1468, opened 2025-08-28
use datasource record

steve-fischer-200:


aurreco-uga (approved):
difficult to test without data so approving

PR #1476, opened 2025-09-04
Remove state mutations which caused duplicate dataPlotter requests

### LINKED ISSUE #1460: Redundant `dataPlotter.pl` back end calls are made and immediately aborted.

bobular:
Fixes #1460 

I didn't convert the culprit to TSX, sorry, but this should solve the issue (see issue for how to test/reproduce).

This will need extensive testing as it affects ALL RecordTable components. However, it should improve performance across the site as we were re-rendering table child rows unnecessarily due to rowData reference instability.

### DIFF
--- a/packages/libs/wdk-client/src/Views/Records/RecordTable/RecordTable.jsx
+++ b/packages/libs/wdk-client/src/Views/Records/RecordTable/RecordTable.jsx
@@ -38,23 +38,26 @@ const maxColumns = 4;
 const defaultSortId = '@@defaultSortIndex@@';
 const defaultSortColumn = [{ name: defaultSortId, isDisplayable: false }];
 
-const getSortIndex = (rowData) => rowData[defaultSortId];
-
-const addDefaultSortId = (row, index) =>
-  Object.assign({}, row, { [defaultSortId]: index });
-
 const getColumns = (tableField) =>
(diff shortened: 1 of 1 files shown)


bobular:
I've tested again after the new commit. All seems fine.

PR #1486, opened 2025-09-15
Fix UserMenu issues in non-genomics sites

bobular:
We were not conditionally rendering the `UserMenu` and new `UserMenuGuest` components properly because we had missed `HeaderNav.jsx` which seems to be different between genomics and non-genomics.

I noticed that the width-responsiveness of the ClinEpi header is broken (I'd fixed this a few weeks ago for genomics) and so I'll add this to this PR

- [ ] fix `UserMenu` issues
- [ ] fix width responsive issues of ClinEpi header (`UserMenu` disappearing at some widths)

### DIFF
--- a/packages/libs/web-common/src/App/Header/HeaderNav.jsx
+++ b/packages/libs/web-common/src/App/Header/HeaderNav.jsx
@@ -228,7 +228,11 @@ class HeaderNav extends React.Component {
           )}
           <div style={{ display: 'flex' }}>
             <IconMenu items={iconMenu} />
-            <UserMenu webAppUrl={webAppUrl} actions={actions} user={user} />
+            {user && !user.isGuest ? (
+              <UserMenu webAppUrl={webAppUrl} user={user} actions={actions} />
+            ) : (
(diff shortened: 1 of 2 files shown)


asizemore:
I was getting an error about can't read property `subscriptionToken` on undefined when i'd log out. The following fix worked for me in `Subscriptions.ts`:
```
...
return (
    !user.isGuest &&
    user.properties &&
    user.properties['subscriptionToken'] != null &&
    subscriptionGroups?.find(
      (g: SubscriptionGroup) => 
        g.subscriptionToken === user.properties['subscriptionToken'] 
    ) != null
  );
```

asizemore (approved):
See above comment about an error, but otherwise nothing blocking. The updates look really nice!!

bobular:
I think we have to raise the `can't read property subscriptionToken on undefined` as a new ticket. Unless it's something this PR has caused?

bobular:
Also cherry-picked to `v1.3.6-patch` @aurreco-uga

PR #1487, opened 2025-09-15
Improve Account page styling

### LINKED ISSUE #1474: Subscription tracking followup issues
Now that it all works, we need to:

- [x] Finish getting the account page header section to match mockups. Just some spacing and font weight left to tweak
- [ ] Remove all of the fieldset thingys. Basically forces our hand to update the registration page. Looks like we can do that with minor inconveniences though.
- [ ] Redo the registration page so that it matches the subscription page
- [ ] Update Preferences and Security to match mockups. Update to give everything nice classes. Current styles are a mess!
- [x] Typahead cursor
- [x] More space for the leave group confirmation dialog
- [ ] Sub-heading below "Account: Joe User" that says "Subscribed/Not subscribed" should not show while `subscriptionGroups` is loading. But the vertical space should be reserved.

asizemore:
Resolves #1474 

So far, updated Account page so that they match the mockups.
Note that the email preferences is a wrapped component. It's not used anywhere else, though, so I propose we just unwrap it and use as-is.

To Do (still)

- [ ]  Still need to check registration page
- [ ] Subheading below Account should not show while `subscriptionGroups` is loading but we should reserve the white space.

Testing
I attempted to use a fancy responsive grid. Not sure it's super great so please test and adjust sizing as needed.

### DIFF
--- a/packages/libs/wdk-client/src/Views/User/ProfileNavigationSection.tsx
+++ b/packages/libs/wdk-client/src/Views/User/ProfileNavigationSection.tsx
@@ -90,12 +90,6 @@ const ProfileNavigationSection: React.FC<ProfileNavigationSectionProps> = ({
 
   const theme = useUITheme();
 
-  const activeSectionStyle = {
-    // Use a light version of the primary color if defined, otherwise use this nice
-    // light blue.
-    backgroundColor: theme ? theme?.palette.primary.hue[200] : '#d5eaf5',
-  };
-
   return (
     <>
       <div className="wdk-ProfileNavigationSection">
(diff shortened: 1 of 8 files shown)


asizemore:
Registration page

Uploading Screen Recording 2025-09-17 at 4.58.05 PM.mov…


Account page

https://github.com/user-attachments/assets/3dee12bc-d581-4eb1-bab3-18d01adf2c84

asizemore:
@bobular did the subheading loading issue get solved elsewhere? I'm not seeing it anymore

packages/libs/wdk-client/src/Views/User/ProfileNavigationSection.tsx
```diff
@@ -90,12 +90,6 @@ const ProfileNavigationSection: React.FC<ProfileNavigationSectionProps> = ({
 
   const theme = useUITheme();
 
-  const activeSectionStyle = {
```
asizemore, line 93: just removing unused code here

bobular:
I guess there's some potential for the Security tab:

* logout button
* clear cookies and logout?
* (above two options are for the current site only)
* "Where am I logged in?" and "Log out all" (our other sites; I don't think the back end gives us this right now)

In a new ticket with UX and prioritisation input from the team, I guess.

Should the forthcoming "delete my account" button be here or under the Profile tab or both?

PR #1488, opened 2025-09-16
User subscriptions: Add more searchable info to auto-complete

bobular:
Now searches on PI name and institution.

Diacritic-squashing behaviour of original is preserved.

Some other clean-up while I was there.


ryanrdoherty (approved):
Looks great!

Issue #1439, opened 2025-07-09
WGCNA EDA-in-WDK further issues


bobular:
The following are not urgent. These updates/fixes should be done first: https://github.com/VEuPathDB/web-monorepo/issues/1381

To Dos:

1. Does mouse wheel scroll accidentally open/close expandable panels?
2. KC: add filter for pos/neg correlations
3. Mock up putting the "get answer" inside the box
4. Encourage GO enrichment by adding a suggestion near Get Answer

PR #1492, opened 2025-10-01
WGCNA upgrades

asizemore:
Partially addresses #1439 

Adds correlation direction plot option. Also moves some definitions around for improved organization. The regular network and bipartite network are currently the same in terms of their visualization. The plot is different, but the inputs are the same. I'm starting to massage the data structure to be more compatible as well, so that if eventually we can get the correlation computes merged, we can merge the vizs as well (thoughts for later i suppose).

The backend is on qa, so just need mbio backend to test both networks, or plasmo backend to test the bipartite network within the wgcna notebook.

Confirmed with Dan that mbio does want to see this correlation direction option ✔️

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
@@ -50,6 +50,7 @@ import {
 } from '@veupathdb/components/lib/types/plots/network';
 import RadioButtonGroup from '@veupathdb/components/lib/components/widgets/RadioButtonGroup';
 import { NetworkCorrelationDirection } from './NetworkVisualization';
+import { isRight } from 'fp-ts/lib/Either';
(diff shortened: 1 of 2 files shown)


packages/libs/eda/src/lib/core/components/visualizations/implementations/BipartiteNetworkVisualization.tsx
```diff
@@ -472,37 +480,67 @@ function BipartiteNetworkViz(
...
+            <RadioButtonGroup
+              options={['Positive', 'Negative', 'Both']}
+              selectedOption={
+                capitalize(vizConfig.correlationDirection) ?? DEFAULT_LINK_TYPE
+              }
+              onOptionSelected={(value) => {
+                const validatedValue = NetworkCorrelationDirection.decode(
+                  value.toLowerCase()
+                );
+                if (validatedValue._tag === 'Right') {
```
bobular: See below for a similar comment.

PR #1493, opened 2025-10-02
Delete user client work

bobular:


bobular:
cherry-picked into `v1.3.6-patch`

PR #1498, opened 2025-10-14
Implement reload button for backend-failed results tables

bobular:
To reproduce:
- Any gene search
- Add a transcriptomics column `sense - UM_572 - unique only` - THE SECOND ONE near the bottom (if you search for `UM_5`)
- The table will attempt to reload but cause a 500
- Without this fix, the user is up a very brown river without an oar.

The Problem:
- Backend errors from problematic columns would show an error message
- The table and "Add Columns" button disappeared entirely
- User preferences persisted server-side, causing the error on every page load
- Users were stuck with no way to recover

The Solution:

1. New Action (resetColumnPreferencesToDefault) - Triggers the reset flow
2. Clear Preferences (clearResultTablePreferences) - Deletes both column and sorting preferences server-side
3. Epic Handlers - 6 handlers that respond to the reset action by:
  - Clearing preferences on the server
  - Fetching default values (columns, sorting, page size, filters)
  - Directly dispatching fulfilled actions (not requests)
  - This was key: bypassing the normal request→fulfill chain to work within the existing epic window
4. Reducer Update - Clears error state but preserves critical state like resultType, resultTypeDetails, searchName, and globalViewFilters
5. New Error Component (ResultTableError) - Shows error message + "Reset to Default Columns and Retry" button
6. Controller Wiring - Connects everything, falling back to regular ContentError when searchName isn't available

The critical insight was that the getRequestAnswer epic requires 8 fulfilled actions in sequence. By directly dispatching fulfilled actions from our reset handlers, we complete that chain and trigger the
table reload!

### DIFF
--- a/packages/libs/wdk-client/src/StoreModules/ResultTableSummaryViewStoreModule.md
+++ b/packages/libs/wdk-client/src/StoreModules/ResultTableSummaryViewStoreModule.md
@@ -102,8 +102,8 @@ getFulfillPageSize                                                              
                     │               │              │               │
                     ▼               ▼              ▼               ▼
             fulfillSorting  fulfillGlobalViewFilters  updateInBasketFilter
-                    │               │              │
-                    └───────────────┼──────────────┘
+                    │               │              │               │
+                    └───────────────┼──────────────┴───────────────┘
                                     │
                                     ▼
                             getRequestAnswer


bobular:
> Wow, I do not know the client code very well. Did you really grok all the action passing, smrate gobbledygook, etc. for this task or had you been initiated already when Dave was still here? All looks as fine as I can assess without immersing myself in the whole mess myself.

Just for the record, I didn't get initiated in the details of this when Dave was here. I had to learn about the window thing and how the epics work from following through this example. Claude helped with a lot of the boilerplate but it couldn't get to a working solution on its own. I had to understand the action/epic flow fully first. It could benefit from some visual documentation. I might have a stab at this now, with a little help...

PR #1499, opened 2025-10-16
add user class icon

asizemore:
Adds a little grad cap badge to our user account icon.

Screenshot 2025-10-16 at 4 06 27 PM

Tested with local plasmo site.


bobular (approved):
Awesome!

PR #1501, opened 2025-10-17
NumberInput and DateInput type cleanup

bobular:
Improved type safety by making generic type parameters consistent throughout NumberInput/DateInput components, eliminating type assertions at call sites. Fixed React hooks violations including conditional hook execution, and missing dependency array entries.

Did the same for NumberRangeInput and DateRangeInput. Cleaner internals. Same API, although some callbacks should need less broad types now.

Testing:
- [x] clinepi/EDA
- [x] search for other consumers of these widgets


bobular:
It's only used in `packages/libs/eda` and the regular EDA usage (subsetting, visuializations) has been tested. EDA-in-WDK usage will be tested while we develop it.

PR #1504, opened 2025-10-28
Add PCA (dimensionality reduction) compute and plot

asizemore:
For the differential expression notebook.

This PR works with [PR 75](https://github.com/VEuPathDB/service-eda/pull/75) in `service-eda`. Can use qa mbio to test and see example requests in that PR.

Tasks

- [x] Create computation
- [x] Create visualization

Notes:
Can't use the scatterplot straight up because the dim red plot needs to set its own x/y axes with the computed variable (in beta div, the client doesn't set x/y. it just gets only that stuff from the backend). So I used the `Options` feature to have the new compute `dimensionalityReduction` tell `ScatterplotVisualization` that it wants to use these computed variables _and_ that these computed vars need to be used in the request to the backend. We could have done it more simply with some hard coding but this will make it easier for us to add the ability to choose which principle coordinate the user wants for each axis. The backend is already prepped for this new feature, so the frontend might as well be, too. 

Testing (updated 11/04/25)
- Backend merged on 11/03 so please use with ablevins mbio dev site backend with the local eda workspace frontend (`yarn ... /eda`).

IMPORTANT - do not merge with index.ts still including this compute. The backend has mbio approved for this compute because we need it for data testing. We can merge all this work but we must take it out of index.ts first so that it doesn't accidentally show up on the mbio live site someday on accident 😬


packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -745,8 +748,14 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
         config: {
           outputEntityId: outputEntity.id,
           valueSpec: hideTrendlines ? undefined : valueSpecValue,
-          xAxisVariable: vizConfig.xAxisVariable,
-          yAxisVariable: vizConfig.yAxisVariable,
+          xAxisVariable:
+            sendComputedVariablesInRequest && computedXAxisDescriptor
+              ? computedXAxisDescriptor
+              : vizConfig.xAxisVariable,
```
bobular: Line 382 (`variablesForConstraints`) should probably be calculated consistently too? (including `sendComputedVariablesInRequest` in the conditional?

e.g.
```
  xAxisVariable: sendComputedVariablesInRequest && computedXAxisDescriptor
    ? computedXAxisDescriptor
    : vizConfig.xAxisVariable,
```

bobular (approved):
This looks good. A few minor comments we can discuss if needed. I assume it's all working, but I haven't tested.

Is there another PR where this is integrated into a notebook?

PR #1505, opened 2025-10-30
Add modal gate/explanation wrapper for all Clustal Omega buttons

### LINKED ISSUE #1449: Gene page and OrthoMCL group page - verbiage/gates to protect cgi-bin Clustal Omega from over-use
Until a future time when we build clustal into a service (it is currently cgi), for now there is no way to control how many jobs a user fires off (eg scripting).

But there are a couple of things we can do (in the gene page orthologs table):

* if the user chooses "all" orthologs to align, and if that is greater than, say, 50, show a popup asking them to confirm they want to do this
* when the user clicks the button, always show a popup that warns them that they will get a blank screen when the job is running, and to not retry the job

Bob's preferred modification to the above:

* Don't do anything when user selects all checkboxes - they might not be intending to Clustal them
* When the user clicks the alignment button there is always a confirmation dialogue that
  - warns about >50 sequences (if relevant)
  - explains about the blank page and do not retry

bobular:
Fixes #1449

Adds a modal around three Clustal Omega submit buttons:

1. gene page orthologs+paralogs in PlasmoDB
2. popset results table
3. orthomcl group page

When the number of sequences is >1000 (only really possible from 1.) it acts as a gate and prevents submission of the form (no screenshot for this)

Not sure what to do about the TODO in `GeneRecordClasses.GeneRecordClass.jsx` - do we mention flanking regions at all? Or make a general comment about long and/or many sequences being potentially slow?

Example:
image

### DIFF
--- a/packages/libs/web-common/src/components/records/ClustalAlignmentForm.tsx
+++ b/packages/libs/web-common/src/components/records/ClustalAlignmentForm.tsx
@@ -130,8 +130,8 @@ export default function ClustalAlignmentForm({
               <p style={{ marginTop: '0px' }}>
                 The alignment results will open in a new browser tab.
               </p>
+              <p>The tab may appear empty while the alignment is running.</p>
               <p>
-                The tab may appear empty while the alignment is running.{' '}
                 <strong>
                   Please be patient and avoid resubmitting - multiple requests
                   will not make it faster and can overload our servers.


packages/libs/web-common/src/components/records/ClustalAlignmentForm.tsx
```diff
@@ -0,0 +1,171 @@
...
+            The alignment results will open in a new browser tab.
+          </p>
+          <p>
+            The tab may appear empty while the alignment is running. Please be
+            patient and avoid resubmitting - multiple requests will not make it
+            faster and can overload our servers.
+          </p>
+
+          {showWarning && (
+            <p
```
asizemore: Idk if this is out of scope, but it seems like the warning should come at the top? So the warning comes first in the dialog and then the regular text follows

asizemore (approved):
Looks and works great! I only had stylistic suggestions so approving.

bobular:
I'm surprised @asizemore that you didn't point out my lack of `Banner` use. Anyway, I've used it now and it looks nice.

I separated the new bolded advice into a new paragraph too. It looked a bit odd otherwise.

image

PR #1524, opened 2025-11-13
allow guests access to datasets and org results (links in header menu)

aurreco-uga:


packages/sites/genomics-site/webapp/wdkCustomization/js/client/routes.jsx
```diff
@@ -161,6 +162,26 @@ function DownloadsRouteComponent() {
  * Wrap Ebrc Routes
  */
 export const wrapRoutes = (ebrcRoutes) => [
+  {
+    path: '/search/dataset/AllDatasets/result',
+    requiresLogin: false,
+    component: () => (
+      <AnswerController
+        {...{ question: 'AllDatasets', recordClass: 'dataset' }}
+      />
```
bobular: This would do the same thing:

```suggestion
      <AnswerController
        question="AllDatasets"
        recordClass="dataset"
      />
```

bobular: It would be good to add a comment above these new entries something like this:

```
// Allow guests to access All Datasets for SEO and public visibility
```

packages/libs/wdk-client/src/Core/routes.tsx
```diff
@@ -325,3 +304,40 @@ function parseSearchParamsFromQueryParams(restQueryParams: {
...
+      parseQueryString(props)
+    );
+    return (
+      <AnswerController
+        {...(routeMatches ?? props.match.params)}
+        parameters={parameters}
+        filterTerm={isArray(filterTerm) ? filterTerm[0] : filterTerm}
+        filterAttributes={castArray(filterAttributes)}
+        filterTables={castArray(filterTables)}
+        history={props.history}
```
bobular: For me in my local dev server, it was necessary to pass all these props to `AnswerController` so that table search worked without a null `history` exception.

PR #1537, opened 2025-11-14
Show group members

ryanrdoherty:


packages/libs/wdk-client/src/Hooks/SubscriptionGroups.ts
```diff
@@ -42,3 +45,28 @@ export function useSubscriptionGroups(): SubscriptionGroup[] | undefined {
...
   return data;
 }
+
+export function useSubscriptionGroupsByLead():
+  | SubscriptionGroupWithMembers[]
+  | undefined {
+  const { wdkService } = useWdkDependenciesContext();
+
+  const { data } = useQuery({
+    queryKey: ['subscriptionGroupsWithMembers'],
```
bobular: I'd consider adding a user ID to the `queryKey` array but it is NOT essential.

Theoretically a user could log out and somehow avoid our hard homepage redirect/reload (perhaps in a future change in UX) and then log back in again as a different user and for the next five minutes get the previous user's data.

Alternatively, just leave a comment that this is a potential weak point if our logout flow ever changes.

PR #1548, opened 2025-11-15
Claude/convert web-common to typescript [Claude Code Web]

bobular:
Note that the base is not `main`


bobular:
By closing/deleting this branch we are ignoring a bunch of edits made in `web-common` but first we need to get `coreui` and `wdk-client` sorted. It's probably better to redo the web-common conversion once those are done via https://github.com/VEuPathDB/web-monorepo/pull/1550

PR #1571, opened 2025-12-13
Max recommended organisms gate

bobular:


bobular:
`prettier` seems to have introduced trailing commas in lists. Is this just because this file hasn't been touched in a while?

PR #1573, opened 2025-12-19
Integrate DE notebook

asizemore:
Work in progress.

Currently - replaced fold change search with differential expression as we await the differential expression search to work (awaiting tuning manager).
But we can see how it looks in the search page. Also note i'm using Cristina's dev site with old data so we don't have the right collections yet.

https://github.com/user-attachments/assets/a37a6104-c401-4050-87e1-2494ca504436


bobular:
Closing in favour of #1590

PR #1578, opened 2026-01-07
Gene page open all sections

aurreco-uga:


bobular (approved):
Looks good! I played with @copilot to make a trivial refactor. Note that you can nest `{ ... }` sections with conditional rendering logic inside TSX regions.

PR #1583, opened 2026-01-08
New VDI API compatibility update.

Foxcapades:
## Changes

* Replaces API io-ts type definitions.
* Updates object property references for the new API types.
* Updates API endpoint paths.
* Removes iRODS system compatibility code which was creating significant complexity for no gain.
* Updates patch request to use new structure.
* Remove dead/outdated draft of the 'extra metadata' feature.
* Adds new actions for updating datasets from the list view.
* Comments out quota percent usage column until that calculation is re-added to the VDI dataset listing API.
* Updates gene-list-export form-submission object structure.

### DIFF
--- a/packages/libs/wdk-client/src/Components/Display/CollapsibleSection.tsx
+++ b/packages/libs/wdk-client/src/Components/Display/CollapsibleSection.tsx
@@ -33,7 +33,7 @@ const buttonStyle: React.CSSProperties = {
   padding: 0,
 };
 
-function CollapsibleSection(props: Props) {
+export function CollapsibleSection(props: Props) {
   const {
     className,
     id,
--- a/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/homepage/PageDescription.tsx
+++ b/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/homepage/PageDescription.tsx
@@ -51,17 +51,11 @@ function DefaultComponentSiteDescription({ displayName }: DescriptionProps) {
         .
       </p>
       <p>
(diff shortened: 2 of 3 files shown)


packages/libs/web-common/src/user-dataset-upload-config.tsx
```diff
@@ -380,7 +380,7 @@ function ReferenceGenomeDepdency(props: DependencyProps) {
                 resourceVersion: buildNumber,
               };
         })
-        .filter((dep) => dep != null) as UserDataset['dependencies'];
+        .filter((dep) => dep != null) as DatasetDependency[];
```
bobular:
```suggestion
        .filter((dep) : dep is DatasetDependency => dep != null);                                                                                                                                                    
```

Slightly nicer to replace the type assertion with type predicate in filter. Using `: dep is DatasetDependency` allows TypeScript to properly narrow the type rather than forcing it with `as`.

packages/libs/user-datasets/src/lib/Components/UploadForm.tsx
bobular: We should touch base on where we are with Ann's previous work vs. the latest mockups etc. 

Actually I've done a bit of detective work here.

From the blame of the previous commit, it looks like you removed a bunch of form elements etc from Ann's #1372 (and possibly #1375)

That was presumably to get things working. That UI may not be current either. We can always refer to Ann's PR to avoid reinventing that wheel. But it's probably good to start with a clean slate and a full understanding of what's needed for clinepi/dataExplorer and genomics.

PR #1612, opened 2026-02-24
Suppress false positive webpack warnings during `bundle:dev` build

bobular:


bobular:
Second commit explanation:

**Root cause**: `ify-loader` processes all `node_modules` JS files (to apply browserify transforms), and `ProvidePlugin` injects a `ProvidedDependency` for `process/browser` into any module that references `process`. When both happen to the same module, webpack's filesystem cache tries to serialize the result but `ProvidedDependency` has no registered serializer — hence the warnings. The build works fine; only that specific cache entry is skipped.

**Fix**: A small webpack plugin in `site-webpack-config/index.js` that taps `compiler.hooks.infrastructureLog`. Returning `true` from that `SyncBailHook` tells webpack to suppress the log entry. The filter is specific: only `warn`-level messages from `webpack.cache.PackFileCacheStrategy` that mention `ProvidedDependency` are dropped, so nothing else is affected.

PR #1621, opened 2026-02-27
PCA scatterplot styling improvements

bobular:
Note, it needed a custom `yarn4` execution environment allowing `repo.yarnpkg.com` (+ default allow list) to be able to install the correct version of yarn. So the prompt below was run, then another session from its branch was run to tidy up any compilation errors.

### Claude Code Web prompt

Making backwards-compatible changes only, increase the dot size in the 'de_pca_plot' PCA plot of @packages/libs/eda/src/lib/notebook/notebooks/differentialExpression.tsx and increase the font weight of the mouseover/tooltips on the plot. The overlay variable label is particularly difficult to read with some of the lighter overlay colors.

For the font weight issue, this can be **universal** - no extra prop-passing and override mechanisms necessary. Likely in @packages/libs/components/src/plots/ScatterPlot.tsx

For the point size customization, this should be specific to the PCA scatterplot only, so we will need to engineer the appropriate option via `ScatterplotVisualization.withOptions()` - see @packages/libs/eda/src/lib/core/components/computations/plugins/dimensionalityReduction.tsx 

Ensure compilation with `yarn workspace @veupathdb/eda build-npm-modules` from the root of this repo. And `yarn workspace @veupathdb/components build-npm-modules` for any ScatterPlot.tsx changes.

### DIFF
--- a/packages/libs/components/src/plots/FindNearestDatumXYProvider.tsx
+++ b/packages/libs/components/src/plots/FindNearestDatumXYProvider.tsx
@@ -1,0 +1,130 @@
+/**
+ * Provides XY-based (Euclidean distance) nearest-datum finding for visx XYChart tooltips.
+ *
+ * By default, visx's XYChart finds the nearest datum using only the X dimension,
+ * which is fine for line/bar charts but not for scatter plots like volcano plots
+ * where points are distributed in 2D space. This provider intercepts the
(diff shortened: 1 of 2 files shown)


bobular:
Used outline/shadow and font weight to make the overlay labels more legible.

PR #1669, opened 2026-04-15
Add condition to creating gene PK values; no longer always has project_id (in UniDB)

ryanrdoherty:
Finds gene recordclass and inspects PK to determine if project_id is needed.  This is a little better than "if projectId === 'UniDB'.


bobular (approved):
Nice solution! Nobody likes a hardcoded `== 'UniDB'` or whatever we're calling it today ;-)

PR #1670, opened 2026-04-21
Fix compound record runtime bugs

bobular:
Runtime errors here:
https://beta.plasmodb.org/plasmo.beta/app/record/compound/CHEBI:15608

Claude and I tracked down and resolved two issues:

1. vendored ChemDoodle import had broken at some point recently-ish (Node 24 upgrade?)
2. `module.startsWith()` runtime error is made safe but is really a backend issue - the `MassSpecGraphs` section has a non-informative placeholder-ish entry.

### DIFF
--- a/packages/libs/web-common/src/components/DatasetGraph.tsx
+++ b/packages/libs/web-common/src/components/DatasetGraph.tsx
@@ -35,7 +35,7 @@ interface RecordTableProps {
 interface RowData {
   source_id: string;
   assay_type: string;
-  module: string;
+  module: string | null;
   paralog_number: number;
   graph_ids: string;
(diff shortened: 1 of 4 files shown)


aurreco-uga (approved):
why is this needed? (seems to check if this is a user dataset)

const isUserDataset =

bobular:
> why is this needed? (seems to check if this is a user dataset)
> 
> const isUserDataset =

Later on it's used for formulating the appropriate URLs for links

```
      if (isUserDataset) {
        return '/a/app/workspace/datasets/' + currentRowData.dataset_id;
      }
      return '../dataset/' + currentRowData.dataset_id;
```

PR #1671, opened 2026-04-21
Ortholog table download

ryanrdoherty:
Show organism selector of the user chooses the ortholog table and pass the orgs along to the reporter.

### DIFF
--- a/packages/libs/web-common/src/components/reporters/TranscriptTableReporterForm.jsx
+++ b/packages/libs/web-common/src/components/reporters/TranscriptTableReporterForm.jsx
@@ -5,9 +5,9 @@ import { useWdkService } from '@veupathdb/wdk-client/lib/Hooks/WdkServiceHook';
 
 // Transcript Table Reporter is the same as a regular Table Reporter, but need to
 //   override the recordClass (Transcript) with Gene to get Gene tables for a Transcript result
-let recordClassOverride = {
-  recordClass: { fullName: 'GeneRecordClasses.GeneRecordClass' },
-};
+let recordClassOverride = Object.assign({}, recordClass, {
+  fullName: 'GeneRecordClasses.GeneRecordClass'
+});
 
 const ORTHOLOGS_TABLE_NAME = 'OrthologsLite';
 const ORGANISM_SEARCH_NAME = "GenesByTaxon";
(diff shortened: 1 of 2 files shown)


packages/libs/web-common/src/components/reporters/TranscriptTableReporterForm.jsx
```diff
@@ -1,23 +1,185 @@
...
+
+  const orgParam = useWdkService(
+    async (wdkService) => {
+      const question = await wdkService.getQuestionAndParameters(
+        ORGANISM_SEARCH_NAME
+      );
+      const param = question.parameters.find(
+        (p) => p.name === ORGANISM_PARAM_NAME
+      );
+      if (param == null) return null;
```
bobular, line 34: I moved the `orgParam` modification inside here and added a "performative" dependency array (`[]` would have been fine too) to the `useWdkService` call so that it's only called once.

Before, the `useWdkService` returned a stable `orgParam` despite no dependency array because of WDK caching under the hood. But then we modified it so it wasn't referentially stable on each render. That didn't seem to cause any problems, but it was passed as a prop to `OrganismParam` so it could have caused some inefficiency.

packages/libs/web-common/src/components/reporters/TranscriptTableReporterForm.jsx
```diff
@@ -1,23 +1,185 @@
...
 };
 
+const ORTHOLOGS_TABLE_NAME = 'OrthologsLite';
+const ORGANISM_SEARCH_NAME = 'GenesByTaxon';
+const ORGANISM_PARAM_NAME = 'organism';
+const MAX_ORTHOLOG_SELECTED_ORGANISMS = 5;
+
+const OrganismSelection = ({ props }) => {
+  const { formState, formUiState, updateFormState, updateFormUiState } = props;
+
```
bobular, line 25: I've replaced most of the `let` with `const` - it's our house style, I find it hard to read otherwise, and it helps identify potential unstable "variables".

packages/libs/web-common/src/components/reporters/TranscriptTableReporterForm.jsx
```diff
@@ -1,23 +1,171 @@
...
+        const leafTerms = new Set(
+          getLeaves(orgParam.vocabulary, (node) => node.children).map(
+            (node) => node.data.term
+          )
+        );
+        const leafOrgs = selectedOrgs.filter((org) => leafTerms.has(org));
+
+        if (leafOrgs.length > MAX_ORTHOLOG_SELECTED_ORGANISMS) {
+          alert(
+            `You have selected ${leafOrgs.length} organisms but no more than ${MAX_ORTHOLOG_SELECTED_ORGANISMS} are allowed. Please review your selection.`
```
bobular, line 151: I tweaked this a bit.

packages/libs/web-common/src/components/reporters/TranscriptTableReporterForm.jsx
```diff
@@ -1,23 +1,171 @@
...
+          props.formState.orthologOrganisms == null ||
+          props.formState.orthologOrganisms.length === 0
+        ) {
+          alert(
+            'You must select at least one organism to request an Orthologs table.'
+          );
+          return;
+        }
+
+        const leafTerms = new Set(
```
bobular, line 142: No more TODO and replicated `findLeaf` function. Use `getLeaves` instead.

bobular (approved):
All good. Glad we also figured out the linting issues.

If we ever had any spare time in the future, I'd replace the `alert()`s with a custom modal like the `maxRecommendedGate` but no biggie.

bobular:
Maybe wait for discussion about the "self" species in ui-infra before merging.

PR #1674, opened 2026-04-23
use permissions response to determine if user dataset or curated (to determine wdk record to use)

aurreco-uga:
(this file offers hooks to use by many components; one reason is to know which wdk record to use.
a couple of recent commits undid  wdkrecord hardcoding, in 2 components, instead set to use hooks in this file:
- 40710a7
- 2eb48da
)

claude helped with this file update,  the prompt was to use the permissions response;
updates are:

  1. useWdkStudyRecord (line 46):                                                                                                                                     
                                                                                                                                                                      
  - Now calls usePermissions() to get permissions data                                                                                                                
  - Dynamically sets STUDY_RECORD_CLASS_NAME based on permissionsResponse.permissions.perDataset[datasetId]?.isUserStudy                                              
  - Returns 'userdataset' if isUserStudy is true, otherwise 'dataset'                                                                                                 
  - Returns undefined while permissions are loading (which causes the hook to wait)                                                                                   
  - Added STUDY_RECORD_CLASS_NAME to the dependency array                                                                                                             
                                                                                                                                                                      
  2. useWdkStudyReleases (line 161):                                                                                                                                  
                                                                                                                                                                      
  - Now calls usePermissions() to get permissions data                                                                                                                
  - Extracts datasetId from studyRecord.id[0]?.value                                                                                                                  
  - Dynamically sets STUDY_RECORD_CLASS_NAME using the same logic as above                                                                                            
  - Added STUDY_RECORD_CLASS_NAME to the dependency array

### DIFF
--- a/packages/libs/eda/src/lib/core/hooks/study.ts
+++ b/packages/libs/eda/src/lib/core/hooks/study.ts
@@ -31,8 +31,6 @@ import { usePermissions } from '@veupathdb/study-data-access/lib/data-restrictio
 import { FetchClientError } from '@veupathdb/http-utils';
 import { useCachedPromise } from './cachedPromise';
 
-const STUDY_RECORD_CLASS_NAME = 'dataset';
-
 interface StudyState {
   studyRecordClass: StudyRecordClass;
   studyRecord: StudyRecord;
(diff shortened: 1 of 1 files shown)


bobular:
In https://github.com/VEuPathDB/web-monorepo/commit/2eb48da7898071f99cda9ab242dce2c2eb4b2562 are `isDatasetAUserStudy` and `isUserStudy` now redundant? Maybe ask Claude to do a deep dive to check?

bobular (approved):
Looks good. `usePermissions` seems to be cached under the hood so we have no problems calling it from all over. See my comment about one of the older commits.

aurreco-uga:
> In [2eb48da](https://github.com/VEuPathDB/web-monorepo/commit/2eb48da7898071f99cda9ab242dce2c2eb4b2562) are `isDatasetAUserStudy` and `isUserStudy` now redundant? Maybe ask Claude to do a deep dive to check?

yeah seems weird but i didnt want to change unless needed.. so left it as is

aurreco-uga:
> Looks good. `usePermissions` seems to be cached under the hood so we have no problems calling it from all over. See my comment about one of the older commits.

so we make the call but it is cached where? not in browser storage.. should not be cached for too long right?  user datasets can change any time...

PR #1675, opened 2026-04-25
Add manual git hooks setup instructions to ensure consistent linting/prettier application

bobular:
Fix pre-commit hook setup for new repo checkouts.

The `postinstall` script was calling `husky` rather than `husky install` (the correct husky v8 command), so the git hooks may never have been installed automatically. Even with the correct command, yarn's `postinstall` only fires when the dependency tree changes, so devs who don't add/remove packages after cloning would still miss it.

Added a manual `npx husky install` step to the README as the reliable solution. Also reorganised the README to surface the install steps more prominently and moved the Cedar Server section out of Setup.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -117,6 +117,17 @@ From anywhere in the repo:
 yarn
 ----
 
+=== Set up git hooks
+
+Run this once after cloning to enable pre-commit formatting checks:
+
+[source, shell]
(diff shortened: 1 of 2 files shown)


bobular:
Merging as it's tooling and documentation only. Can't break anything, right?!

PR #1678, opened 2026-05-06
wdk-client package tsc 4.5.4 error fixes

Foxcapades:
Resolves compile errors that are encountered when compiling the wdk-client package using typescript `^4.5.4`.

### DIFF
--- a/packages/libs/wdk-client/src/StoreModules/QuestionStoreModule.ts
+++ b/packages/libs/wdk-client/src/StoreModules/QuestionStoreModule.ts
@@ -84,6 +84,7 @@ import {
   mergeMapRequestActionsToEpic as mrate,
 } from '../Utils/ActionCreatorUtils';
 import { GlobalParamMapping, ParamValueStore } from '../Utils/ParamValueStore';
+import { ServiceError } from '../Service/ServiceError';
 
 export const key = 'question';
 
@@ -566,7 +567,7 @@ const observeLoadGroupCount: QuestionEpic = (action$, state$, { wdkService }) =>
       }
 
       return from(
-        [{ name: '__total__' }]
+        [{ name: '__total__' } as ParameterGroup]
           .concat(questionState.question.groups)
           .filter(
             (group) =>
@@ -1059,7 +1060,7 @@ async function loadQuestion(
       submissionMetadata,
     });
   } catch (error) {
-    return error.status === 404
+    return error instanceof ServiceError && error.status === 404
       ? questionNotFound({ searchName })
       : questionError({ searchName });
   }
--- a/packages/libs/wdk-client/src/StoreModules/StepAnalysis/StepAnalysisObservers.ts
+++ b/packages/libs/wdk-client/src/StoreModules/StepAnalysis/StepAnalysisObservers.ts
@@ -69,7 +69,8 @@ import {
 import { transitionToInternalPage } from '../../Actions/RouterActions';
 import { StepAnalysisType } from '../../Utils/StepAnalysisUtils';
 import { InvalidStepValidation, extractParamValues } from '../../Utils/WdkUser';
-import { getTypedError } from '../../Utils/Errors';
+import { getTypedError, makeCommonErrorMessage } from '../../Utils/Errors';
+import { ServiceError } from '../../Service/ServiceError';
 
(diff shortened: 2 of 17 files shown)


aurreco-uga (approved):

packages/libs/wdk-client/src/Views/ReporterForm/WdkServiceJsonReporterForm.tsx
```diff
@@ -22,15 +22,17 @@ type Props<T, U> = {
   recordClass: RecordClass;
   summaryView: string;
   ontology: Ontology<CategoryTreeNode>;
-  formState: any;
-  formUiState: any;
+  formState: T & any;
+  formUiState: U & any;
```
bobular, line 26: These can both stay as `any`

`X & any` is always `any`

The deeper problem is a mix of JS and TS and deep down even in packages/libs/wdk-client/src/StoreModules/DownloadFormStoreModule.ts, formState and formUiState are `any` anyway:

https://github.com/VEuPathDB/web-monorepo/blob/88dbde8d6c47fe74964b7a618aeae3309931eee7/packages/libs/wdk-client/src/StoreModules/DownloadFormStoreModule.ts#L36-L37

packages/libs/wdk-client/src/Utils/IterableUtils.ts
```diff
@@ -224,7 +224,7 @@ export class Seq<T> {
     for (let iter = this[Symbol.iterator](); ; ) {
       let { done, value } = iter.next();
       if (done) break;
-      fn(value);
+      fn(value as T);
```
bobular, line 227: This cast can be avoided by keeping the discriminated union intact until the `done` case has been eliminated.

```
  forEach(fn: (t: T) => void) {
    for (let iter : Iterator<T> = this[Symbol.iterator](); ; ) {
      const item = iter.next();
      if (item.done) break;
      fn(item.value);  // value is now T
    }
    return this;
  }
```

or replacing the idiomatic `for` loop:

```
  forEach(fn: (t: T) => void) {
    const iter : Iterator<T> = this[Symbol.iterator]();

    while (true) {
      const item = iter.next();
      if (item.done) break;
      fn(item.value);
    }

    return this;
  }
```

bobular (commented):
Here are my comments on the non-Error-related TypeScript changes.

I'm still looking at the Error-related stuff.

packages/libs/wdk-client/src/StoreModules/QuestionStoreModule.ts
```diff
@@ -566,7 +567,7 @@ const observeLoadGroupCount: QuestionEpic = (action$, state$, { wdkService }) =>
       }
 
       return from(
-        [{ name: '__total__' }]
+        [{ name: '__total__' } as ParameterGroup]
```
bobular, line 570: not ideal but better than a bigger refactor - the `__total__` sentinel is handled just a few lines below so won't cause issues further afield.

packages/libs/wdk-client/src/StoreModules/StepAnalysis/StepAnalysisObservers.ts
```diff
@@ -237,7 +241,7 @@ export const observeStartLoadingChosenAnalysisTab = (
             parseAnalysisInitializationError(ex);
 
           if (!isValidationError) {
-            wdkService.submitErrorIfNot500(ex);
+            wdkService.submitErrorIfNot500(getTypedError(ex));
```
bobular, line 244: this wrapping isn't needed

packages/libs/wdk-client/src/Utils/ClientPlugin.tsx
```diff
@@ -122,7 +123,7 @@ function makeCompositePluginComponentUncached<T>(
     if (resolvedReferences == null) return <>{props.fallback}</> ?? null;
 
     if ('error' in resolvedReferences) {
-      return resolvedReferences.error.status === 404 ? (
+      return (resolvedReferences.error as ServiceError).status === 404 ? (
```
bobular, line 126: would be better as
`resolvedReferences.error instanceof ServiceError && resolvedReferences.error.status === 404`

bobular (commented):
All good - just a few minor comments.

PR #1685, opened 2026-05-13
Pre-select featured/only variables in notebook computations and visualizations

bobular:
## Demo/testing

When out on beta, this is a good demo 

https://beta.plasmodb.org/plasmo.beta/app/search/transcript/GenesByRNASeqEvidence#GenesByRNASeqpfal3D7_pf3d7_IDC2020_Subudhi_ebi_rnaSeq_RSRCDESeq

Because
* it's stranded so we get to see the sense variable auto-selected
* too many `label` annotations for PCA overlay, only one suitable annotation `timepoint` remains and this is auto-selected (and it looks pretty)
* **the PCA is beautiful**
* `label` is still auto-selected for the DESeq2 group comparison.

## Changes

In the DE and Antibody Array notebook PCA cells, the overlay variable selector now auto-populates on load:

- **`autoSelectFeatured`** — picks the first `isFeatured` variable that satisfies the overlay constraints. The user can clear the selection with the existing × button, and it won't re-assert itself.
- **`autoSelectWhenPossible`** — if there is only one valid overlay choice (e.g. a dataset with a single metadata variable like `timepoint`), it is selected automatically and rendered as read-only.

Both behaviours are gated on `hasComputedXAxis && hasComputedYAxis` in `ScatterplotVisualization`, so they only activate when the axes are locked to PCA components and the overlay is the only free input.

`autoSelectFeatured` is also applied in two further places:

- The **shared compute inputs cell** (gene identifier / measurement type), so featured expression variables are pre-filled there.
- The **DESeq2/Limma configuration panel**, so the `comparatorVariable` (the metadata variable used to define comparison groups) is pre-filled with the first featured variable that satisfies its constraints.

Also did:
- Made the `AnnotationPanel` (scatterplot mouse-over annotations) be vertically resizable
- Added a `useDebounceTrailing` hook to complement the leading+trailing `useDebounce` - which may be useful (see below)

## Limitations
The auto-select of the PCA scatterplot overlay causes two backend viz requests: one without and one with the overlay configuration. This is now solved: :heavy_check_mark: in https://github.com/VEuPathDB/web-monorepo/pull/1685/commits/465a72a72eca8a3ce013a5adc5baec013ef8e3b0


bobular:
If you would like to test this before the database is updated, you can apply this patch (don't commit it).

It adds the featured variable annotation to the sense variable and hopefully most `label` annotation vars.

Save the contents below into `/tmp/patch.diff` and then from the root of this repo do: `git apply /tmp/patch.diff`

```diff
diff --git a/packages/libs/eda/src/lib/core/api/SubsettingClient/index.ts b/packages/libs/eda/src/lib/core/api/SubsettingClient/index.ts
index 137f8ba8d1..ddaeebe28b 100644
--- a/packages/libs/eda/src/lib/core/api/SubsettingClient/index.ts
+++ b/packages/libs/eda/src/lib/core/api/SubsettingClient/index.ts
@@ -42,6 +42,11 @@ export default class SubsettingClient extends FetchClientWithCredentials {
           const { study } = await ioTransformer(StudyResponse)(res);
           assertValidStudy(study);
           orderVariables(study);
+          // TODO: remove — spoofing isFeatured for front-end testing
+          spoofFeaturedVariables(study, [
+            { entityId: 'GENE_EXPRESSION_RNASEQ_DATA', variableId: 'SEQUENCE_READ_COUNT_SENSE' },
+	    { entityId: "ENT_8151325d", variableId: "VAR_64c65374" }
+          ]);
           return study;
         },
       })
@@ -136,6 +141,27 @@ function orderVariables(study: StudyMetadata) {
   return study;
 }
 
+// TODO: remove — spoof for front-end featured-variable testing
+function spoofFeaturedVariables(
+  study: StudyMetadata,
+  targets: Array<{ entityId: string; variableId: string }>
+) {
+  for (const entity of preorder(
+    study.rootEntity,
+    (entity) => entity.children ?? []
+  )) {
+    for (const variable of entity.variables) {
+      if (
+        targets.some(
+          (t) => t.entityId === entity.id && t.variableId === variable.id
+        )
+      ) {
+        (variable as any).isFeatured = true;
+      }
+    }
+  }
+}
+
 function assertValidStudy(study: StudyMetadata) {
   const entitiesWithoutVariables: StudyEntity[] = [];
   for (const entity of preorder(
```

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterPlotAnnotationTooltip.scss
```diff
@@ -3,9 +3,12 @@
   background: #fafafa;
   font-size: 12px;
   height: 200px;
+  min-height: 60px;
   width: 400px;
   display: flex;
   flex-direction: column;
+  overflow: hidden;
+  resize: vertical;
```
bobular, line 11: This was just an easy nice-to-have that I slipped in. The mouse-over annotation panel can now be resized (useful in antibody arrays with lots of annotations)

packages/libs/eda/src/lib/core/components/visualizations/implementations/ScatterplotVisualization.tsx
```diff
@@ -770,6 +789,18 @@ function ScatterplotViz(props: VisualizationProps<Options>) {
           computedOverlayVariableDescriptor,
         };
 
+  const isDataReady = dataRequestDepsBase != null;
+
+  useEffect(() => {
```
bobular, line 794: This part adds a small react-query-disabling delay to prevent multiple `scatterplot` back-end requests (see PR description) when the auto-select effects in `InputVariables` fire.

packages/libs/eda/src/lib/core/hooks/debouncing.ts
bobular, line 1: We didn't use this but it might be a useful addition anyway.

PR #1688, opened 2026-05-13
UserDataset download component

steve-fischer-200:
Add the user dataset download component to the user datasets Download tab in EDA and in the User Datasets record page


bobular:
Useful testing guidance here: https://github.com/VEuPathDB/web-monorepo/blob/db44ccfc7d15f46260e82e35f83e87869c10e294/implementation-plan-user-dataset-files.md#manual-testing-urls

bobular:
I tried the genomics-site test first - having some problems getting to the right test pages.

I set up my genomics-site dev server pointing at beta.plasmodb.org (while on this branch, of course)

Uploaded a phenotype dataset: https://github.com/VEuPathDB/vdi-plugin-wrangler/tree/main/tests/testthat/phenotype/10-performance-10k-genes-OK (data.tsv) as **pheno-good**

I see it in My Datasets (http://localhost:8080/a/app/workspace/datasets):

image

If I click on "pheno-good" I get this: http://localhost:8080/a/app/workspace/datasets/0ZE5hEsV9c1I1

image

Then I can get to the search via the "Phenotype user datasets" link to http://localhost:8080/a/app/search/transcript/GenesByPhenotypeUserDataset?param.eda_dataset_id=EDAUD_0ZE5hEsV9c1I1

Which gives me the EDAUD_xxx ID I need to plug in to the URL in the implementation.md file: https://plasmodb.org/plasmo.beta/app/record/userdataset/EDAUD_45ILWccdqB so that gives me

http://localhost:8080/a/app/record/userdataset/EDAUD_0ZE5hEsV9c1I1

Which doesn't have a download section...?

Although the "Browse and Subset Data" link at the bottom takes me to vanilla EDA:

http://localhost:8080/a/app/workspace/analyses/EDAUD_0ZE5hEsV9c1I1/new

Where the download tab DOES have the new feature :-)

image

I guess the appropriate test path for genomics is
1. upload some phenotype data
2. find the EDAUD_xxx ID from the "Phenotype user datasets" link on the "My Dataset" page
3. go to `/a/app/workspace/analyses/EDAUD_xxx/new` and navigate to the Download tab

packages/libs/eda/src/lib/workspace/DownloadTab/Downloads.tsx
```diff
@@ -237,6 +276,24 @@ export default function Downloads({
             analysisState={analysisState}
           />
         )}
+        {isUserStudy &&
+          (userDatasetFilesResult?.error ? (
+            <div className="error-message">
+              <h2>Data Files</h2>
+              <p>{userDatasetFilesResult.error}</p>
```
bobular: I know it's a pain but it's good practice to hide the raw error message from the user. Does this error get logged already or do we have to log it explicitly here? (Off the top of my head I'm not sure how we do this.)

packages/libs/web-common/src/components/records/DataFilesSection.tsx
```diff
@@ -0,0 +1,80 @@
...
+      id={name}
+      className={`wdk-RecordAttributeSectionItem`}
+      headerContent={headerContent}
+      isCollapsed={isCollapsed}
+      onCollapsedChange={onCollapsedChange}
+    >
+      <ErrorBoundary>
+        {userDatasetFilesResult?.error ? (
+          <div className="error-message">
+            <p>{userDatasetFilesResult.error}</p>
```
bobular: See above on surfacing raw errors. (Maybe these are already sanitized and user-facing?)

packages/libs/eda/src/lib/workspace/DownloadTab/Downloads.tsx
```diff
@@ -67,6 +70,39 @@ export default function Downloads({
...
+        console.error('Failed to fetch user dataset files:', error);
+        return {
+          data: null,
+          error: 'Failed to load dataset files. Please try again later.',
+        };
+      }
+    },
+    [isUserStudy, vdiDatasetId]
+  );
+
```
bobular: You should be able to ask for this to be refactored to a custom hook, as there's an almost identical copy elsewhere in this PR.

The only difference is the early return on `isUserStudy` not present in the other usage. A constant `true` could be passed to the hook in that case.

bobular (commented):
This looks and works great. (Though tbh I only tested in genomics-site.) There are some maintainability issues that need addressing though.

Firstly there's a hook that can DRY-up some repetition (see inline comments).

Then you should ask why `renderFileSection` in `packages/libs/user-datasets/src/lib/Components/Detail/UserDatasetDetail.tsx` doesn't make use of the new `UserDatasetFiles` component. I think it should. Otherwise we have dual implementations of the UD files download interface.

bobular:
That review was all human Bob.

I also just reviewed with CC using Opus. Here's what it found (verbatim). Looks like I caught all the biggies. Note also that we should remove the implementation plan before merge.

---
prompt:
> Hi, could you critically review this branch w.r.t. `main`? You can see the implementation plan .md file that was followed, but focus mainly on the actual code changes.

---

Here's my (Claude Opus in Claude Code) critical review of the branch. Overall, the feature is implemented and the second commit (`db44ccfc7d`) is a real improvement over the first — switching from inlining file fetching in the record-class file to a proper WDK attribute section is the right move. But the diff has several rough edges worth fixing before merge.

## Significant issues

**1. The same fetch block is duplicated in two places.** The ~30-line `useWdkService` closure (with the VDI compatibility check, try/catch, console.error, identical error strings) appears verbatim in both `Downloads.tsx:79-103` and `DataFilesSection.tsx:29-54`. This should be either:
- pulled into a small custom hook (`useUserDatasetFiles(vdiDatasetId)`) co-located with `UserDatasetFiles.tsx`, or
- moved *into* `UserDatasetFiles` so consumers just pass `datasetId` and the component owns its own loading/error states.

The latter would shrink both call sites to a single `<UserDatasetFiles datasetId={…} />`.

**2. `UserDatasetDetail` was not refactored to use the new component.** The plan explicitly said to extract `renderFileSection` + `getFileTableColumns` from `UserDatasetDetail.tsx` and reuse it. Instead the new file is a copy-paste of those methods (lines 521-640 of `UserDatasetDetail.tsx` are duplicated almost line-for-line in `UserDatasetFiles.tsx`). Now there are two implementations to keep in sync — any future fix to the original won't propagate.

**3. `installStatus="complete"` is hardcoded at both call sites — a download-button behavior regression.** In the original `UserDatasetDetail.tsx:606-610`, `enableDownload` for processed files is derived from `userDataset.status.install[...].data.status === 'complete'`. Both new call sites pass `"complete"` unconditionally, which means the Download button on install.zip is always enabled — even for datasets whose install actually failed or is still pending. That'll surface as a download attempt against a missing/incomplete artifact.

Either fetch install status alongside the file listing, or accept a real `installStatus` from the caller (in the EDA case, it's reachable via `studyMetadata`/permissions).

**4. `wdkRecordIdToDiyUserDatasetId` throws on non-`EDAUD_` IDs** (`diyDatasets.ts:6`).
- In `Downloads.tsx:75-77`, the call is guarded by `isUserStudy` — assumes `isUserStudy === true` always implies `EDAUD_` prefix. Probably true today, but it's an implicit coupling worth a comment, or a safer `isDiyWdkRecordId(datasetId)` check.
- In `DataFilesSection.tsx:25-26`, the call is unconditional. The `<ErrorBoundary>` will catch it, but if the section ever gets wired onto a non-userdataset record (admin misconfiguration), users will see a broken section rather than a missing one. A defensive `isDiyWdkRecordId(...)` guard would be cleaner.

**5. The "VDI service not configured" error path is user-visible in environments where it's expected to be unconfigured.** In WDK pages opened on a site where VDI is genuinely not wired in, every user-dataset record will now show a red error message. Is that actually what you want? In the EDA Downloads page the section is at least gated by `isUserStudy`, so if the site has no user studies, the error never appears — but on the record page, opening any user-dataset record on a non-VDI site will surface this. Consider rendering nothing in that case, the same way the original code does for non-user studies.

## Smaller issues

**6. Non-null assertion `datasetId={vdiDatasetId!}`** (`Downloads.tsx:287`). The outer `isUserStudy &&` guarantees it's set, but the `!` papers over the type system. Lifting the early return / using `vdiDatasetId != null` in the condition would let the type narrow naturally.

**7. Dead check in `DataFilesSection.tsx:31`.** `wdkRecordIdToDiyUserDatasetId` either throws or returns a non-empty string — `if (!vdiDatasetId)` can never be true. Remove.

**8. Header rendered twice on the WDK record page.** `DataFilesSection` wraps the children with `CollapsibleSection` whose `headerContent` is `DefaultSectionTitle(displayName="Data Files")` (assuming that's the attribute's display name in the record class XML). Inside, `UserDatasetFiles` renders its own `<h2>Data Files</h2>`. The user will see "Data Files" stacked on "Data Files". Either add a prop to suppress the inner `<h2>` when used inside a WDK section, or set `displayName` to something different.

**9. Unused props.** `UserDatasetFiles.tsx` declares `projectId?: string` (line 33) and never references it. `dataNoun.plural` is in the interface but never read. Drop them.

**10. Default `dataNoun` reads ungrammatical.** Default is `{ singular: 'dataset', plural: 'datasets' }`, which yields `"Uploaded Files in dataset"`. Either require the prop (no default) or change to `"Uploaded Files in your dataset"` / sentence-cased.

**11. `installStatus?: string`** (line 31) — currently any non-`'complete'` value disables downloads, including typos. Make it a union (`'complete' | 'pending' | ...`) or a boolean (`canDownloadInstall`).

**12. Inline `<style>` block at the bottom of `UserDatasetFiles.tsx`.** Works, but injects a fresh `<style>` tag per mount and bypasses the codebase's stylesheet conventions. Move to a `.scss` next to the component.

**13. Prettier-style whitespace changes are folded into the feature commit** in `UserDatasetRecordClasses.UserDatasetRecordClass.jsx` (tabs → spaces, line wrapping, trailing-comma additions). Makes the substantive diff harder to read; would have been better as a separate prep commit. Not blocking.

**14. No tests** for `UserDatasetFiles` or `DataFilesSection` despite the test plan in the spec. At minimum, a snapshot test of `UserDatasetFiles` with both upload/install present, neither present, and download-disabled would catch the `installStatus` regression in #3.

**15. Hook deps `[isUserStudy, vdiDatasetId]`** (`Downloads.tsx:103`) — `isUserStudy` is redundant since `vdiDatasetId === null` already encodes `!isUserStudy`. Cosmetic.

## What looks good

- Switching to a WDK attribute section (`DataFilesSection` + `case 'dataFiles'`) is much cleaner than the plan's original "stuff it into RecordHeading" approach. It correctly delegates section ordering/collapsing to WDK config.
- `useNonNullableContext(WdkDependenciesContext)` instead of threading `wdkService` through props (plan deviation) is the right call.
- `ErrorBoundary` wrapping in `DataFilesSection` is good defensive practice.

## Top-priority fixes

If I had to pick three things to land before merge: **#3** (the silent download-button regression), **#1** (deduplicate the fetch block), and **#8** (the duplicate "Data Files" heading on the record page).

packages/libs/eda/src/lib/workspace/DownloadTab/Downloads.tsx
bobular: Correction! The error strings are thrown by the client and are under our control and can only be
  - 'VDI service is not configured. Unable to load dataset files.'
  - 'Failed to load dataset files. Please try again later.'

packages/libs/web-common/src/components/records/DataFilesSection.tsx
bobular: Oops yes they are!

aurreco-uga (approved):

PR #1690, opened 2026-05-15
dataExplorer dataset upload form and vdi compatibility updates.

Foxcapades:
replaces the existing dataset upload form with an expanded form as outlined in https://www.figma.com/design/KVsfQDUWa8pCUi78BkfXjO/dataExplorer-website which makes use of additional dataset metadata and upload fields available in the VDI API


packages/libs/user-datasets/src/lib/Components/Upload/UploadForm/Components/SubmissionModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import { Modal } from "@veupathdb/coreui";
+import { Loading } from "@veupathdb/wdk-client/lib/Components";
+
+export interface SubmissionModalProps {
+  readonly submitting: boolean;
+  readonly uploadProgress: number;
+}
+
+export function SubmissionModal(props: SubmissionModalProps): ReactElement {
```
Foxcapades, line 11: This was extracted from the original `UploadForm.tsx` file

bobular (commented):
Looks and works great.

I'm not crazy about the two "Upload Dataset" submit buttons - I realise these were in the mockups. If one could press the top button without checking "no" on the three radio button groups, I would understand.

Collapsing the sections when clicking "no" would likely save the user scrolling to the bottom, so no need for the "easy-reach" button at the top?

**Edge case bug:**
(Phase 2 if not a trivial fix)
If I successfully submit a file after failing to complete all the fields (and I get the "Could not upload dataset: field-study-toggle: selection is required" warnings) - these warnings are sticky when I go back to the form to submit a **second dataset**.

Image

packages/libs/user-datasets/src/lib/Components/Upload/UploadForm/Components/InputPair.tsx
```diff
@@ -0,0 +1,152 @@
+import { Consumer } from '../../../../Utils';
+import { ChangeEvent, ReactElement, ReactNode } from 'react';
+import { FieldHelpText } from './FieldHelpText';
+
+interface BaseInputProps<T extends object = object> {
+  readonly label: string;
+  readonly fieldName: keyof T & string;
```
bobular, line 7: This gives the following in my IDE:

`Type 'string' is not assignable to type 'never'. ⮐ [2322]` from  `fieldName={isPrimary}` 
in `packages/libs/user-datasets/src/lib/Components/Upload/UploadForm/Sections/Core/CollaboratorsSection.tsx` for example.

I'm not sure we need the generic `T` at all?

bobular (commented):
Another thing for you to look at while I look at the store module :-)

packages/libs/user-datasets/src/lib/Components/Upload/UploadForm/Components/InputPair.tsx
Foxcapades: I'm not getting errors for that in my IDE, and that had `T` to guard against decoder type changes, and provide smart editor autocomplete when modifying those components.

For example
```ts
        <InputPair<PostCharacteristics>
          fieldName="participantAges"
```

If I tried to change `fieldName` to "foobar", it would be a compile error, because "foobar" is not a valid key for the type `PostCharacteristics`.


It's sort of future proofing in that if the API changes and someone updates the io-ts types, this will help catch possible errors due to missed references to that type or its properties.

bobular: Duh, restarted tsserver and it went away. Weird.

bobular (approved):
I've touched on all the main parts and it looks good. It obviously works great. The submission gate/modal coming up before the form validation check is just temporary, so no biggie.

PR #1700, opened 2026-05-27
Spike: MuPDF client-side deploy and test

bobular:
Note: not targeted at `main`!

**This is a proof of concept feasibility test of MuPDF.js - not intended for merging - educational only**

Added you as a reviewer just for interest, @ryanrdoherty 

To test, run `yarn && yarn nx start @veupathdb/mupdf-test`

---
### What we built

A minimal standalone React site (`packages/sites/mupdf-test`) that lets a user pick a PDF file and extracts its text entirely in the browser — no server involved. The text appears in a `<pre>` block on the page.

The core library is **mupdf** (v1.27), which ships a WebAssembly build of the MuPDF C library. WebAssembly is a low-level binary format that browsers can run near-natively — mupdf compiles its C PDF engine to it, so the full rendering and text-extraction logic runs client-side. The JS wrapper (`mupdf.js`) fetches and initialises the `.wasm` binary when first imported, then exposes a clean API: `Document.openDocument()`, `loadPage()`, `toStructuredText().asText()`.

The site itself is a stripped-down version of the genomics-site structure — same webpack/react-scripts tooling, but with no WDK, no VEuPathDB headers, no backend proxy, and no auth. This made it possible to develop and test during the SSL cert. outage that was blocking all dev/beta logins.

---

### Lessons learned

**1. mupdf types aren't found by default**

mupdf publishes its TypeScript types only via the `exports` field in `package.json`. TypeScript's default `moduleResolution: "node"` ignores that field. Fix: add a `paths` entry in `tsconfig.json` pointing directly at `node_modules/mupdf/dist/mupdf`, which bypasses the resolution problem without changing the base config.

**2. mupdf uses top-level await**

`mupdf.js` awaits the WASM initialisation at module scope (outside any function) — this is the ES module pattern for async-at-load-time libraries. Webpack 5 needs `experiments: { topLevelAwait: true }` to handle this correctly, otherwise it can't represent the suspension in its output bundle. `asyncWebAssembly: true` is also needed for the `.wasm` file itself. Both flags are now stable despite still living under `experiments`.

**3. mupdf's Node.js dead-code causes webpack errors**

`mupdf.js` and `mupdf-wasm.js` detect whether they're running in Node.js and conditionally use `node:fs`, `node:path`, `node:url`, `module`, etc. These branches never execute in a browser, but webpack analyses imports statically regardless of runtime conditions. Three things were needed:
- `NormalModuleReplacementPlugin` to strip the `node:` URI prefix (webpack 5.84 doesn't handle that scheme) so that `node:fs` → `fs` etc.
- `resolve.fallback: { module: false, url: false }` to suppress polyfills for Node built-ins not already covered by `site-webpack-config`
- The plugin must live in the same `plugins` array that gets passed to `configure()` — spreading `additionalConfig` into a new object and then overwriting the `plugins` key silently drops it

**4. `process/browser` resolution needs an absolute alias**

`site-webpack-config` uses webpack's `ProvidePlugin` to inject `process/browser` as the `process` global. When webpack injects this into code deep inside `node_modules/mupdf/dist/`, it resolves `process/browser` from that directory — and in some workspace configurations the standard upward traversal misses the workspace-root `process` package. Specifying it as an absolute path via `resolve.alias: { 'process/browser': require.resolve('process/browser') }` guarantees it's always found.

**5. Plugin ordering in webpack-merge**

When you spread `additionalConfig` into a new object literal and then declare a `plugins` key after the spread, JavaScript's spread semantics mean the explicit `plugins` array overwrites the one from `additionalConfig`. webpack-merge never sees the original plugins. Always merge plugin arrays explicitly: `plugins: [...(additionalConfig.plugins || []), ...newPlugins]`.

---

### Checklist: integrating mupdf into genomics-site

- [ ] Add `mupdf` to `packages/sites/genomics-site/package.json` devDependencies and run `yarn workspace @veupathdb/genomics-site add mupdf`
- [ ] Add to `genomics-site/webpack.config.js`:
  - `experiments: { asyncWebAssembly: true, topLevelAwait: true }`
  - `resolve.alias: { 'process/browser': require.resolve('process/browser') }`
  - `resolve.fallback: { module: false, url: false }`
  - `plugins: [ new webpack.NormalModuleReplacementPlugin(/^node:/, r => { r.request = r.request.replace(/^node:/, '') }) ]`
- [ ] Ensure those plugins aren't dropped when `webpack.config.local.mjs` builds its `plugins` array — spread `additionalConfig.plugins` explicitly
- [ ] Add `paths` to `genomics-site/tsconfig.json` for mupdf types (or accept that `moduleResolution: "node"` won't find them and use `@ts-ignore` with a comment)
- [ ] Add `PdfTextExtractor` component (or a production equivalent) to the genomics-site component tree
- [ ] Add a route in `routes.jsx` with `requiresLogin: false` if the feature should be publicly accessible
- [ ] Decide on UX: inline upload widget, modal, or dedicated page
- [ ] Test bundle size impact — the mupdf WASM binary is large (~7 MB uncompressed); confirm it's lazy-loaded via the dynamic `import('mupdf')` and not included in the initial bundle


packages/sites/mupdf-test/webapp/js/PdfTextExtractor.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [error, setError] = useState<string | null>(null);
+  const [loading, setLoading] = useState(false);
+
+  async function handleFile(file: File) {
+    setError(null);
+    setText(null);
+    setLoading(true);
+    try {
+      const arrayBuffer = await file.arrayBuffer();
+      const mupdf = await import('mupdf');
```
ryanrdoherty, line 15: Seems like this should be done after the form loads, in preparation for the user selecting a file, no?  Rather than make them wait more here.

ryanrdoherty (approved):
Pretty cool.  Have never done anything with WebAssembly- neat that it is so well integrated.  Seems like a good way to go for the PDF extraction.

packages/sites/mupdf-test/webapp/js/PdfTextExtractor.tsx
bobular: Yes, that would make things a bit smoother for the user.

bobular:
Closing as this will never be merged. It was just a demo/spike.

PR #1704, opened 2026-05-28
filter out 'empty' contact objects before form submission

### LINKED ISSUE #1703: User Dataset Upload Form: contact objects are not being scrubbed before upload.
This leads to invalid submissions due to incomplete objects that should have been removed from the request.

Example POST request dataset details:
```json
{
  "type": {
    "name": "genelist",
    "version": "1.0"
  },
  "origin": "direct-upload",
  "visibility": "private",
  "installTargets": [
    "PlasmoDB"
  ],
  "contacts": [
    {
      "middleName": "",
      "firstName": "asdfasdfasdf",
      "lastName": "asdfasdfasdf"
    },
    {
      "middleName": "",
      "isPrimary": false
    }
  ],
  "name": "asdfasdfas",
  "summary": "asdfasdfasdf"
}
```

Foxcapades:
Presently, if a user creates a contact object, edits it, then blanks it out, we are submitting a contact object with empty values.

This filters out contact objects that contain no truthy values.

### DIFF
--- a/packages/libs/user-datasets/src/lib/Service/process/create-dataset.ts
+++ b/packages/libs/user-datasets/src/lib/Service/process/create-dataset.ts
@@ -1,5 +1,4 @@
 import {
-  DatasetContact,
   DatasetPostDetails,
   DatasetPostResponseBody,
   DatasetUploads,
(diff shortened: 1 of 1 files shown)


packages/libs/user-datasets/src/lib/Service/process/create-dataset.ts
```diff
@@ -95,6 +118,17 @@ function scrubDatasetCharacteristics(
   };
 }
 
+/**
+ * Tests if a given object contains truthy values.
+ */
+function isEmptyObject(obj: Record<string, any>): boolean {
+  for (const key of Object.keys(obj))
+    if (obj[key])
```
ryanrdoherty, line 126: Are empty strings falsy?  Maybe should trim those too if easy.

ryanrdoherty (approved):

packages/libs/user-datasets/src/lib/Service/process/create-dataset.ts
Foxcapades: they are falsy, but vdi already removes those.  Empty strings are the reason the 'empty' objects were being submitted when they shouldn't.  When the user edited the fields, then cleared their edits, the contact object was left with empty strings in it.

PR #1718, opened 2026-05-29
filter char from Contents

steve-fischer-200:


bobular (approved):
Code looks clean. Haven't tested.

No need to change the visibility filter in `packages/sites/clinepi-site/webapp/js/client/component-wrappers/index.js` as I had suggested in Slack, as this does filtering on the `categoryTree` - nothing to do with visibility flags

PR #1740, opened 2026-06-04
Fix: render multi-line database text with line breaks

bobular:
Multi-line text stored in the database was being displayed as one continuous block of text. This is because HTML collapses whitespace by default — the line breaks in the raw data were simply ignored by the browser.

### What changed

Added a new `renderAttributeValueAutoBreak` helper in `ComponentUtils.tsx` that converts ALL line breaks (`\n`) in attribute values into `<br/>` elements before rendering **plain text**. If the attribute value already contains HTML it is not processed in this way.

The helper is now used in two places:

- **`RecordTable.jsx`** — single-column tables rendered as a `<ul>` list (`wdk-RecordTableList`)
- **`RecordAttribute.jsx`** — block-level attribute values on record pages (via `BlockRecordAttributeSection`)

The underlying `renderAttributeValue` function is unchanged and all existing call sites are unaffected.

### Next steps

**Other call sites:** `renderAttributeValue` is called in around 15 places across the codebase. It's worth checking whether any of them display free-text database content that could benefit from the same treatment. Key candidates to look at: `AnswerTableCell.jsx`, `RecordAttribute.jsx` (already done), and any site-level wrappers in `genomics-site` and `ortho-site`.

**Could autoBreak break anything?** The conversion inserts raw `<br/>` tags into the string, which then goes through `DOMPurify.sanitize()` before rendering — so there's no XSS risk. However, it's worth checking: if any attribute values already contain HTML (e.g. they arrive from the database with `<br/>` or `<p>` tags already in them), the double conversion could produce extra spacing. Those values would already be going through the `dangerouslySetInnerHTML` path, so visually inspect a few rich-HTML attributes to confirm nothing looks doubled-up.

The new autoBreak could potentially detect existing HTML tags and do nothing in that case? **Yes** we have now implemented that.


aurreco-uga (approved):

Foxcapades:
> Single line breaks are left alone, since those are typically soft wraps with no display intent.

Single line breaks that are coming from a user were manually inserted by that user.  I would think that the user's intent in that situation would actually be to create the line break they inserted.

bobular:
I've made the changes as discussed in scrum.
* single `\n` -> `<br/>`
* if and only if it's plain text (no HTML or HTML entities)

Currently this "filter" is deployed in two generic WDK record display situations:
* `RecordAttribute`
* Single attribute record tables (whatever they are!)

The two multiline items are currently using one of each (see below, respectively).

However, I believe we will be switching both to regular `RecordAttribute` - in which case do we want to undo the change on single attribute record tables?

image

PR #1742, opened 2026-06-05
Render variable definition as sanitized HTML on Browse and Subset page

jbrestel:
## Summary
- Variable `definition` on the Browse and Subset page now renders as sanitized HTML instead of plain text
- Uses `safeHtml` from `wdk-client/Utils/ComponentUtils` (DOMPurify under the hood), consistent with the pattern established in #1731
- No new dependencies required

## Test plan
- [ ] Open the Browse and Subset page for a study
- [ ] Confirm a variable with a plain-text definition still displays correctly
- [ ] Confirm a variable with HTML in the definition (e.g. `<b>bold</b>`, `<a href="...">link</a>`) renders correctly
- [ ] Confirm malicious HTML (e.g. `<script>`) is stripped by DOMPurify

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bobular (approved):
That looks good. `safeHtml` is now actually safe, which is useful!

PR #1757, opened 2026-06-11
fix upload form rapid redraw due to lack of state equality check

Foxcapades:
the form state being grabbed via `useSelector` is a complex object which was resulting in `useSelector` triggering a re-render on every call as the object instance was changing, but its value was not.


aurreco-uga (approved):

Foxcapades:
converted back to draft due to bug uncovered in testing

Foxcapades:
bug squashed by way of deleting custom code :tada:

PR #1761, opened 2026-06-16
Allow saving/loading of organism param value set across questions (user pref by project)

### LINKED ISSUE #1616: organism parameter: save selected values as a user preference, let them choose this set when they go to a new search.
from DR slack Feb 25 2026

Couple of notes regarding how to allow users to select list of species once and apply throughout strategy.  This might be easiest to implement … although note that we should discuss further before deciding whether/if/when to prioritize1:
- Could provide user with a button to click to “Copy this species list” (or possibly “Save this species list”) … in either or both of the places shown (screenshot of the organism parameter  in search page and upon revise)
- Then allow user to paste copied (or saved) list into another query

ryanrdoherty:


packages/libs/wdk-client/src/Views/Question/Params/TreeBoxEnumParam.tsx
```diff
@@ -184,13 +185,16 @@ export function TreeBoxEnumParamComponent(props: TreeBoxProps) {
       ? checkboxTreeProps
       : props.wrapCheckboxTreeProps(checkboxTreeProps);
 
+  const postSelectionElement = props.postSelectionElement || <></>;
```
bobular: It should be OK just to render the null prop? Or at least fall back to `null` rather than `<></>`. `null` is probably a bit more commonly understood to mean "don't render anything".

packages/libs/preferred-organisms/src/lib/components/OrganismParam.tsx
```diff
@@ -145,6 +175,34 @@ function TreeBoxOrganismEnumParam(
...
+  const selectOrgValuePref: React.MouseEventHandler<HTMLButtonElement> = (
+    event
+  ) => {
+    event.preventDefault();
+    if (localOrganismValuePreset == null) return;
+    const trimmedPresets = _.intersection(
+      localOrganismValuePreset,
+      Array.from(leafTerms)
+    );
+    wrappedOnChange(trimmedPresets);
```
bobular: If you want to avoid the annoying gate, maybe just call `onChange(trimmedPresets)` here?

bobular (approved):
This looks good. Just a few comments before I take the day off tomorrow.

PR #1765, opened 2026-06-24
Add Experimental Organism to dataExplorer with toggle, keep genomics with no toggle

aurreco-uga:


Foxcapades (approved):

aurreco-uga:
i had to hardcode the dataExplorer project to render or not the toggle...
but maybe ellie may come up with a way to deal with this in vdi config.yml 
or maybe  an improvement is to check for the existence of the Field Charact section instead of project name...

PR #1768, opened 2026-06-30
Add plasmoFAST client-side tool for PlasmoDB

bobular:
Adds a new `/plasmoFAST` tool that detects *Plasmodium falciparum* lab strains directly from FASTQ (or gzipped `.fastq.gz`) sequencing files. Analysis runs entirely in the browser via the `@veupathdb/plasmofast` library (k-mer counting in a Web Worker), so user data is never uploaded. The page shows a live progress bar with reads processed, a Stop button to cancel a run, and a per-strain results table that highlights the most likely strain. It's added to the Tools menu (PlasmoDB only) just above PlasmoAP.

- New route + `PlasmoFast` page/controller, modelled on the existing PlasmoAP tool
- Uses coreui buttons and wdk-client's `FileInput` for UI consistency
- Adds the `@veupathdb/plasmofast` dependency

### DIFF
--- a/packages/sites/genomics-site/package.json
+++ b/packages/sites/genomics-site/package.json
@@ -100,5 +100,8 @@
   },
   "files": [
     "dist"
-  ]
+  ],
+  "dependencies": {
+    "@veupathdb/plasmofast": "0.2.0"
+  }
 }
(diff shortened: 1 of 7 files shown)


ryanrdoherty (approved):
Would love to understand everything going on here.  Need to read up on all the basic react hooks.

bobular:
@ryanrdoherty 
Oops I noticed after merging that the two new dependencies weren't devDependencies.
It will still work but I'll resolve this tomorrow.

PR #1774, opened 2026-07-02
Allow users to expand only treebox branch nodes that have selected leaves

### LINKED ISSUE #1615: organism parameter (tree): add option to open the tree to show selected options (after collapsing it)
from DR on slack Feb 25 2026

, on another note (which we def should do, although not urgent), in addition to “select all | clear all | expand all | collapse all” on species tree, we should also provide a “show all selected” option, returning user to default view.  If I choose Expand All (or Collapse All), there is no way I can get back to the default view.  It is common that I want to Expand all, scroll through the tree, selecting as I go, and then review to see if I want to add or remove others … but only way to do this is to save and select revise again

ryanrdoherty:


bobular (approved):
Haven't run/tested it but it LGTM - good call on the lodash imports.

PR #1787, opened 2026-07-08
cleanup patch request using post request scrub function, correct form…

Foxcapades:
… validation regarding dataset properties files

### DIFF
--- a/packages/libs/user-datasets/src/lib/Components/Update/UpdateForm.tsx
+++ b/packages/libs/user-datasets/src/lib/Components/Update/UpdateForm.tsx
@@ -17,11 +17,12 @@ import { useDatasetFormState } from '../../StoreModules/UserDatasetUploadStoreMo
 import { isDatasetFormValid } from '../../Common/Forms/form-validation';
 import { DatasetFormProps } from '../../Common/Forms/DatasetFormProps';
 import { isEmpty, isEqual } from 'lodash';
-import { DatasetUploads, PartialDatasetDetails } from '../../Service';
+import { DatasetGetResponseBody, DatasetUploads, PartialDatasetDetails } from '../../Service';
 import { hasUploads } from '../../Service/Model/utility-types';
 import { DatasetTypeConfig } from '../../Common/Configuration';
 
 export interface UpdateFormProps extends DatasetFormProps {
+  readonly originalDataset: DatasetGetResponseBody;
   readonly originalDetails: PartialDatasetDetails;
   readonly scrollContainerRef: RefObject<HTMLElement>;
 }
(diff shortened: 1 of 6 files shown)


packages/libs/user-datasets/src/lib/Components/Update/UpdateForm.tsx
```diff
@@ -56,12 +57,20 @@ export function UpdateForm(props: UpdateFormProps): ReactElement {
 
   const updateSection = useRef<HTMLElement>(null);
 
+  const isMissingDatasetProperties = useMemo(
```
steve-fischer-200, line 60: consider this to avoid re-rendering when props change:

```
const isMissingDatasetProperties = useMemo(
    () => datasetDetails.visibility === 'public'
      && props.formConfig.dataType.vdiConfig.usesDataProperties
      && isEmpty(fileUploads.dataPropertiesFiles)
      && isEmpty(props.originalDataset.files.datasetProperties),
    [
      datasetDetails.visibility,  // Just the specific property
      props.formConfig.dataType.vdiConfig.usesDataProperties,  // The exact nested value
      props.originalDataset.files.datasetProperties,  // The exact nested value
      fileUploads.dataPropertiesFiles  // The specific property
    ]
  );
```

packages/libs/user-datasets/src/lib/Components/Update/UpdateForm.tsx
steve-fischer-200: remove this old function

packages/libs/user-datasets/src/lib/Service/process/update-dataset.ts
```diff
@@ -38,8 +42,30 @@ export interface UpdateResult {
 export async function submitUpdate(submission: UpdateSubmission): Promise<UpdateResult> {
   const { vdi, datasetId } = submission;
 
+  const mutableSubmission: Mutable<PartialDatasetDetails> = { ...submission.updated };
+
+  if (!submission.formState.hasExternalSources) {
```
steve-fischer-200, line 47: use `=== false` if undefined values are possible

steve-fischer-200 (commented):
i added three comments

packages/libs/user-datasets/src/lib/Service/process/update-dataset.ts
Foxcapades: Undefined is equivalent to false in this context.

packages/libs/user-datasets/src/lib/Components/Update/UpdateForm.tsx
Foxcapades: I included `props` itself just to to silence the linter warning, if any of the values in the props object changed there would be a component rerender regardless.

steve-fischer-200 (approved):

PR #1798, opened 2026-07-13
Expand project constants

Foxcapades:
Expand the "project constants" to
* include both the project ID and project display name as separate values
* unify and relocate the site url creation as part of each project's config

### DIFF
--- a/packages/libs/wdk-client/src/Utils/ProjectConstants.ts
+++ b/packages/libs/wdk-client/src/Utils/ProjectConstants.ts
@@ -4,7 +4,9 @@ export interface ProjectConstants {
   readonly siteUrl: string;
 }
 
+// WARNING!!! The site footer depends on the order of this array!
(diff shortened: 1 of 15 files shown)


aurreco-uga (approved):

packages/libs/wdk-client/src/Utils/ProjectConstants.ts
```diff
@@ -1,40 +1,61 @@
...
+export function isGenomicsProjectId(projectId: string): boolean {
+  return GENOMICS_PROJECTS.some((it) => it.projectId === projectId);
+}
+
+export function projectIdToDisplayName(projectId: string): string | undefined {
+  return ALL_VEUPATHDB_PROJECTS.find((it) => it.projectId === projectId)
+    ?.displayName;
+}
+
+function newProjectConstants(
```
bobular, line 48: Oh, you've merged already. I was going to suggest the singular form for this function name. Otherwise all looks good. Thanks for cleaning up this dirty corner!

PR #1803, opened 2026-07-15
fix(web-common): render EdaDatasetGraph data table when a dataset has no graph

jbrestel:
## Problem

On the gene record page, a phenotype dataset with **no configured graph** (e.g. `PlasmoDB_pfal3D7_HS_Response_Phenotype_Phenotype_RSRC`) shows a row in the "Phenotype Datasets" table but renders **neither a graph nor its data table** — even though the gene has data for that dataset.

## Root cause

`EdaDatasetGraph` type-guards its props and does `return null` if `plot_configs_json` is not a string:

```ts
if (typeof plot_configs_json !== 'string' || ... ) { console.error('EdaDatasetGraph: bad props'); return null; }
```

A dataset with no `edagenegraph` config produces a SQL `NULL` for `plot_configs_json` (via the `LEFT JOIN` in `GeneTables.EdaPhenotypeDatasets`), which WDK serializes as JSON `null`. The guard then bails out of the **entire** child row before reaching the data-table block — so the graph *and* the (graph-independent) data table both disappear.

Confirmed against production data for `PF3D7_0102100`:

| Row | `plot_configs_json` | Renders |
|-----|---------------------|---------|
| HS_Response (no edagraph) | `null` | ❌ nothing |
| piggyBac (has edagraph) | valid JSON string | ✅ graph + data table |

The data-table source (`EdaPhenotypeGraphsDataTable`) *does* have rows for the graph-less dataset, so this is purely the early return.

## Fix

Drop `plot_configs_json` from the fatal guard and treat a non-string value as "no graph": `plotConfigs` becomes `undefined`, `showGraph` is `false`, the graph block is skipped, and the data table renders normally. Identity fields (`dataset_id`, `source_id`, `graph_ids`) remain required. All other uses of `plotConfigs` in the component already handle `undefined`.

This lets the existing "Phenotype Datasets" table show data for **every** phenotype dataset, graphed or not.

## Verification

- `tsc --noEmit` on `web-common`: no errors introduced (changed file clean; pre-existing errors are in unrelated packages).
- `web-common` has no unit-test harness, so no automated test was added. Behavior traced by hand; live verification requires this package to flow into a site build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bobular (approved):
Logic looks good!

PR #1808, opened 2026-07-21
DE notebook - fix incorrect auto-selection locking out valid Gene identifier options in shared-inputs cells

bobular:
### The bug

The Differential Expression notebook's "Select Expression Data" cell auto-selects sensible defaults for its inputs (Gene identifier, Measurement type) to save users a click when there's only one reasonable choice. On studies with a single RNASeq dataset this works fine.

On studies with **multiple RNASeq datasets** (e.g. a dual host/pathogen experiment with separate `hsapREF` and `pfal3D7` entities), the auto-selection logic _for featured variables_ picked a default *Measurement type* before the user had chosen which dataset/organism they meant. Because Gene identifier and Measurement type are required to come from the same dataset, that early pick silently eliminated the other dataset's Gene identifier from consideration — leaving users with a single, read-only "only option" dropdown showing the wrong organism's gene, with no way to switch (other than cancelling the measurement type with the :x: button, which is not obvious or desirable). In the reported case, a P. falciparum gene search defaulted to a human reference gene.

This wasn't a data problem or a one-off glitch — the code was doing exactly what it was told. The auto-select-first-featured-variable rule just didn't account for the case where more than one dataset could equally satisfy the selection.

### The fix

Before auto-selecting a value for one input in a linked group (like Gene identifier + Measurement type), we now check whether more than one dataset could still satisfy *every* input in that group. If the choice is genuinely ambiguous, we leave it to the user to pick first — auto-selection only kicks in once there's a single, unambiguous dataset left, which then applies normally.

### Testing

- New unit tests cover the ambiguous (two valid datasets) and resolved (one valid dataset) cases directly.
- Full `eda` package test suite and type-check pass.
- Manually verified against the reported reproduction case: both Gene identifier options now appear, and choosing one correctly narrows Measurement type to the matching dataset.

### DIFF
--- a/packages/libs/eda/src/lib/core/components/visualizations/InputVariables.tsx
+++ b/packages/libs/eda/src/lib/core/components/visualizations/InputVariables.tsx
@@ -6,6 +6,7 @@ import {
   DataElementConstraintRecord,
   filterConstraints,
   disabledVariablesForInput,
+  jointCandidateEntities,
   VariablesByInputName,
 } from '../../utils/data-element-constraints';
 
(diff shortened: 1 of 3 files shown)


bobular (commented):
I understand and approve all this code from Claude.

steve-fischer-200 (approved):

PR #1812, opened 2026-07-22
move source-type checkboxes into table toolbar, standardize casing

bobular:
Move the VEuPathDB curated/Public user/Private user dataset checkboxes into the CommonResultTable children slot so they sit in the Mesa toolbar next to the search box instead of as a separate block above the table. Also standardizes label casing across the toolbar ("Filter datasets:", "Public/Private user datasets") and widens the gap between the search box and the checkbox group.

image

### DIFF
--- a/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/InternalGeneDataset.tsx
+++ b/packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/InternalGeneDataset.tsx
@@ -367,41 +367,8 @@ function InternalGeneDatasetContent(props: Props) {
           ))}
         </div>
       </div>
-      <div className={cx('SourceFilters')}>
-        <label>
-          <input
-            type="checkbox"
-            checked={showDataSources}
-            onChange={(e) => setShowDataSources(e.target.checked)}
-          />
-          <img
-            src={`${webAppUrl}/images/${projectId}/favicon.ico`}
-            alt="VEuPathDB workflow dataset"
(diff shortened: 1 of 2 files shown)


packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/InternalGeneDataset.tsx
```diff
@@ -367,41 +367,8 @@ function InternalGeneDatasetContent(props: Props) {
           ))}
         </div>
       </div>
-      <div className={cx('SourceFilters')}>
```
bobular, line 370: This block moves to be a child of the `InternalGeneDatasetTable` and some CSS magic sorts out the layout.

packages/sites/genomics-site/webapp/wdkCustomization/js/client/components/questions/InternalGeneDataset.tsx
```diff
@@ -566,7 +533,41 @@ function InternalGeneDatasetContent(props: Props) {
...
+            {' VEuPathDB curated datasets'}
+          </label>
+          <label>
+            <input
+              type="checkbox"
+              checked={showPublicUserDatasets}
+              onChange={(e) => setShowPublicUserDatasets(e.target.checked)}
+            />
+            <PublicIcon style={{ width: '20px', height: '20px' }} />
+            {' Public user datasets'}
```
bobular, line 558: Note that I made the case (sentence case) consistent across all three checkbox labels.

PR #1833, opened 2026-08-04
Use wdk citation endpoint

Foxcapades:


packages/libs/user-datasets/src/lib/Components/Sharing/UpdateErrors.tsx
```diff
@@ -60,99 +60,3 @@ export function UpdateErrors({
     </div>
   );
 }
-
```
Foxcapades, line 63: Unrelated to PR. dead code removal.

aurreco-uga (approved):

PR #1835, opened 2026-08-06
Fix UD page refresh button bug

steve-fischer-200:
Clean up logic of the Refresh button in the Dataset Management page. Previously was using hacky approach that allowed for race conditions

### DIFF
--- a/packages/libs/user-datasets/src/lib/Components/Management/DatasetManagement.tsx
+++ b/packages/libs/user-datasets/src/lib/Components/Management/DatasetManagement.tsx
@@ -155,6 +155,9 @@ enum CommunityPromotability {
 class DatasetManagement<
   S extends DatasetManagementState = DatasetManagementState
 > extends React.Component<DatasetManagementProps, S> {
+  private statusUnchangedTimer?: NodeJS.Timeout;
+  private previousInstalledState?: boolean;
+
   constructor(props: DatasetManagementProps) {
     super(props);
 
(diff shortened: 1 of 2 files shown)


packages/libs/user-datasets/src/lib/Components/Management/DatasetManagement.tsx
```diff
@@ -155,6 +155,9 @@ enum CommunityPromotability {
 class DatasetManagement<
   S extends DatasetManagementState = DatasetManagementState
 > extends React.Component<DatasetManagementProps, S> {
+  private statusUnchangedTimer?: NodeJS.Timeout;
```
Foxcapades, line 158: We have access to NodeJS namespace things for client work?

packages/libs/user-datasets/src/lib/Components/Management/DatasetManagement.tsx
```diff
@@ -248,37 +285,49 @@ class DatasetManagement<
...
+      // The thunk will dispatch actions that update Redux state
+      await loadUserDatasetDetailWithoutLoadingIndicator(userDataset.datasetId);
+
+      // After the async operation completes, check if status changed
+      // Note: We need a small delay to ensure React has processed the prop updates
+      setTimeout(() => {
+        const newStatusJson = JSON.stringify(this.props.userDataset.status);
+        const unchanged = currentStatusJson === newStatusJson;
+
+        this.setState({
```
Foxcapades, line 314: Wont this setState call wipe out the other state values (`isCommunityModal`, `datasetUpdateAction`, etc.)?

I was under the impression that `setState` entirely replaced the component state with the given object.  That's why elsewhere we are destructuring the rest of the existing state into the new state object with either `this.setState({...this.state, prop: value})` or `this.setState(oldState => ({...oldState, prop:value}))`.  (See the other `this.setState` calls in this class).

Foxcapades (changes requested):

packages/libs/user-datasets/src/lib/Components/Management/DatasetManagement.tsx
Foxcapades: Apparently react merges the state objects, so this should be fine.

steve-fischer-200:
made irrelevant by introduction of polling instead of refresh button

